PR #23, opened 2015-06-11
remove hidden characters

letsgetrandy:
There were a couple of hidden characters in here causing me errors in Chrome.

I've removed them.

### DIFF
--- a/src/Cerebral.js
+++ b/src/Cerebral.js
@@ -229,7 +229,7 @@ function Cerebral(initialState) {
   // Add recorder signals
   cerebral.signal('recorder.play', function play () { helpers.recorder.play(); });
   cerebral.signal('recorder.stop', function stop () { helpers.recorder.stop(); });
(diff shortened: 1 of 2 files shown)


marbemac:
+1

christianalfoni:
Okay, pushed this now. Will release new version very soon :-)

Btw, how do I avoid these hidden chars? Or at least identify them?

marbemac:
@christianalfoni What I'd suggest is installing the Highlighter sublime plugin, and then adding this to the user settings for that plugin:

```
{
    "highlighter_regex": "(\t+ +)|( +\t+)|[^\\x00-\\x7F]|[\u2026\u2018\u2019\u201c\u201d\u2013\u2014]|[\t ]+$"
}
```

That'll highlight these sorts of annoying hidden ascii characters, for easy removal.

marbemac:
You can also actually see them if you go to File > Reopen with encoding > ISO 8859-1.

PR #33, opened 2015-07-06
Replace all anonymous functions with fat arrow es6 funcitons

alanrsoares:


christianalfoni:
Hi man, thanks for the pull!

Personally I have not decided on using arrows everywhere as it is shorter syntax, or use them where they actually give value... which is situations where you would normally "bind(this)". 

But the trend seems to be "all in" on arrows... so yeah, let me share this on twitter and see if we can get some "status quo" on this :-)

alanrsoares:
Thx for the reply bro

RJ:
I think it's less readable and completely unnecessary to change stand-along functions to arrow functions

arrow funs are great inline, or where you want the this binding. otherwise what's the point? the function keyword makes it more readable imo

Yomguithereal:
Personally, I tend to use the fat arrow notation only for quick and often single expression inline anonymous functions, typically in `map`, `filter` etc. or when I want to avoid a `this` binding. I find it unnecessary and hardly readable in other cases.

alanrsoares:
Thanks for your point G

insin:
Compact method syntax would be a better fit for the change to `demo/cerebral.js`: if the proposed change was only making use of arrow syntax for compactness:

```js
let state = localStorage.store ? JSON.parse(localStorage.store) : {
  todos: {},
  visibleTodos() {
    return {
      initialState: [],
      lookupState: ['todos'],
      get(cerebral, sourceState, refs) {
        return refs.map(ref => sourceState.todos[ref]);
      }
    };
  },
```

Personally, I use arrow syntax to take advantage of its `this` binding or ability to omit `return`, or where I'm defining an inline anonymous function, rather than something top-level or named.

It's a question of where you draw the line - e.g. if you take full advantage of arrow functions the above could be written like this, which some people will find less readable:

```js
import Cerebral from './../src/Cerebral.js';
 
let state = localStorage.store ? JSON.parse(localStorage.store) : {
  todos: {},
  visibleTodos: () => ({
    initialState: [],
    lookupState: ['todos'],
    get: (_, sourceState, refs) => refs.map(ref => sourceState.todos[ref])
  }),
```

christianalfoni:
As Cerebral has taken a turn I am closing this, though thanks so much for contributing :-)

PR #50, opened 2015-08-11
Travis should pick up on failing tests

saulshanabrook:
Currently [failing builds are reported as succeeding on travis](https://travis-ci.org/christianalfoni/cerebral/jobs/74983291#L406-L406). This is because the `test.js` script will always exist succesfully no matter how the tests went.

I [couldn't figure out how to get the exit code from the nodeunit reporter run function](https://github.com/caolan/nodeunit/blob/d8b4f70c868d7c840f39f9e8ef7a848276169138/lib/reporters/default.js#L32), so instead of using the custom `test.js` file I replaced the test commadn with the `nodeunit` cli command. There doesn't seem to be any loss of functionality using that instead of the `test.js` command.

The other option would be to continue using the `test.js` command but somehow get the text outputted by `nodeunit` (it calls `console.log(...)`) and see if there is a failure.


christianalfoni:
Ah, great! Thanks for checking this. I create a commit myself as I want to keep the test.js file. It allows me to run specific test files, but I can just do it running `node test signals`. 

Also fixed the broken tests, forgot to fix test after a change on type checking :-)

Thanks for the pull though!

saulshanabrook:
@christianalfoni Cool I see you have fixed the tests. If they do fail, however, will that show up on travis?

christianalfoni:
Yup, they will, now it runs like it did before. When I did get a few failing tests :-) But I can do a test to verify!

saulshanabrook:
I can do it too :) I am trying it now https://github.com/christianalfoni/cerebral/pull/53

saulshanabrook:
Yep it fails properly!

christianalfoni:
Great, thanks! :-)

PR #52, opened 2015-08-11
Fix `0` and `''` not being recognized properly by the type checker

### LINKED ISSUE #49: Type Checking On Actions Fails on False Values
First off, thank you so much for the new type checker for actions. It really does work wonders for productivity. It seems like when I pass a false-y value in, it says I am giving the wrong input. By that I mean `""` or `0`, even when it is the right type.

saulshanabrook:
This will close #49.

I was having trouble getting it to pick up `0` and `""` as the right types, so I move to using the [type-detect](https://github.com/chaijs/type-detect)t library for all type checking.

If this is accepted, it might make sense to change type checking to use the string that the `type-detect` library uses, instead of passing in the base type things. That we we get a bunch of types for free, like `date`. Of course we still want support for custom functions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,5 +31,8 @@
     "todomvc-common": "^1.0.1",
     "webpack": "^1.5.3",
     "webpack-dev-server": "^1.7.0"
+  },
+  "dependencies": {
+    "type-detect": "^1.0.0"
   }
 }
--- a/src/types.js
+++ b/src/types.js
@@ -1,29 +1,24 @@
+var typeDetect = require('type-detect');
+
+
+var typeToTypeDetect = {}
+typeToTypeDetect[String] = 'string'
+typeToTypeDetect[Number] = 'number'
+typeToTypeDetect[Array] = 'array'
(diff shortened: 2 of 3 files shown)


christianalfoni:
Hi there!

I found the problem, I did a wrong check in regards of functions. It is fixed now.

The reason I am a bit stubborn on not bringing in a default type checker is to avoid having any Cerebral dependencies inside the action definition file. I want it to be as pure as possible.

But it is fixed now, so let us keep it like this for a while longer at least. Maybe people will talk about other type checkers and one of them as voted as the "winner" as default implementation :-)

I am updating the packages to new version of cerebral now.

saulshanabrook:
@christianalfoni Sounds good, i know what you mean about all the type checker options.

On that subject, have you used Flow at all with cerebral?

christianalfoni:
I have not gotten Flow to work with Atom yet. Have not spent much time on it though, but yeah, would be interesting to see how that would look. I have not worked on many large teams, so type checking is not something that is close to my heart, but I do recognize its importance ;-)

PR #54, opened 2015-08-11
Add note about state parameter being read-only in async actions.

bfitch:
I didn't catch that async actions should not mutate state. Added a warning in the README in case others are confused as well.


christianalfoni:
Yes, very good, thanks :-)

PR #67, opened 2015-08-17
Update analyze.js to check to undefined action

idream3:
Avoid the dreaded error: `Uncaught TypeError: Cannot read property 'output' of undefined` that can lead you down a huge timewasting spiral :)

We now get `Uncaught Error: Cerebral: Action number "0" in signal "keyPressed_left" does not exist. Check that you have spelled it correctly!`


christianalfoni:
Sweetness! :-) thanks!

PR #79, opened 2015-08-23
Started changelog to track api changes

a-s-o:
Hello,

I started a change log based on what I could understand from the commits. I think it would be helpful when migrating from one version to the next


christianalfoni:
This is really great, thanks! :-)

PR #106, opened 2015-09-25
Allow for tcomb in type check function

idream3:
tcomb returns the checked value if it passes assertions, eg. 0 is a `Number` but cerebral will return false because of type inference. `Boolean` of value `false` will return false, where you want it to return `true`.


idream3:
Sorry, not going to work... if the result of the type() function returns true/false it will get caught by the boolean check and return true. Hmm, need to think about it more

PR #107, opened 2015-09-25
update todomvc demo to latest cerebral-router

Guria:
[todomvc demo](http://www.christianalfoni.com/todomvc/):
* published with latest router which isn't in deps yet
* `setFilter` action wasn't updated to latest router changes

@christianalfoni  say if I should also add updated `demo_dist` to this commit


Guria:
I also would suggest to completely remove demo_dist from repo. It can be build and published with travis

Guria:
ping @christianalfoni

christianalfoni:
Believe I did update this a few days ago. Let me know if you find any issues. Thanks for the help :-)

PR #146, opened 2015-12-28
Improve controller events

### LINKED ISSUE #138: Improve controller events
I think the controller events could need a little upgrade to be more informative/useful.

Before I create a pull request I would to like to clear a few things up:

The expected event flow should look like this - right?
```
signal-start
  action-start/end
  action-start (with state modification)
  controller-change
  action-end (with state modification)
  action-start/end
signal-end
controller-change
```

Would it make sense to pass the initiator to the change event along with the new state?
e.g. { signal, recorder, state }

Could there be anything else I might want to pass along with a signal event other then the signal itself?
Should I pass `{ signal }` or `signal` - I would prefer the first due to flexibility.

Why is [there](https://github.com/cerebral/cerebral/blob/master/src/CreateSignalFactory.js#L158) a change event fired after the `actionStart`.

maxfrigge:
Resolves #138


christianalfoni:
Cool, awesome, just have to convert it to ES5... but no worries, I fix that :-)

maxfrigge:
Ahh.. didn't know that was a requirement :)

Thanks!

PR #149, opened 2015-12-30
controller.extends now supports signals, services and other meta

garth:
Controller keeps a reference to modules which can be used by cerebral-react so that we can pass modules to components.

Example module:

```js
// /app/modules/home/index.js
import Component from './components';

export default {
  init() {
    return {
      Component
    };
  },
  signals: {
    opened: []
  },
  services: {}
};
```

Loading modules:

```js
// /app/main.js
// load the modules
import pouchdb from 'cerebral-pouchdb';
import home from './modules/home';
import notFound from './modules/notFound';
const modules = {
  db: pouchdb({
    localDb: 'mydb',
    remoteDb: 'http://localhost:3000/db/mydb'
  }),
  home,
  notFound
};

// init the modules
import controller from './controller';
controller.extends(modules);

// init the router
import Router from 'cerebral-router';
Router(controller, {
  '/': 'home.opened',
  '*': 'notFound.opened'
});
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -53,21 +53,35 @@ var Controller = function (Model, services) {
   controller.devtools = devtools;
   services.recorder = recorder;
 
+  controller.modules = {};
   controller.extends = function (modules) {
-    var meta = {};
     Object.keys(modules).forEach(function (moduleName) {
       var module = modules[moduleName];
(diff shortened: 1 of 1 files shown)


christianalfoni:
Sweetness :-)

PR #150, opened 2015-12-30
Add name property to modules

garth:
Seems like it might be useful to add `name` to each `cerebral.modules` incase they are passed around individually.


garth:
Added by @christianalfoni

PR #151, opened 2015-12-30
Feature(modules): make signals optional and support sync signals

abalmos:
1. Adds support for modules without signals.
2. Adds support for registering sync signals. A module should be able to require that a signal is sync (same justification behind `controller.signalSync()`.

Modules now look like:

```js
export default {
   init({controller, name, signals}) {
     return ...
   },
  signals: {},
  syncSignals: {},
  services: {}
}
```

Not sure I'm sold on `syncSignals`. Maybe `signalsSync` to stay consistent with `signalSync()`?

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -63,6 +63,12 @@ var Controller = function (Model, services) {
           controller.signal(signalName, module.signals[key]);
         }
       });
+      Object.keys(module.syncSignals || {}).forEach(function (key) {
+        if (Array.isArray(module.syncSignals[key])) {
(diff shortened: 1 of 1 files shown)


garth:
Do we need to export sync signals from modules?

christianalfoni:
Will merge in an remove syncSignals until we have a module that actually needs it. There are two other strategies for now, either manually register in init or call `.sync()` on the signal from the component using it

christianalfoni:
Will just change the name of the property to `signalSync`, for consistency

PR #161, opened 2016-01-10
Register signals to a global hook for the cerebral debugger

NervosaX:
This is reliant on this pull request: https://github.com/cerebral/cerebral-chrome-extension/pull/15

This will register all signals to a global hook (if found). This allows the debugger to inspect actions (show them in the source tab)


christianalfoni:
Okay, let me do this manually, I will move the logic a bit... but looks good. Let you know when I have tested :)

christianalfoni:
Closing this as we almost got it working, just need to test a bit more :-)

PR #168, opened 2016-01-13
Changed Event to CustomEvent for IE compatibility

### LINKED ISSUE #167: Internet Explorer throws Object doesn't support this action
Internet Explorer up to 11 breaks on [this line](https://github.com/cerebral/cerebral/blob/master/src/Devtools.js#L149) and basically any other line that calls `new CustomEvent()`. 

The error thrown and a fix is described in [this Stack Overflow post](http://stackoverflow.com/questions/26596123/internet-explorer-9-10-11-event-constructor-doesnt-work)

maxfrigge:
Fixes #167


christianalfoni:
aha, let me move that in

PR #171, opened 2016-01-18
Standard

Guria:
@christianalfoni make a quick check before I would merge it


Guria:
pls do not use Merge pull request button

PR #179, opened 2016-01-22
Computed factories

Guria:
we can create a pr to make reviews

### DIFF
--- a/src/Compute.js
+++ b/src/Compute.js
@@ -1,6 +1,7 @@
 module.exports = function (model) {
   var registered = []
   var computed = []
+  var cachedByRef = {}
 
   var createMapper = function (cb) {
     var initialRun = true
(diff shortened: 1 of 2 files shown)


Guria:
you have one extra commit, which is better to avoid. you shouldn't merge `master` to `feature-branch`, but run `git rebase master` to actualize feature-branch

Guria:
now you should run
```sh
git rebase @~2 -i # nano would be bring, just remove line with merge commit, save and exit
git rebase origin/master # rebase feature-branch on top of master
git push origin computed-factories - f # force push to feature-branch to update pr
```

Guria:
Then you have to wait until travis finish to check pr. But today travis is slow as hell.

christianalfoni:
Okay, cool! I need @garth url scheme compiler as well. Moving it into its own repo with same configuration as addons. So that will take some time :)

Guria:
tests are done

Guria:
maybe we should batch some prs into one push, since travis is slow

PR #182, opened 2016-01-22
introduce `signalTrigger` event

Guria:
includes #179


Guria:
cc @christianalfoni need your LGTM
I need this to override `isSync` option in router

PR #191, opened 2016-01-28
Avoid using deprecated methods internally

gor181:


christianalfoni:
Thanks @gor181 , but these two modules will actually be split out in the next version :-) Which is right around the corner! 0.33 is a short "in between" version to move to new api signature

PR #192, opened 2016-01-28
Action errors

Guria:


src/CreateSignalFactory.js
```diff
@@ -153,7 +153,24 @@ module.exports = function (controller, model, services, compute, modules) {
...
+                if (utils.isDeveloping()) {
+                  try {
+                    actionFunc(actionArg)
+                  } catch (e) {
+                    action.error = {
+                      name: e.name,
+                      message: e.message,
+                      stack: actionFunc.toString()
+                    }
+                    controller.emit('signalError', {action: action, signal: signal})
```
Guria: why not `actionError`?

christianalfoni: Because it is the signal that is aborted, not only the action. The signal is aborted because of an action, which is also passed on the event :)

Guria: would the rest of actions ran?

Guria: I just confused that we pass argument for particular action. What if several actions fails? We will get several `signalError` for the same signal.

christianalfoni: Can actually only get one signalError. So what happens is:

1. Signal starts and runs first action successfully
2. Signal continues on next action which throws an error
3. Cerebral picks it up, attaches it to the action, triggers and error event and then rethrows the exception
4. Cerebral just stops, because the current execution stack has thrown an error. But it has gotten a chance to inform the debugger about it

Guria: let make it then:
```js
signal.error = {
  name: e.name,
  message: e.message,
  action: action,
  stack: actionFunc.toString()
 }
 controller.emit('signalError', {signal: signal})
```

PR #193, opened 2016-01-28
feat(debugger): wrap services to track usage

Guria:
tracking service usage will allow debugger to display the information

### DIFF
--- a/src/createActionArgs.js
+++ b/src/createActionArgs.js
@@ -40,17 +40,54 @@ var createStateArg = function (action, model, isAsync, compute) {
   return state
 }
 
+var createServicesArg = function (action, services, moduleKeys) {
+  var path = []
+
+  var convertServices = function (moduleServices) {
+    return Object.keys(moduleServices).reduce(function (newModuleServices, key) {
+      path.push(key)
+      if (typeof moduleServices[key] === 'function') {
+        var servicePath = path.slice()
+        var method = servicePath.pop()
+        newModuleServices[key] = function () {
+          action.serviceCalls.push({
+            name: servicePath.join('.'),
+            method: method,
+            args: [].slice.call(arguments)
+          })
+          return moduleServices[key].apply(moduleServices[key], arguments)
+        }
+      } else if (typeof moduleServices[key] === 'object' && !Array.isArray(moduleServices[key]) && moduleServices[key] !== null) {
+        newModuleServices[key] = convertServices(moduleServices[key])
+      } else {
+        newModuleServices[key] = moduleServices[key]
+      }
+      path.pop(key)
+      return newModuleServices
+    }, {})
+  }
+
+  return Object.keys(services).reduce(function (newServices, key) {
+    path.push(key)
+    newServices[key] = convertServices(services[key], key)
+    path.pop(key)
+    return newServices
+  }, {})
+}
+
 module.exports = {
-  sync: function (action, signalArgs, model, compute) {
+  sync: function (action, signalArgs, model, compute, services, moduleKeys) {
     return [
       signalArgs,
-      createStateArg(action, model, false, compute)
+      createStateArg(action, model, false, compute),
+      createServicesArg(action, services, moduleKeys)
     ]
   },
-  async: function (action, signalArgs, model, compute) {
+  async: function (action, signalArgs, model, compute, services, moduleKeys) {
     return [
       signalArgs,
-      createStateArg(action, model, true, compute)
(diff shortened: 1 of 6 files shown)


src/createActionArgs.js
```diff
@@ -40,17 +40,37 @@ var createStateArg = function (action, model, isAsync, compute) {
...
+      newServices[key] = function () {
+        action.services.push({
+          name: parentKey || 'root',
+          method: key,
+          args: [].slice.call(arguments)
+        })
+        return services[key].apply(services[key], arguments)
+      }
+    } else {
+      newServices[key] = createServicesArg(action, services[key], key)
```
Guria: it is breaking change since it breaks support of non-function values in services. Am I right?

tests/staticTree.js
```diff
@@ -13,6 +13,7 @@ suite['should convert actions to objects'] = function (test) {
     duration: 0,
     path: [0],
     mutations: [],
+    services: [],
```
Guria: maybe better `serviceCalls` or smth else?

src/createActionArgs.js
christianalfoni: If anyone has done that, yes :-) Though I can not see why a service would be anything else than methods or nested objects with methods?

tests/staticTree.js
christianalfoni: Hm, yeah, I agree. `serviceCalls` is good! Should I make another commit or do we change existing? I can try to change existing myself :)

src/createActionArgs.js
Guria: in `cerebral-module-hoodie` I was putting pre-configured `hoodie` object which is EventEmitter and has keys that not functions.

tests/staticTree.js
Guria: Yeap. It is better to change existing.

src/createActionArgs.js
Guria: I don't want to lose such ability.

christianalfoni: Hm, looking at the hoodie module you could still do that. Instead of:

```js
    services: {
      store: store,
      detach: function detach () {
        store.off('change', callbacks.onStoreChange)
        store.off('clear', callbacks.onStoreClear)
      }
    }
```

You would:

```js
    store

    services: {
      store: {
        func1: store.func1.bind(store),
        func2: store.func2.bind(store),
        getSomeValue: function () { return store.someValue; }
      },
      detach: function detach () {
        store.off('change', callbacks.onStoreChange)
        store.off('clear', callbacks.onStoreClear)
      }
    }
```

In my opinion it opens up new possibilities to do it like this. Cerebral will now understand how you use your services, which might not only be beneficial for the debugger, but could lead to other solutions as well. If we let services be "whatever", there is no way Cerebral can understand them :-/ I think we will miss out on possibly great things

tests/staticTree.js
christianalfoni: Cool, let me try without your help first :)

src/createActionArgs.js
Guria: current hoodie module wraps only store api. [compete hoodie-client api](https://github.com/hoodiehq/hoodie-client#api) would be a lot of painful and pointless wrapping.
I can wrap it with one method like `getHoodie(): { return hoodie }` but debugger will lost actual service run as well.
Can't we just leave non-function values untouched or wrapped with getter and setter?

christianalfoni: Okay, been thinking. We can do what you suggest. I was just unsure how to identify what module the service belonged to when it was nested, but found an easy way to identify now. Let me do another iteration and lets have a new look :)

Wont be ready before tomorrow morning though, got the chat on Falcor and need some rest. But yeah, tomorrow morning I will have a happy solution :)

tests/store.js
```diff
@@ -71,7 +71,9 @@ suite['should not store default args'] = function (test) {
   ]
 
   ctrl.services({
-    utils: 'test'
+    utils: function () {
```
Guria: @christianalfoni what about better test case?

christianalfoni: Hm, I think the main test suite above covers the complexity of service registration. Not really sure what this test actually does? I think we should have a test review when moving to 0.34, as we have to go through all tests anyway to change usage of signals vs addSignals and so on.

Guria:
I have added #195 and #192 here.
releasing it

christianalfoni:
Very very awesome!

PR #197, opened 2016-01-30
fix(signalFactory): prevent circular reference stack overflow when wr…

Guria:
…apping services

### DIFF
--- a/tests/signals.js
+++ b/tests/signals.js
@@ -982,4 +982,32 @@ suite['should wrap and track use of services'] = function (test) {
   test.done()
 }
 
+suite['should handle circular references in services'] = function (test) {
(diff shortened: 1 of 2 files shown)


tests/signals.js
```diff
@@ -982,4 +982,32 @@ suite['should wrap and track use of services'] = function (test) {
...
+  var service = {}
+  var obj1 = {}
+  var obj2 = {}
+
+  service.obj1 = obj1
+  service.obj2 = obj2
+  service.obj1.obj2 = obj2
+  service.obj2.obj1 = obj1
+
+  ctrl.addServices({
```
Guria: es6 things left here

PR #198, opened 2016-01-30
fix(signalStore): wrap prototyped methods

Guria:
wrapped services also needs to wrap prototype methods


Guria:
`signalStore` again :)

PR #202, opened 2016-02-02
feat(signalFactory): emit active and idle events

Guria:
when controller starts and stops execution, emit events

### DIFF
--- a/src/CreateSignalFactory.js
+++ b/src/CreateSignalFactory.js
@@ -5,13 +5,24 @@ var analyze = require('./analyze.js')
 var staticTree = require('./staticTree')
 var createModulesArg = require('./createModulesArg.js')
 
-var batchedSignals = []
-var pending = false
 var requestAnimationFrame = global.requestAnimationFrame || function (cb) {
   setTimeout(cb, 0)
 }
 
 module.exports = function (controller, model, services, compute, modules) {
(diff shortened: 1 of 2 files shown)


src/CreateSignalFactory.js
```diff
@@ -100,6 +113,7 @@ module.exports = function (controller, model, services, compute, modules) {
             signal.isExecuting = false
             controller.emit('signalEnd', {signal: signal})
             controller.emit('change', {signal: signal})
+            emitControllerStatus(-1)
```
Guria: Is it possible that signal fail before reaching here and it will not be executed?

src/CreateSignalFactory.js
```diff
@@ -5,13 +5,24 @@ var analyze = require('./analyze.js')
...
 var requestAnimationFrame = global.requestAnimationFrame || function (cb) {
   setTimeout(cb, 0)
 }
 
 module.exports = function (controller, model, services, compute, modules) {
+  var currentlyRunningSignals = 0
+  var batchedSignals = []
+  var pending = false
+
+  var emitControllerStatus = function (change) {
```
Guria: seems overcomplicated

christianalfoni:
Lets close this as we took a different direction

PR #211, opened 2016-02-10
fix(errors): don't throw exceptions when listeners attached to signalError

jeroenverfallie:
Straight forward.

See docs for the boolean: https://nodejs.org/api/events.html#events_emitter_emit_event_arg1_arg2

### DIFF
--- a/src/CreateSignalFactory.js
+++ b/src/CreateSignalFactory.js
@@ -173,9 +173,12 @@ module.exports = function (controller, model, services, compute, modules) {
                       message: e.message,
                       stack: actionFunc.toString()
                     }
-                    controller.emit('signalError', {action: action, signal: signal})
+                    var hadListeners = controller.emit('signalError', {action: action, signal: signal})
                     controller.emit('change', {signal: signal})
-                    throw e
+
+                    if (!hadListeners) {
+                      throw e
+                    }
                   }
                 } else {
                   actionFunc(actionArg)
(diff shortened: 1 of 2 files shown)


jeroenverfallie:
This does of course mean that you are responsible for handeling the exception, as otherwise it's suppressed.

But I believe it makes sense you can catch it, otherwise the event is missing its use.

jeroenverfallie:
Did the same for the normal `error` listener, and changed `const` to `var`.. sorry.. habits.

Guria:
@jeroenverfallie could be it also tested?

Guria:
@jeroenverfallie please squash commits and use `signalFactory` as scope in message:
```js
git rebase @~2 -i # follow instructions to mark commits to squash, save and exit
git push -f # force push to your fork
```

jeroenverfallie:
Will do..

@christianalfoni any reason for https://github.com/cerebral/cerebral/blob/master/src/CreateSignalFactory.js#L174 
to be different to 
https://github.com/cerebral/cerebral/blob/master/src/CreateSignalFactory.js#L258

Is it used in the debugger?

jeroenverfallie:
Had to use a try catch instead of asserts throws, because I had no way of knowing when the signal ended, because signalEnd doesn't reach, nor could I attach to signalError for testing purposes :)

jeroenverfallie:
After really thinking this through, this is not the way to go as you do want your state to update with errors, and update the UI, thus breaking a signal doesn't work, obviously. 

Closing until there's a better solution.

Issue #222, opened 2016-03-07
decide correct form of `sync` vs `isSync` signal flag


Guria:
https://github.com/cerebral/cerebral/commit/03afa0f179e66018ab25aadf0abe8d1705d15901#commitcomment-16534275

Guria:
@christianalfoni will we support both or one of them should be deprecated?

christianalfoni:
Lets move to `sync`, deprecate `isSync`. More consistent with `async` prop on actions and `routed` prop

Guria:
ok. got it

Guria:
btw, there is no `routed` prop.

Guria:
https://github.com/cerebral/cerebral/blob/master/src/CreateSignalFactory.js#L39-L55

christianalfoni:
Hm, I am confusing with debugger. Hmf, is a difference between user API and internal API in this matter. Like `sync` is exposed API, and `isRecorded` and `isRouted` is more internal...

hm hm...

Since `isSync` is in between maybe we should support both. `isSync` and `sync`.

christianalfoni:
Either that or only support `isSync`.

Guria:
wip in `pre-0.34`

Guria:
decided to use `immediate` option

Issue #221, opened 2016-03-07
rename back test suite


Guria:
https://github.com/cerebral/cerebral/commit/9c81b36785b73f8dfeb80b2a64a311e4b0792349#commitcomment-16533549
`'should make copy of payload'` => `'should register signals'`

Guria:
wip in `pre-0.34`

christianalfoni:
Oh, how did that happen? But yeah, rename back :)

Issue #219, opened 2016-03-07
deprecate signal.sync(payload) in favor of signal(payload, { isSync: true })


Guria:
cc @christianalfoni 
do you agree?

christianalfoni:
Yes, though use `sync` as pointed out in previous issue

Guria:
wip in `pre-0.34`

Issue #218, opened 2016-03-07
A tiny bug in cerebral/src/createNext.js


Searle:
Hi! I was just looking at your code and my eyes catched a potential problem in cerebral/src/createNext.js:

I believe
`if (checkers === undefined && arg === undefined) {`
should be
`if (checkers === undefined || arg === undefined) {`

Otherwise, if checkers xor args is undefined, the following code would blow up. Object.keys(undefined) is not allowed and arg[key] would be a problem to.

Guria:
cc @christianalfoni

christianalfoni:
Yes indeed, thanks for the heads up @Searle . Validation of inputs/outputs is not as broadly used so this has slipped by :-)

Guria:
took it to pre-0.34

Guria:
@christianalfoni this change breaks test: `should throw error when output is missing`

christianalfoni:
Looking into it now :)

christianalfoni:
Okay, this is now pushed as well

PR #224, opened 2016-03-07
Pre 0.34

### LINKED ISSUE #200: support `path` argument for `get*()` methods
```js
controller.getServices().module.submodule
controller.getServices('module.submodule')
controller.getServices(['module', 'submodule'])
```

Guria:
fixes #200 #221 #219 #222 #218 #225 #226


src/index.js
```diff
@@ -1,3 +1,4 @@
+var get = require('lodash/get')
```
Guria: @christianalfoni don't you mind on using lodash here? we are importing only separate function.

christianalfoni: I think that is okay, I see other projects does this as well and they should really :) No need to reinvent the wheel

Guria:
Ok. LGTM Let's make it live

PR #237, opened 2016-04-09
0.34: contextProviders and deprecation cleanups

Guria:


src/CreateSignalFactory.js
```diff
@@ -1,9 +1,13 @@
 var utils = require('./utils.js')
-var createActionArgs = require('./createActionArgs.js')
+// var createActionArgs = require('./createActionArgs.js')
```
Guria: seems could be removed

src/CreateSignalFactory.js
```diff
@@ -140,30 +143,25 @@ module.exports = function (controller, model, services, compute, modules) {
...
                 var inputArg = actionFunc.defaultInput ? utils.merge({}, actionFunc.defaultInput, signalArgs) : signalArgs
-                var actionArgs = createActionArgs.async(action, inputArg, model, compute, services, Object.keys(modules))
+                var next = createNext.async(actionFunc, signal.name)
+                var context = createContext([
+                  inputProvider(inputArg),
+                  stateProvider(action, model, compute, true),
+                  servicesProvider(action, modules, services),
+                  outputProvider(next.fn)
+                ].concat(options.context || []))
+                // var actionArgs = createActionArgs.async(action, inputArg, model, compute, services, Object.keys(modules))
```
Guria: remove

src/CreateSignalFactory.js
Guria: also `createActionArgs.js` should be removed as well

src/CreateSignalFactory.js
```diff
@@ -222,7 +220,14 @@ module.exports = function (controller, model, services, compute, modules) {
...
               var inputArg = actionFunc.defaultInput ? utils.merge({}, actionFunc.defaultInput, signalArgs) : signalArgs
-              var actionArgs = createActionArgs.sync(action, inputArg, model, compute, services, Object.keys(modules))
+              var next = createNext.sync(actionFunc, signal.name)
+              var context = createContext([
+                inputProvider(inputArg),
+                stateProvider(action, model, compute, false),
+                servicesProvider(action, modules, services),
+                outputProvider(next)
+              ].concat(options.context || []))
+              // var actionArgs = createActionArgs.sync(action, inputArg, model, compute, services, Object.keys(modules))
```
Guria: remove

src/CreateSignalFactory.js
```diff
@@ -1,9 +1,12 @@
 var utils = require('./utils.js')
-var createActionArgs = require('./createActionArgs.js')
```
Guria: you forgot to remove createActionArgs.js file

christianalfoni: Ops, thanks for checking, fixing :)

Guria:
@christianalfoni please point to corresponded issues on separate line either with `fix #issue` or `close #issue`

Guria:
this line should have one blank line before it

Guria:
@christianalfoni let's mark `feat(events): pass options passed to signal on the events as options` as breaking with note to update devtools

christianalfoni:
Actually it is not breaking :) It only updates the signal from options if options exist and signal does not have the keys already (old behaviour)

christianalfoni:
I can set as breaking, though devtools works with both new and old behaviour :)

Guria:
Ahh, ok. Not breaking here :)

christianalfoni:
Okay, cool, leaving it then :)

Guria:
technically `signal` object passed to events looks still mutable. Am I right?

christianalfoni:
Yeah, technically it is, though we just do not do it anymore :) Just to much unnecessary processing power to deep copy everything when we can just say... "Don't do it" :-)

Guria:
@christianalfoni how about moving services wrapper into devtools:
```js
function Devtools (options) {
  return function (module, controller) {
    /* ... */
    controller.addContextProvider(function (context) {
        return {
          ...context,
          services: convertServices(context.services)
        }
    })
    /* ... */
  }
}
```

christianalfoni:
Oh! I love it :-)

This will also allow for context plugins for Cerebral. We should get that in there! I can fix that and debugger problem this evening :)

Guria:
Yey, `rc.2` is coming :)

src/index.js
```diff
@@ -85,6 +86,9 @@ var Controller = function (Model) {
     })
     return controller.getServices()
   }
+  controller.addContextProvider = function (provider) {
+    externalContextProviders.push(provider)
```
Guria: I think we can check provider existence in this array.

christianalfoni: Have to do it later because you might add a provider from inside a module:

```js
module.addSignals({}, {
  context: ModuleProvider
});
```

Guria:
> Guria - Today at 3:04 PM
do we really need per signal context override?
maybe stick to controller.addContextProvider and module.addContextProvider ?
christianalfoni - Today at 3:07 PM
Good question, I think we need to keep it completely open now because we do not know how it will be used yet. Would be terrible if we close it up now and suddenly there is a scenario we can not solve 😕
Guria - Today at 3:08 PM
But if we'll open it now it would painfull to close if we decide that it was wrong
christianalfoni - Today at 3:08 PM
Yeah, but it is doable 😃 If we close it now we can not go back 😃
Or... hm
that was maybe a bit dramatic, hehe
Or
hm
Guria - Today at 3:09 PM
We would also havemore consistent API between controller.add* and module.add* methods(edited)
christianalfoni - Today at 3:09 PM
Yeah, that part I like
Though that would mean we should remove context all together from signals options
Guria - Today at 3:10 PM
And there will be no special option for signals
christianalfoni - Today at 3:10 PM
yeah
Hm...
And then we could later add the option back if somebody absolutely have to do it differently

Guria:
I think it worth to keep current commits for history reasons, and just move forward from current point. Do you agree, or would prefer make clean pr?

christianalfoni:
That is okay by me! I do not have time to implement this today, but I can do it tomorrow :)

christianalfoni:
That is okay by me! I think I can fix this later today, or I will do it tomorrow :)

src/CreateSignalFactory.js
```diff
@@ -150,45 +139,45 @@ module.exports = function (controller, model, services, compute, modules) {
...
             } else {
               var promises = currentBranch.map(function (action) {
-                controller.emit('actionStart', {action: action, signal: signal})
+                controller.emit('actionStart', {action: action, signal: signal, options: options, payload: payload})
                 var actionFunc = actions[action.actionIndex]
                 var inputArg = actionFunc.defaultInput ? utils.merge({}, actionFunc.defaultInput, signalArgs) : signalArgs
-                var actionArgs = createActionArgs.async(action, inputArg, model, compute, services, Object.keys(modules))
+                var next = createNext.async(actionFunc, signal.name)
+                var contextProviders = [
+                  inputProvider(inputArg),
```
Guria: It looks like context provider factories, but not context providers.
Maybe better call it like `createInputProvider`?

Guria: Ohh, I think `inputProvider` could be rewritten to be `provider`. Aren't it's possible to extract `inputArgs` from `execution`?

Guria:
@christianalfoni okay, I have a challenge for you. How about to make default providers to be actual providers without factories:
* `inputProvider` could extract input from `execution` argument
* `outputProvider` could create `next.fn` from `execution` argument
* `servicesProvider` could get services from controller if we will pass it as third argument
```js
function servicesProvider (context, execution, controller) {
  context.services = controller.getServices()
  return context
}
```
* `devtoolsServicesProvider` could get modules from `controller`
* `stateProvider`:
  * could extract `action.isAsync` from `execution` argument.
  * could use `controller.get` instead of direct access to `compute.getComputedValue`
  * could get model from `controller` if it will be exposed on it
  * could wrap mutators to store `mutation` list in `execution` (potentially could be moved to devtools / recorder with predefined execution logic)

christianalfoni:
Really great insight here @Guria , "CHALLENGE ACCEPTED" :D

christianalfoni:
Hm, I see one issue here and that is `outputProvider`. It creates the promise the signal uses to control the flow. So it is not just changing the context it is actually returning a resolver to the signal. So that will not work.

Though... hm, I could create the resolver outside and pass it is as part of the execution. Yeah, that makes sense as it would be possible to create other kinds of resolvers, not just outputs :)

src/CreateSignalFactory.js
```diff
@@ -1,5 +1,5 @@
 var utils = require('./utils.js')
-var createNext = require('./createNext.js')
+// var createNext = require('./createNext.js')
```
Guria: should be removed

Guria:
@christianalfoni 
```js
// cerebral-module-predefined-execution
var mutationProvider = require('mutationProvider')

module.exports = function PredefinedExecution () {
  return function (module, controller) {
    module.alias = 'cerebral-module-predefined-execution'

    controller.addContextProvider(mutationProvider)
    controller.on('signalTrigger', function (event) {
      if (event.signal.options.branches) {
        event.signal.preventSignalRun()
        /* predifined execution here */
      }
    })
  }
}
```
```
// cerebral-module-signal-store
var PredefinedExecution = require('cerebral-module-predefined-execution')

module.exports = function SignalStore () {
  return function (module, controller) {
  /* ... */
  if (!controller.getModules('cerebral-module-predefined-execution')) {
    controller.addModules({
      'cerebral-module-predefined-execution': PredefinedExecution()
    })
  }
  /* ... */
  }
}
```
```
// cerebral-module-recorder
var PredefinedExecution = require('cerebral-module-predefined-execution')

module.exports = function Recorder () {
  return function (module, controller) {
  /* ... */
  if (!controller.getModules('cerebral-module-predefined-execution')) {
    controller.addModules({
      'cerebral-module-predefined-execution': PredefinedExecution()
    })
  }
  /* ... */
  }
}
```

`'cerebral-module-predefined-execution'` name are subject to change :)

src/CreateSignalFactory.js
Guria: commented code here and below should be removed

christianalfoni:
Hm, I think this actually should be inside Cerebral. So the scope of Cerebral is to run the signal and produce the context for each action using its providers. So running the signal includes also running predefined signals. Decoupling is good, but should not decouple just because we can :)

Does that make sense?

src/CreateSignalFactory.js
```diff
@@ -150,64 +137,71 @@ module.exports = function (controller, model, services, compute, modules) {
...
-                if (utils.isDeveloping() && actionFunc.input) {
-                  utils.verifyInput(action.name, signal.name, actionFunc.input, inputArg)
-                }
+                var contextProviders = [
+                  inputProvider,
+                  stateProvider,
+                  servicesProvider,
+                  outputProvider
+                ].concat(utils.extractExternalContextProviders(externalContextProviders, options.modulePath))
+                .concat(options.context || [])
```
Guria: do we still want to allow define context providers for each signal?
I thought we decided to keep options clean from "special options". Also it introduces extra way to register context provider which is redundant.

src/CreateSignalFactory.js
```diff
@@ -1,15 +1,18 @@
...
+var stateProvider = require('./contextProviders/stateProvider')
+var servicesProvider = require('./contextProviders/servicesProvider')
+var outputProvider = require('./contextProviders/outputProvider')
 
 var requestAnimationFrame = global.requestAnimationFrame || function (cb) {
   setTimeout(cb, 0)
 }
 
-module.exports = function (controller, model, services, compute, modules) {
+module.exports = function (controller, model, services, compute, modules, externalContextProviders) {
```
Guria: `model` could be now got from `contoller.getModel()`
`services, compute, modules` is not used anymore :)

Guria:
@christianalfoni it means that mutation-provider should be still in `cerebral` too. Let ship it with `cerebral` package, just not using by default. `signal-store` and `recorder` would just add it with `controller.addContextProvider()`

src/CreateSignalFactory.js
christianalfoni: Ah, yeah, though I thought we could just keep it there in case we meet some scenario. But its not an official part of the API.

src/CreateSignalFactory.js
christianalfoni: Ah, yeah, good point, let me go through that again :)

christianalfoni:
That is a good idea :-) Let me change that around

src/CreateSignalFactory.js
Guria: I really don't like this part. I thought we decided that reintroduce it will be easy.

christianalfoni: Okay, let me remove it :)

Guria:
@christianalfoni rebased to include #239 fix

Guria:
`cerebral-0.34.0-rc.2` release using `next` tag.
`cerebral-module-devtools@next` and `cerebral-module-recorder@next` could be used with this release as well

Guria:
:fire: it's awesome

Guria:
+ cerebral@0.34.0-rc.4
'dist-tags': { latest: '0.33.33', next: '0.34.0-rc.4' },

providers/deprecationProvider.js
```diff
@@ -0,0 +1,19 @@
+module.exports = function (context, execution, controller) {
+  if (!context.module && !context.modules) {
+    Object.defineProperty(context, 'module', {
+      get: function () {
+        console.warn('Cerebral 0.34 uses a provider to access "module" and it has become more powerful. Please install: https://github.com/cerebral/cerebral-provider-modules')
```
Guria: make warning more context aware: signalName, action, etc

providers/deprecationProvider.js
```diff
@@ -0,0 +1,19 @@
...
+        return null
```
Guria: maybe better just throw error here instead of return?

christianalfoni: I completely agree, let me fix :-)

PR #238, opened 2016-04-10
feat(deprecation): deprecate router using isSync on signals

christianalfoni:


Guria:
okay, let me prepare new router first :)

PR #249, opened 2016-06-02
fix(providers): action mutations provider should handle select paths

christianalfoni:


Guria:
I think you know better what are you doing here.
Is it just bug fix and not related to any new *-tree stuff?

christianalfoni:
Yeah, just a bug fix for the debugger. It did not pass paths correctly using the `select` api on `state` :) I will merge it in

PR #267, opened 2016-07-07
feat(operators): Export cerebral-operators under cerebral/operators

garth:


saulshanabrook:
Is this correct for the new imports?

```javascript
import {copy} from 'cerebral/operators'
```

garth:
@saulshanabrook yes, either of the following should be ok:

```js
import { copy } from 'cerebral/operators'

import copy from 'cerebral/operators/copy'
```

christianalfoni:
Jup, tha t is how it should work @saulshanabrook

PR #276, opened 2016-07-15
fix(computed): make callback part of cache key

christianalfoni:
We actually got into a situation where two computeds has the exact same state dependencies, but just runs them differently. So we need to put the actual function string representation into the computed cache key as well :)


maxfrigge:
The function name is not unique .. so it won't fully fix the issue.. just make it less frequent :)

christianalfoni:
If it is the same deps, properties passed and the same function it will give the same result... even though the second computed will actually grab the value from the first ;-)

PR #277, opened 2016-07-16
fix(debugger): mutation needs a datetime for debugger to sort them in…

christianalfoni:
… correct order

I have fixed the debugger, but the solution is not performant. It reruns the whole history of mutations on every signal change. With this fix the debugger can control the order of mutations. Even datetimes that are the same will keep their order in the list, so this works great. Tested, and will not break old version.


Guria:
Lgtm

PR #280, opened 2016-07-22
fix(computed): make nested computed update as well

christianalfoni:
There was a bug related to nested computed


Guria:
Lgtm. Push to master

PR #281, opened 2016-07-25
Fix serverController get method and add test case to check state overwrite

psypersky:


unity-haiwei:
Please merge this ASAP. it's breaking our build....

src/index.js
```diff
@@ -112,10 +112,9 @@ module.exports.ServerController = function (state) {
     accessors: {
       get: function (path) {
         path = typeof path === 'string' ? path.split('.') : path
-        while (path.length) {
-          state = state[path.shift()]
-        }
-        return state
+        return path.reduce(function (currentState, key) {
```
unity-haiwei: Cannot read property 'setting' of undefined, when the path is 'head.setting'. In fact, simply no ‘head’.

Guria: can you provide test case?

unity-haiwei: var ctrl = ServerController({foo: null, 'boo': 'bee'}) from server render.

Guria: Cc @christianalfoni

christianalfoni: Hm... yeah, I think you are right @unity-haiwei. It should return "undefined". So:

```js
  return path.reduce(function (currentState, key) {
      return currentState === undefined || currentState === null ? undefined : currentState[key]
   }, state)
```

Now it returns undefined if the path does not exist or you try to access a key on `null`.

PR #282, opened 2016-07-27
feat(computed): allow * and ** selectors in computed path

christianalfoni:
to experiment with the new strict render syntax we need computed to handle the glob pattern


Guria:
LGTM

PR #286, opened 2016-08-02
fix(computed): force computed to return value in actions

christianalfoni:
Due to computed now being "view driven", meaning it updates on "flush event" it needs to force recomputing its value when used in actions using: `state.computed`. The bottleneck of computed is not in signal, it is in rendering, so this is perfectly okay :)

### DIFF
--- a/providers/stateProvider.js
+++ b/providers/stateProvider.js
@@ -43,7 +43,7 @@ module.exports = function (context, execution, controller) {
     }
 
     state.computed = function (computed) {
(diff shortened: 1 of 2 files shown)


Guria:
Lgtm

christianalfoni:
What is that patch thingy error?

Guria:
It just about coverage. You have increased number of possible branches, but don't cover with tests. You have two options. Ignore or add tests. :)

fmal:
@christianalfoni could you please merge this?

PR #289, opened 2016-08-05
fix(computed): throw error when not passing props as object

christianalfoni:


edgesoft:
Looks good

src/Computed.js
```diff
@@ -33,6 +33,16 @@ function traverseDepsMap (deps, cacheKey) {
 
 function Computed (paths, cb) {
   return function (props) {
+    if (
```
Guria: should we wrap it with production env check?

PR #296, opened 2016-08-14
fix(staticTree): always create new static tree during development

christianalfoni:


Guria:
Is it really only solution?

christianalfoni:
The challenge is that we attach mutations to the action that runs on the signal tree. And this signal tree is serliazed and passed to debugger. So we would have to store this somewhere else and do lookups.

We could for example expose the path to the action on the context providers. Then use the path as a key to the action :)

PR #301, opened 2016-08-21
fix(computed): by default send object as props to computed

christianalfoni:
Fixes issue where not passing in props to computed gives "undefined" as the props. Should be empty object.


Guria:
@christianalfoni what happen with tests?

christianalfoni:
Wow, wut? Let me review

christianalfoni:
Okay. this is fixed now. The cache key changed as we pass empty object as default, so test was wrong, not code :)

PR #305, opened 2016-08-26
docs(async): warn about using spread operator on auto async

christianalfoni:
The warning did not really tell you the reason it happens. Now it does :) Been asked a couple of times about the warning and when I say what this update says, they get it right away. So hopefully this will make devs understand it instantly :)


Guria:
#306

PR #308, opened 2016-09-25
Cerebral2

christianalfoni:


Guria:
@christianalfoni rebased to latest master to remove conflict warnings

src/Controller.js
```diff
@@ -0,0 +1,123 @@
...
+        .map(provider => typeof provider === 'function' ? provider : ContextProvider(provider))
+    )
+
+    this.runTree = new FunctionTree(allProviders)
+    this.runTree.on('asyncFunction', this.flush)
+    this.runTree.on('end', this.flush)
+
+    if (this.devtools) {
+      this.devtools.init(this)
+      window.addEventListener('cerebral2.debugger.changeModel', (event) => {
```
Guria: Why it's done not in Devtools itself? It has reference to controller already.

src/Controller.js
```diff
@@ -0,0 +1,123 @@
...
+        this.model.set(event.detail.path, event.detail.value)
+        this.flush()
+      })
+    }
+
+    if (this.router) this.router.init()
+  }
+  /*
+    Whenever the the view needs to be updated this method is called.
```
Guria: "the" appears twice.

src/DefaultModel.js
```diff
@@ -0,0 +1,113 @@
+class Model {
+  constructor(initialState) {
+    this.state = initialState || {}
+    this.changedPaths = []
+  }
+  /*
+    Converts an array of paths changed to a change object that
+    will be traversed by the dependency store
+  */
+  flush() {
```
Guria: Will it be exposed to actions as state.flush? Is it safe?

src/DefaultModel.js
```diff
@@ -0,0 +1,113 @@
...
+
+    this.changedPaths = []
+
+    return changes
+  }
+  /*
+    A generic method for making a change to a path, used
+    by multiple mutation methods
+  */
+  updateIn(path, cb) {
```
Guria: Will it be exposed to actions as state.flush? Is it safe?

src/Module.js
```diff
@@ -0,0 +1,58 @@
+class Module {
+  constructor(path, module, parentModule) {
+    this.name = path.slice().pop()
+    this.path = path
+    this.controller = parentModule.controller
```
Guria: See a question about parentModule in Controller.js

src/Module.js
```diff
@@ -0,0 +1,58 @@
...
+    this.signals = moduleDescription.signals || {}
+    this.routes = moduleDescription.routes || {}
+    this.provider = moduleDescription.provider
+    this.registerModules(moduleDescription.modules || {})
+  }
+  /*
+    Returns an object of routes and their appointed signals
+    based on the module path
+  */
+  getRoutes() {
```
Guria: Very hard to read this part :)

src/devtools/index.js
```diff
@@ -0,0 +1,216 @@
+/*
+  Connects to the Cerebral debugger
+  - Triggers events with information from function tree execution
+  - Stores data related to time travel, if activated
+*/
+class Devtools {
+  constructor(options = {}) {
```
Guria:
```js
constructor(options = {storeMutations: true}) {
```
?

src/devtools/index.js
```diff
@@ -0,0 +1,216 @@
...
+    this.VERSION = 'v1'
+    this.storeMutations = options.storeMutations === undefined ? true : options.storeMutations
```
Guria: see above

src/devtools/index.js
```diff
@@ -0,0 +1,216 @@
...
+        console.warn('Cerebral Devtools - You tried to time travel, but you have turned of storing of mutations')
+      }
+      this.remember(event.detail)
+    })
+    window.addEventListener('cerebral2.debugger.pong', () => {
+      // When debugger responds to a ping
+      this.isConnected = true
+      this.sendInitial(initialModel)
+    })
+    window.addEventListener('cerebral2.debugger.ping', () => {
```
Guria: callbacks seems to be identical. maybe put it to method?

src/devtools/index.js
```diff
@@ -0,0 +1,216 @@
...
+    Send initial model. If model has already been stringified we reuse it. Any
+    backlogged executions will also be triggered
+  */
+  sendInitial(type, initialModel) {
+    const initEvent = new window.CustomEvent('cerebral2.client.message', {
+      detail: JSON.stringify({
+        type: 'init',
+        version: this.VERSION,
+        data: {
+          initialModel: this.initialModelString ? '$$INITIAL_MODEL$$' : initialModel,
```
Guria: move special strings to const

src/index.js
```diff
@@ -1,135 +1,5 @@
-var getByPath = require('./getByPath')
-var CreateSignalFactory = require('./CreateSignalFactory.js')
-var CreateRegisterModules = require('./CreateRegisterModules.js')
-var EventEmitter = require('events').EventEmitter
-var Computed = require('./Computed')
+import ControllerExport from './Controller'
```
Guria: export {default as Controller} from './Controller'

src/router/index.js
```diff
@@ -0,0 +1,128 @@
...
+  addressbar = {
+    pathname: '/',
+    value: '',
+    origin: '',
+    on() {},
+    removeListener() {}
+  }
+}
+
+export default function Router(options = {}) {
```
Guria: Might be better to use base router without default mapper. It will free router from knowing mapper special options (like query), and also will allow to bring breaking changes into mapper without need to bump cerebral major version. It will also shrink a build size for ones who uses own mapper implementation.

Guria: Could take this task

src/router/utils.js
```diff
@@ -0,0 +1,38 @@
+export function isObject(value) {
```
Guria: exist in top level utils

package.json
```diff
@@ -4,10 +4,13 @@
   "description": "A state controller with its own debugger",
   "main": "src/index.js",
   "scripts": {
-    "pretest": "standard",
-    "test": "istanbul cover nodeunit -- tests",
-    "test:watch": "nodeunit-watcher tests",
-    "semantic-release": "semantic-release pre && npm publish && semantic-release post"
+    "build": "webpack",
+    "prepublich": "npm run build",
```
christianalfoni: You fix that with the build setup?

src/Controller.js
```diff
@@ -0,0 +1,123 @@
...
+  and router is attached directly. Also a top level module is created.
+  The controller creates the function tree that will run all signals,
+  based on top level providers and providers defined in modules
+*/
+class Controller extends EventEmitter {
+  constructor({state = {}, routes = {}, signals = {}, providers = [], modules = {}, router, devtools}) {
+    super()
+    this.flush = this.flush.bind(this)
+    this.devtools = devtools
+    this.model = new DefaultModel({})
```
christianalfoni: Good question. I initially started that path, but as I moved on I could not see any reason to attach other models. If it becomes necessary at a later point it is easy to add it. Let me just rename it to model.

src/Controller.js
```diff
@@ -0,0 +1,123 @@
...
+    this.module = new Module([], () => ({
```
christianalfoni: You are perfectly right, let me fix that :)

src/Controller.js
```diff
@@ -0,0 +1,123 @@
...
+      state,
+      routes,
+      signals,
+      modules
+    }), {controller: this})
```
christianalfoni: Good point, let me fix that as well :)

package.json
Guria: ok

src/Controller.js
```diff
@@ -0,0 +1,123 @@
...
+    ).concat(
+      providers.concat(this.module.getProviders())
+        .map(provider => typeof provider === 'function' ? provider : ContextProvider(provider))
+    )
+
+    this.runTree = new FunctionTree(allProviders)
+    this.runTree.on('asyncFunction', this.flush)
+    this.runTree.on('end', this.flush)
+
+    if (this.devtools) {
```
christianalfoni: Hehe, I see your point. The thing is that we try to squeeze them into the "community modules" circle, when they in fact are core features of Cerebral... which are square... so, does not fit... kinda ;-)

For example the router needs to expose a method to components (convert signal to url). Not a community module requirement. The router now also extracts "routes" properties from all modules. Not a community module requirement.

Also squeezing these core features into the concept of a "community module" makes them behave differently, because they are special. For example the devtools has to be the first provider registered to correctly trigger a signal execution to the debugger, which means if it was a "community module" it would always have to be put first... which is a gotcha. Been at war with Redux because of ordering, causing so many weird issues.

What I mean to say is that the router and devtools are special core features that should be integrated to keep the public API as clean and simple as possible. Tweaking the public API to enable what these features need is not good I think.

Also what we are seeing now is that modules will not be such a common thing. `forms` will just be a set of helpers to define state and compose chains. `firebase` will become just a provider. `fuse` is a computed. `http` is overkill... reinventing the wheel. Now with `function-tree` you can just attach axios/request or whatever as a provider and you get the debugging experience and all out of the box.

On the new website I want to remove `community-modules` and just call it "Community tools" or whatever. They can be computeds (fuse), modules (useragent), providers (firebase), collection of factories and actions (forms), boilerplates etc.

src/Controller.js
```diff
@@ -0,0 +1,123 @@
...
+      this.devtools.init(this)
```
christianalfoni: We can have an `initialized` event that modules can listen to.

src/Controller.js
christianalfoni: Yeah, already noted that, will fix :)

src/Controller.js
christianalfoni: Ops, will fix!

src/DefaultModel.js
christianalfoni: Provider has a list of methods to expose, so will not be exposed

src/DefaultModel.js
christianalfoni: Provider has a list of methods to expose, so will not be exposed

src/Module.js
christianalfoni: Jup, will only pass controller here

src/Module.js
christianalfoni: Hehe, will rewrite it ;-)

src/devtools/index.js
christianalfoni: mmm, nice, will fix ;-)

src/devtools/index.js
christianalfoni: AH, yeah, good point, will fix

src/devtools/index.js
christianalfoni: good point! will fix

src/index.js
christianalfoni: Oh, nice :)

christianalfoni: will fix!

src/router/index.js
christianalfoni: Does that mean user will have to:

1. Install a second dependency, `url-mapper`
2. Intitilize with:

```js
Router({mapper: urlMapper({query: true)})
```

I think we should have a default router. What frameworks has multiple routers? :) If we want to add a different router that should be the whole router, not just an option. So:

```js
import Router from 'cerebral/router'

router: Router({query: true}) // etc.
```
or custom router:
```js
import OtherRouter from 'some-other-cerebral-router'

router: OtherRouter({query: true}) // etc.
```

Though technically it is cool to have different mappers etc., it makes it unnecessarily complicated for someone who just wants a router. I think the router that is integrated now should "just work" as part of Cerebral. If you want to do some edge case routing you just replace `cerebral/router` with something else.

src/router/utils.js
christianalfoni: Ops, will fix :)

src/Controller.js
bfitch: I had the same thoughts....providers via `function-tree` largely replace community-modules. And are  better in just about every way, IMO. More flexible, simpler api, less unique to cerebral...just including any library on context.

src/router/index.js
Guria: I see your point and even see a benefits router got as internal module. But I still want to decouple url-mapper for a few reasons. URLs is very important part of web application and must be a stable. It means even "bug fix" or little cosmetic change in mapper could lead to major problems for cerebral users. And it would be a very strange to bump cerebrals major version because of changes in optional module. And of course there is a great chance to miss this bump as well.
It's not a "might be" use case. I have postponed plans on url-mapper which also needs new major on URLON.

> If we want to add a different router that should be the whole router, not just an option. 

Disagreed. Changing mapper is most likely the only change needed. I know several users using router with custom mapper already (including @garth). `OtherRouter` would have copy pasted blocks from original one. It also means that special router module interface must be treated as public.

src/inferno/index.js
```diff
@@ -0,0 +1,5 @@
+import containerExport from './container'
```
Guria: export default as

src/react/index.js
```diff
@@ -0,0 +1,5 @@
+import containerExport from './container'
```
Guria: export default as

src/router/index.js
schotime: I think the way the urls being mapper would be something you might swap. urlon does some funkiness that sometimes you might want to avoid.

src/Controller.js
schotime: I'm assuming this is the mutable model? We ran into one issue using the mutable model, and maybe we can add an option which freezes it, or copies it once you retrieve it from the store. 

```
function({state}) {
 var user = state.get("user");
 user.name = "Modified";
}
```
and because user is a reference to an object it actually changes it in the store. Soo easy to make this mistake, so we have turn on the immutability one now so that if anyone does this accidentally it throws an error.

Thoughts?

src/router/index.js
Guria: @schotime already decided to publish router as separate package, but most likely it would live in main repo.

src/Controller.js
christianalfoni: Ah, yes, I have been thinking about this. With devtools I am thinking:

```js
Controller({
  devtools: Devtools({freezeState: true})
})
```

Unsure if it should be default though. Maybe it should be. And if you do not care about timeTravel and this freeze thing, you do:

```js
Controller({
  devtools: Devtools({freezeState: false, storeMutations: false})
})
```

I will officially put it on my list :)

schotime: Agreed. On by default. thx

src/cerebral.d.ts
```diff
@@ -0,0 +1,107 @@
+﻿declare module "cerebral" {
```
Guria: we shouldn't ship old style global modules definitions. I'll exclude this commit for now. Anyway thanks for initial typings. I'll prepare new setup based on this.

schotime: Not sure what you mean? Show me where they aren't like this anymore?

schotime: Not sure that's gonna work. We need to bundle them separately from cerebral in the @types repo. There will be different versions of them.

Guria: Publishing to @types is usually a fallback way for projects that doesn't provide typings right in npm package for any reason. Typings provided in npm package should be "import / export style", not "declare module style".
I think we should provide normal as-is typings right with cerebral2 and should be another package with your wrapping helpers and extended typings. That way it wouldn't require to install anything from @types

schotime: As a long as it'll work. I'm very sceptical that I'll be able to change the signature of a method by importing a types. Eg. Set on state provider is defined like

```
set(path: string, value: any)
```
Not

```
set<T>(path: T, value: T)
```

I want nothing to do with the standard types in my cerebral projects. No point using typescript then.

Guria: We can (should) declare both right in cerebral repo:
```
set<T>(path: string, value: T)
set(path: string, value: any)
```

src/viewFactories/Hoc.js
```diff
@@ -0,0 +1,116 @@
...
+        } else if (hasChange) {
+          this._update()
+        }
+      }
+      shouldComponentUpdate () {
+        // We only allow forced render by change of props passed
+        // or Container tells it to render
+        return false
+      }
+      componentWillUnmount () {
```
fmal: Just a small typo. Should probably be `isUnmounting` :smiley:

Guria:
@christianalfoni had to comment few test in scheme-parser. Feel free to fix it :)
Also function-tree now could be rewritten using all es6 goodies.

PR #313, opened 2016-10-04
perf(computed): check cached value as soon as possible.

gaspard:
The check in computed factory should happen before resolving paths or depsMap.

### DIFF
--- a/src/Computed.js
+++ b/src/Computed.js
@@ -86,12 +86,10 @@ export class Computed {
 class ComputedFactory {
   constructor (paths, func) {
     if (
-      !paths ||
-      (
-        !isObject(paths) &&
-        typeof paths !== 'function'
+      !(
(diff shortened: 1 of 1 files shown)


christianalfoni:
Hi there!

Thanks for looking into this. The reason we grab those things after the `for` loop is because the for loop might result in a function `return`. Basically calling `getDepsMap` and `this.paths` is not needed. Tiny optimization that hurts readability a bit.

Or did you meet a bug with this behaviour?

gaspard:
Actually the change is to move paths/depsMap to after the loop, they were run before. A tiny performance change but it also improves readability to have cache resolution before anything else in the function (makes clear what the cache resolution depends on)...

christianalfoni:
Haha, it is? That says something about my ability to read git change logs :-D

PR #320, opened 2016-10-12
Cerebral2: getSignals

schotime:
A way to have all the signals placed on props in by using connect as was available in Cerebral 1.
We can discuss if there is a way we can optionally turn this on, but here is the implementation.
This is extremely useful to have when using TypeScript.

### DIFF
--- a/packages/cerebral/src/Controller.js
+++ b/packages/cerebral/src/Controller.js
@@ -58,6 +58,8 @@ class Controller extends EventEmitter {
 
     if (this.router) this.router.init()
 
+    this.signalsClass = new Signals(this)
+
     this.emit('initialized')
   }
   /*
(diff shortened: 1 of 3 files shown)


Guria:
@schotime what iabout to make prefetchSignals to be lazy loaded on first getSignals call

@christianalfoni what do you say about this api addition at all

schotime:
@guria that's a good idea. Will update.

packages/cerebral/src/Controller.js
```diff
@@ -130,7 +132,35 @@ class Controller extends EventEmitter {
...
 
     return this.runSignal.bind(this, path, signal)
+  } 
+
+  getSignals () {
+    return this.allSignals;
+  }
+}
+
+function prefetchSignals(controller) {
```
Guria: just put into a class and get controller from `this`

schotime:
@Guria updated

packages/cerebral/src/Controller.js
```diff
@@ -130,6 +130,34 @@ class Controller extends EventEmitter {
...
+
+  getSignals() {
+    const controller = this
+    function extractSig(path, signals) {
+      return Object.keys(signals).reduce((sigs, key) => {
+        sigs[key] = controller.runSignal.bind(controller, `${path}.${key}`, signals[key])
+        return sigs
+      }, {});
+    }    
+    function getModuleSignals(path, module) {
```
Guria: Hm, now this functions recreated each getSignals call. But it used only once.
And i just remembered that we wanted to allow delayed module registration.
Lets don't hurry now and wait @christianalfoni answer

christianalfoni:
@schotime, @Guria and I have discussed a bit. Instead of extending the code itself, we are thinking about a hook for extending it. Since everything is "single class instance based" these instances can easily be extended. So in your scenario you want to extend the `Controller` and you want to extend the `HOC`. So one way to do this would be with extension hooks:

```js
Controller({
  extend: {
    controller: [getSignals],
    hoc: [signalsProp]
  }
})
```

`getSignals` and `signalsProp` is simply objects (or function passed the instance). These objects will be merged into the instance of, in this example, the controller and hoc instances. So for example here `getSignals` would be something like:

```js
{
  getSignals() {
    this // controller instance
  }
}
```

And `hoc`, using a function:

```js
(hoc) => {
  const originalGetProps = hoc.getProps
  hoc.getProps = (...args) => {
    const props = originalGetProps.apply(hoc, ...args)
    props.signals = ...

    return props
  }
}
```

Something like that. Would you be up for creating the first "extensions" for this? And if you have any ideas on how this should work differently etc, lets discuss it more :)

christianalfoni:
Btw, @schotime, I was also thinking about challenging you to think of a more `typescript` friendly API by extending. So for example:

```js
Controller({
  extend: typescript({
    // Other extensions
  })
})
```

So this typescript extension can change the API to work much more the way typescript wants things to work. Anyways... just a shower thought :)

Guria:
@christianalfoni I saw a suggestion to implement getSignals in a devtools under a flag. Should we close this PR?

PR #325, opened 2016-10-15
feat(DependencyStore): introduce strict render as option

christianalfoni:
- Added new option:

```js
Controller({
  strictRender: true // False by default
})
```

- Dependency store now has two ways to extract components/computed to render/cachebust based on the change map

### DIFF
--- a/packages/cerebral/src/Controller.js
+++ b/packages/cerebral/src/Controller.js
@@ -16,8 +16,9 @@ import {dependencyStore as computedDependencyStore} from './Computed'
   based on top level providers and providers defined in modules
 */
 class Controller extends EventEmitter {
-  constructor ({state = {}, routes = {}, signals = {}, providers = [], modules = {}, router, devtools = null}) {
+  constructor ({state = {}, routes = {}, signals = {}, providers = [], modules = {}, router, devtools = null, strictRender = false}) {
     super()
+    this.strictRender = strictRender
     this.flush = this.flush.bind(this)
     this.devtools = devtools
     this.model = new Model({}, this.devtools)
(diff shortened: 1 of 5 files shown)


Guria:
Lint failed

packages/cerebral/src/operators/debounce.js
```diff
@@ -1,3 +1,3 @@
-import debounce from 'function-tree/lib/factories/debounce'
+import debounce from 'function-tree/factories/debounce'
```
Guria: nope, you shouldn't change it

packages/cerebral/src/Controller.js
```diff
@@ -67,7 +68,12 @@ class Controller extends EventEmitter {
   */
   flush () {
     const changes = this.model.flush()
-    const computedsAboutToBecomeDirty = computedDependencyStore.getUniqueEntities(changes)
+    const computedsAboutToBecomeDirty = (
+      this.strictRender ?
```
Guria: lint rule failed here

packages/cerebral/src/operators/debounce.js
Guria: use `lerna build` to get all packages compiled

christianalfoni: It was due to tests, will tests work when everything is built?

Guria: Yes it will. Moreover tests run with `lerna run test` builds all packages before running tests.
And  I can't see how 'function-tree/factories/debounce' path is worked for you.

christianalfoni: Yeah, it does :) Okay, getting the hang of it now!

packages/cerebral/src/Controller.js
```diff
@@ -67,7 +68,7 @@ class Controller extends EventEmitter {
   */
   flush () {
     const changes = this.model.flush()
-    const computedsAboutToBecomeDirty = computedDependencyStore.getUniqueEntities(changes)
+    const computedsAboutToBecomeDirty = this.strictRender ? computedDependencyStore.getStrictUniqueEntities(changes) : computedDependencyStore.getUniqueEntities(changes)
```
Guria: standard allows multiline here, just question mark and colon should start a new line

christianalfoni: Yeah, though do not like that format... hehe... just keep as it is now? :-) It just reads weird to me that:

```js
this.strictRender
? ...
``` 

Cause the first line does not indicate anything, it looks like a syntax error, compared to:

```js
this.strictRender ?
  ...
```

It just indicates what follows :) Anyways, opinions... just keep as is now?

PR #326, opened 2016-10-15
feat(cerebral-router): add routes config as option to router

christianalfoni:
I have added the routes config as an option to the router itself. I have kept the "routes by modules" to allow some more testing. Let us conclude after some discussion on it :)

### DIFF
--- a/packages/cerebral-router/src/index.js
+++ b/packages/cerebral-router/src/index.js
@@ -33,12 +33,12 @@ export function goTo (url) {
 export default function Router (options = {}) {
   options.mapper = urlMapper({query: options.query})
 
-  return (passedRoutesConfig, controller) => {
+  return (controller) => {
     if (!options.mapper || typeof options.mapper.map !== 'function') {
       throw new Error('Cerebral router - mapper option must be provided.')
     }
 
-    const routesConfig = flattenConfig(passedRoutesConfig)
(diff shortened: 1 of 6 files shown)


packages/cerebral-router/src/index.js
```diff
@@ -38,7 +38,7 @@ export default function Router (options = {}) {
       throw new Error('Cerebral router - mapper option must be provided.')
     }
 
-    const routesConfig = flattenConfig(passedRoutesConfig)
+    const routesConfig = options.routes || flattenConfig(passedRoutesConfig)
```
Guria: Wouldn't it be frustrating that modules level config would be ignored when also routes passed to router itself?
I still feel sceptic about allowing modules level routes by default. It ok as extension but too opinionated for default

Guria:
in #320 we suggested controller extension api.
my suggestion is to reimplement this special `router` field on modules as extension and doesn't ship it by default

packages/cerebral-router/src/index.js
christianalfoni: Ah, okay, let me fix flattening

christianalfoni:
Ah, yeah, that is a great idea :) Let me rebase this and remove the module routing stuff

christianalfoni:
Okay, I have removed module routes config

PR #333, opened 2016-10-18
Extend

christianalfoni:
Added ability to extend controller, hoc and container. Please check tests for API (will update website if successful feature). @schotime is looking into testing it :)

### DIFF
--- a/packages/cerebral/src/Controller.js
+++ b/packages/cerebral/src/Controller.js
@@ -1,7 +1,7 @@
 import FunctionTree from 'function-tree'
 import Module from './Module'
 import Model from './Model'
-import {ensurePath, isDeveloping, throwError, isSerializable} from './utils'
+import {ensurePath, isDeveloping, throwError, isSerializable, addExtensions} from './utils'
 import VerifyInputProvider from './providers/VerifyInput'
 import StateProvider from './providers/State'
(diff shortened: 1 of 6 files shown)


Guria:
@christianalfoni I don't like this api. Would be hard to combine several extensions into groups.
Eg @schotime case requires to extend controller and hoc for one exact purpose and hoc extensions depends on controller extension.
With suggested api it requires to define it separately.
my suggestion that it should work like:
```js
Controller({
  plugins: [devtools(), router(), exposeSignals()]
})
```
so all extensions combined by purpose.

christianalfoni:
Hm, though it is a bit more complicated here as we have multiple places to extend. Also the HOC is not just a class, it is a function, returning a function, returning the HOC :)

Also a matter of ease of use. It is easier to just give this extend API, rather than explaining how to extend all these different prototypes.

christianalfoni:
How do they work? the plugins? what is the syntax inside them?

christianalfoni:
Btw, extending all with extend is quite simple:

```js
Controller({
  extend: typescript({
    controller: [],
    hoc: [],
    container: []
  })
})
```

christianalfoni:
Lets us close this, as we have the `signalsProp` option instead. Can revisit this later

PR #334, opened 2016-10-18
feat(Operators): implement template tag based operators

christianalfoni:
Based on your idea @Guria I implemented template tag based operators. I think they are awesome for these reasons:

1. It is pure JS
2. A lot less code (no deps)
3. Implementing operators is a breeze, lot less code there as well
4. We get nesting for free

Ready for review and feedback :)


Guria:
Looks great.

Guria:
scheme-parser could be dropped then at all? right?

PR #344, opened 2016-10-23
style(Operators): change parameter name to clarify meaning

gaspard:


Guria:
merged as part of #345 
github had failed to detect it because it was altered commits during rebase and merge action

PR #356, opened 2016-10-25
fix(function-tree): pass final resulting payload to end event

maxfrigge:
Pass the resulting payload to the end event - before the initial payload was passed instead.

### DIFF
--- a/packages/function-tree/src/index.js
+++ b/packages/function-tree/src/index.js
@@ -211,9 +211,9 @@ class FunctionTree extends EventEmitter {
     })
 
     this.emit('start', execution, payload)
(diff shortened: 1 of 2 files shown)


packages/function-tree/src/index.js
```diff
@@ -211,9 +211,9 @@ class FunctionTree extends EventEmitter {
     })
 
     this.emit('start', execution, payload)
-    executeTree(execution.staticTree, execution.runFunction, payload, () => {
-      this.emit('end', execution, payload)
-      cb && cb(null, execution, payload)
+    executeTree(execution.staticTree, execution.runFunction, payload, (result) => {
```
maxfrigge: @christianalfoni Do you mean rename this to `finalPayload`?

Issue #318, opened 2016-10-11
Investigate possibility to provide codemods


Guria:
assigned to @jeroenverfallie

NervosaX:
https://github.com/facebook/jscodeshift Perhaps this is useful?

Guria:
Exactly this tool was meant by "codemods". :)
But haven't heared any news from @jeroenverfallie on this task yet :(

NervosaX:
Played around for an hour trying to understand how jscodeshift works. 

Managed to create a codemod that converts the new controller code from here: http://cerebral.github.io/cerebral-website/migration/01_start_migration.html

https://github.com/NervosaX/cerebral-codemods

Might try and find the time to continue with the rest. Very interesting tool. :) 

Some useful documentation:
https://www.toptal.com/javascript/write-code-to-rewrite-your-code

Absolutely required tooling:
https://astexplorer.net/

Guria:
@NervosaX would you like to continue explore on this task? We haven't heard anything from @jeroenferfallie

NervosaX:
@Guria Sure thing. Based on the migration guide, the hardest (or possibly impossible part) to code mod will be actions and providers.

Are we considering some sort of migration path between what we have and what we're moving to? Are promises expected, or does async = true still exist in cerebral 2.x? 

My only issue is I need some catchup on the full implications of the changes.

NervosaX:
This AST Type definition files are also required when writing these (just noting this for others that may come across this).

https://github.com/benjamn/ast-types/tree/master/def

christianalfoni:
Just closing this as codemods has been created?

Guria:
It is still need to be merged

NervosaX:
Work could continue in the repo, if we're happy with its direction. Probably best I finish off the latest changes and docs however.

Guria:
@NervosaX yeah, lets update and merge it. We can iterate on it later.

edgesoft:
@Guria Can this be closed?

Guria:
Yes

PR #362, opened 2016-10-26
feat(Computed): simplify use of computed without props

gaspard:
```js
connect({
  foo: someComputed,
  foo2: someComputed.props({foo: 'bar'})
})
```

```js
function someAction({state}) {
  state.compute(someComputed)
  state.compute(someComputed.props({foo: 'bar'})
}
```

BREAKING CHANGE:

Use "myComputed" instead of "myComputed()"

### DIFF
--- a/packages/cerebral-demo/src/components/UpperTitle/index.js
+++ b/packages/cerebral-demo/src/components/UpperTitle/index.js
@@ -3,7 +3,7 @@ import {connect} from 'cerebral/react'
 import upperCaseTitleComputed from '../../computeds/upperCaseTitle'
 
 export default connect({
-  title: upperCaseTitleComputed()
+  title: upperCaseTitleComputed
 },
   function UpperTitle (props) {
     return (
--- a/packages/cerebral-todomvc/src/components/App/index.js
+++ b/packages/cerebral-todomvc/src/components/App/index.js
@@ -8,7 +8,7 @@ import visibleTodosRefs from '../../computed/visibleTodosRefs'
 export default connect({
   todos: 'app.todos',
   isSaving: 'app.isSaving',
-  visibleTodosRefs: visibleTodosRefs()
+  visibleTodosRefs: visibleTodosRefs
 },
   function App (props) {
     return (
--- a/packages/cerebral-todomvc/src/components/Footer/index.js
+++ b/packages/cerebral-todomvc/src/components/Footer/index.js
@@ -5,7 +5,7 @@ import cn from 'classnames'
 
 export default connect({
   filter: 'app.filter',
-  counts: counts()
+  counts: counts
 }, {
   filterClicked: 'app.filterClicked',
   clearCompletedClicked: 'app.clearCompletedClicked'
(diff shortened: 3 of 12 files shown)


NervosaX:
They seemed fine before. What's the reason for the change?

Every breaking change is another reason upgrading to the next cerebral takes longer. Can handle both working? Or create a deprecation?

If there's a good reason, I'm not concerned. I just want to make sure we're not causing needless work to upgrade.

gaspard:
As I understand it, the 2.0 release is the opportunity to clarify and sanitize API to make it easier for newcomers to get started and to reduce confusion. Not returning a factory as in 1.0 reduces the level of complexity. The goal is to have an API with least surprises and gotchas.

gaspard:
@NervosaX There are lots fo breaking changes in 2.0 especially from pre 1.x (no more Baobab model, no immutable model, no support for snabbdom, totally different operators, etc). Could you give more details as to why this specific change does not seem valid compared to other changes ?

For reference: [list of changes from 1.x to 2.x](http://cerebral.github.io/cerebral-website/migration/01_start_migration.html)

NervosaX:
Nope, simplicity is a good goal. I just would think handling the current method of computed functions would be just as easy to handle without forcing another upgrade collision.

henri-hulski:
@NervosaX  I think in this case the upgrade is just a simple search and replace maybe with some regex.
We could even provide an upgrade script, which automates the upgrade as far as possible.

NervosaX:
@henri-hulski: Agreed.

 I'm not attacking this change in particular, there's nothing wrong with it. I've been using cerebral for a good 8 months or so now, and upgrading has been a huge pain point (which is my problem, having used it before a 1.0 release).

I think an upgrade script is a great way to go about this.

gaspard:
I thought about keeping the old method along the new one (by adding 'getValue' to the factory) but the only solutions I could come up with either involved passing the factory in places where a computed is expected and this did not look ideal or involved messing with the prototype (and this is also quite ugly). If there is a not too complex way to make an object (the computed) callable so that the old way still works we could use it... Any idea here very welcome... It could be used for deprecation without breaking (like in 1.99)...

gaspard:
@NervosaX I also have a big app and the changes are hard (am waiting for 2.x, skipping 1.x altogether). We definitely need a good migration guide and some helper scripts if someone knows how to make them (and good demo apps I think). In my case, I also have to move from immutable+Snabbdom to React/Inferno (haven't decided yet) and Cerebral model. It's a pain but I really prefer how 2.x is becoming simpler and focussed.

I think there is a hangout planed.. Not sure. This should be discussed for sure.

edgesoft:
@gaspard @NervosaX https://github.com/cerebral/cerebral/issues/318

Guria:
I'll merge it at weekend then, if no more objections would be posted here.

Guria:
I have also planned to make alpha release of all the things after this merged

gaspard:
I will try to release the new operators tonight... Still got this question
on merge..

Guria:
rebasing and merging now

PR #371, opened 2016-10-28
feat(Operators): add more operators to be orthogonal with model methods

gaspard:
```
concat, merge, pop, push, shift, splice, unset, unshift
```

I know the operators.test file is too big but since each operator has only one or two tests, I did not feel like adding tons of files for this. Could be changed if anyone thinks it is better to have a test file per operator.


Guria:
@gaspard need rebase and fix of linting errors

gaspard:
Sorry. I fixed the linting errors but did not push them correctly in the commit. Should be ok now.

Guria:
@gaspard most operators looks pretty same. Maybe we can use factory here?

gaspard:
Yes, that's a good idea. You can do it if you like our I'll do this
tomorrow morning.

gaspard:
I forgot to export them from `index`. Doing it now.

gaspard:
I updated todomvc with new operators, found a bug in `toggle`, added a test and fixed it. Also using `import {set, foo, ..} from 'cerebral/operators'` in tests (same as usage, catches missing exports in index.js).

gaspard:
I am splitting operator tests in separate files (500 loc is too long). This also give better incentive to properly cover each operator with tests (bad argument tests for example).

christianalfoni:
Looking good! Looking good! 👍

PR #372, opened 2016-10-30
feat(Providers): add Recorder provider

christianalfoni:
New Recorder provider which records mutations, not signals like the old one. Implementation is simplified and can be later connected to debugger to activate recording from debugger. Will implement demo of this when todomvc is brought into the repo.


Guria:
Todomvc is already part of the repo

christianalfoni:
@Guria I noticed on `cerebral-demo` I had to `npm run setup` on every code change of other packages... is that necessary?

Guria:
You can run only `npm run build` in changed package. `npm run setup` justmakes you sure that's everything synced

packages/cerebral/src/providers/Recorder.js
```diff
@@ -0,0 +1,205 @@
+/*
+  - Wrap mutation methods and track their usage
+  -
```
Guria: empty line

PR #380, opened 2016-10-31
fix: flush on functionEnd, needed because state changes in path-chain…

fopsdev:
… did not work

eg.
```
...
GetData,
        {
          success: [
            set(state`data`, input`result`),
...
```
-> Component depending on state.data did not rerender

### DIFF
--- a/packages/cerebral/src/Controller.js
+++ b/packages/cerebral/src/Controller.js
@@ -51,6 +51,7 @@ class Controller extends EventEmitter {
 
     this.runTree = new FunctionTree(allProviders)
     this.runTree.on('asyncFunction', () => this.flush())
+    this.runTree.on('functionEnd', () => this.flush())
     this.runTree.on('end', () => this.flush())
 
     if (this.devtools) {


christianalfoni:
Ah, yeah... hm... we can not do it on `functionEnd` though... because then it will flush after every action, which it does not need to. It only needs to flush after a set of synchronous actions has run.

But yeah, we need to identify this better. Maybe we need a `pathStart` and `pathEnd` event... these trigger whenever function tree goes down a path or ends at a path.

It would still be an issue though, if:

```js
[
  doSomething, {
    success: [doThis]
  },
  doThat
]
```

It would trigger both after `doThis` and after `doThat`, though in this case it only needed to trigger after `doThat`, as `doThis` and `doThat` triggers sync.

Hm need to think on this, let @Guria chime in as well :)

Guria:
Hmm, looked bit closer, but didn't get what in @fopsdev example cause a fail. Why current strategy (flush before async and at the end) has failed?

fopsdev:
repro:
flush

Guria:
Is problematic code could be found on tutorial branch?

fopsdev:
yep

christianalfoni:
Hm... we have to look at this some more. Cause your are saying it did not update? It should have updated no matter what, so must be something else here. Do you have some more code here @fopsdev ? It seems more like the signal never "ended" for some reason.

christianalfoni:
Ah, wait, I think I see it... you are waiting as the last action on the chain. So the "ended" does not happen until 2 seconds has passed. So it should update, it just happens after 2 seconds. 

```js
[
  doSomething, {
     success: [setSomething]
  },
  wait(2000)
]
```

The flush will not happen until the 2 seconds has happened at the end there.

christianalfoni:
Hm, I wonder if this is because you have forgotten to spread:

`showToast(`Total Stars @C{getStars}`, 3000)`

This is async and returns an array, so should be `...showToast(`Total Stars @C{getStars}`, 3000)`.

I think this is an extreme edge case, but good to know. We should still maybe do this fix because:

```js
[
  [async1000, {...}], // This output does not cause flush
  [async5000, {...}] // This output does not cause flush either
] // It will flush here, after 5 seconds, though the 1000 second path should have flushed
```

Guria:
@fopsdev could you check out latest master?

Guria:
fixed by #384

fopsdev:
@christianalfoni @Guria ok, works now as expected, thanks guys!

PR #382, opened 2016-11-01
fix(Model): fix unset to flush the path changes up to the removed key

gaspard:
```js
model.unset(['foo', 'bar'])
```

Should notify changes like

```js
{foo: {bar: true}}
```

not

```js
{foo: true}
```

### DIFF
--- a/packages/cerebral/src/Model.js
+++ b/packages/cerebral/src/Model.js
@@ -205,6 +205,7 @@ class Model {
     })
   }
   unset (path) {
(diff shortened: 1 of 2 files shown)


henri-hulski:
@Guria This fixed the bug in TodoMVC!
It was also a problem with unset.
It worked in strict mode but not in non strict.
With this fix it works in both.

henri-hulski:
@gaspard  Good find! Thanks for the fix.

PR #384, opened 2016-11-02
feat(function-tree): trigger pathStart and pathEnd events

christianalfoni:
Related to https://github.com/cerebral/cerebral/pull/380, we need to trigger a `pathEnd` event in function-tree to handle ended path executions that does not cause flush event due to pending async parallel action


packages/function-tree/README.md
```diff
@@ -464,8 +464,11 @@ execute.on('error', (error, execution, payload) => {})
 // When a function tree is executed
 execute.on('start', (execution, payload) => {})
 
-// When a function tree ends its execution
-execute.on('end', (execution, payload) => {})
```
henri-hulski: @christianalfoni Why is the end event removed from the docs? 
Is it no public API anymore?

christianalfoni: Removed?

christianalfoni: wow, what happened there

PR #387, opened 2016-11-03
feat(Views): add ability to inject props to component

christianalfoni:
Added new signature:

```js
connect({/*paths*/}, {/*signals*/}, {/*props*/}, () => {})
```

Allows you to add arbitrary props passed into component.

### DIFF
--- a/packages/cerebral/src/inferno/connect.js
+++ b/packages/cerebral/src/inferno/connect.js
@@ -1,19 +1,23 @@
 import HOC from './hoc'
 
-export default function connect (paths, passedSignals, passedComponent) {
+export default function connect (paths, passedSignals, injectedProps, passedComponent) {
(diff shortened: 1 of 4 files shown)


gaspard:
What is the difference between this implementation and having fixed props in the file ?

```js
const fixedProps = //...

export default connect //...
```

christianalfoni:
It is to make it dynamic :) So you can pass different props to diferene instances of the same component.

gaspard:
So it is used like this ? Makes sense but maybe we should have a test with dynamic props to better show the use case don't you think ? :-)

```js
connect(
  { /* paths */ },
  { /* signals */ },
  props => ({ /* dynamic props */ }),
  function MyThing () {
    // ...
  }
)
```

PR #392, opened 2016-11-06
feat(Views): always return wrapped component even in test env

Guria:
cc @christianalfoni @garth 

Returning an undecorated component in test environment doesn't allows us to have e2e tests.

Now we have support for testing with dummy controller:
```js
import React from 'react'
import {mount} from 'enzyme'
import {Container} from 'cerebral/react'

import Foo from './Foo'

describe('<Foo />', () => {
  it('allows us to set props', () => {
    const state = {
      foo: 'bar'
    }
    const wrapper = mount(
      <Container state={state}>
        <Foo />
      </Container>
    )
    expect(wrapper.find('.foo')).to.have.length(1)
  })
})
```

And if you want to test undecorated component you can just have separate export for it.

Another reason is that Computed now also doesn't have this conditional return.


garth:
Yes, I saw the changes. Seems like a good thing mostly. I think that it could still be useful to have some helper lib to do the test controller setup for signal/module testing.

Guria:
So do you approve this pr?

PR #394, opened 2016-11-06
feat(Devtools): dependencyStore usage refactored

### LINKED ISSUE #391: must work with multiple Containers on page
```js
ReacDOM.render((
  <Container controller={controller} >
    <App />
  </Container>
), document.querySelector('#root'))
ReacDOM.render((
  <Container controller={controller} >
    <Overlay />
  </Container>
), document.querySelector('#overlay'))
```

christianalfoni:
- components dependency store now in controller (allows multiple Container)
- Debugger logic for components moved into devtools
- Only verifyStrict on components as computeds are attached to components, no need for double check
- Fixed issue with merge, missing functionality (allowing template tags on merged prop values)
- Added `options` to controller for `strictRender`, more options coming

fixes #391

### DIFF
--- a/packages/cerebral/src/Controller.js
+++ b/packages/cerebral/src/Controller.js
@@ -1,3 +1,4 @@
+import DependencyStore from './DependencyStore'
 import FunctionTree from 'function-tree'
 import Module from './Module'
 import Model from './Model'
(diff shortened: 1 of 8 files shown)


packages/cerebral/src/Controller.js
```diff
@@ -16,9 +17,11 @@ import {dependencyStore as computedDependencyStore} from './Computed'
   based on top level providers and providers defined in modules
 */
 class Controller extends EventEmitter {
-  constructor ({state = {}, signals = {}, providers = [], modules = {}, router, devtools = null, strictRender = false}) {
+  constructor ({state = {}, signals = {}, providers = [], modules = {}, router, devtools = null, options = {}}) {
     super()
-    this.strictRender = strictRender
+    this.computedDependencyStore = computedDependencyStore
```
Guria: Let move instantiating of `computedDependencyStore` here, because now it seems not safe to have few controllers in one app since all of computeds will share one instance.
Is it possible to share computedDependencyStore instance from model and hoc?

Guria:
Great. I like how HOC's code was cleaned. 👍

packages/cerebral/src/Controller.js
christianalfoni: Hm, issue here is that computed does not know anything about controllers. There is no "create instance of computed"... cause we do: `import {Computed} from 'cerebral'`.

We would have to change this to: `import {Computed} from './controller'`

Sharing computed store across controllers is no problem though. You might get into an extreme edge case where one controller flags the computed used in a different controller... cause they for some reason have the exact same state path. But this is not a real issue.

So suggest that we keep "as is"... cause this is not a real issue and very likely will never be ;-)

Issue #398, opened 2016-11-08
import website


Guria:
* import cerebral-website v3 branch as 'packages/docs'
* create 'docs/README.md' with link to 'packages/docs/content'
* cleanup cerebral-website repo
* update cerebral-website readme with links to monorepo and cerebral-website v2 branch
* move cerebral-website to legacy org

I will do CI part myself

henri-hulski:
@Guria 
> create 'docs/README.md' with link to 'packages/docs/content'

Do you mean in the root directory or change 'packages/docs/README.md'?

Guria:
@henri-hulski it is smth like https://github.com/facebookincubator/create-react-app/tree/master/template

Guria:
so: https://github.com/cerebral/cerebral/tree/master/docs/README.md should point to https://github.com/cerebral/cerebral/tree/master/packages/docs/content

Guria:
Great work as always. Thanks

PR #408, opened 2016-11-09
refactor(website): integrate cerebral-website into monorepo

henri-hulski:
Related to #398


Guria:
hmm, somehow you have imported previous website as well. There shouldn't be app and server folders.

Guria:
v3 branch contains only 93 commits and it doesn' contains previous website at all. but this pr contains 250+ commits including previous website.

henri-hulski:
Yeah I have merged it in master first. 
So the branch is self contained right? 
I will do it again without merging. 

I think I did something similar with cerebral-module-useragent. I hope there it's ok.

Guria:
No problems with useragent. There was branches with shared parents.
But website v3 started as orphan branch.
Also don't forget to include lint fixes

henri-hulski:
@Guria Should be fixed now.

Issue #410, opened 2016-11-09
useragent: error when using offline or window functionality


fopsdev:
When using newest useragent on newest cerebral2 i'm running into the following issue(console) whenever i execute a http.post or the window resizes (i did a console log inside controller.js to see which signal is running):
```
useragent.windowChanged
Controller.js:269 [function, function]
Controller.js:270 Object {windowSpec: Object, media: Object}
populatePath.js:14 Uncaught TypeError: valueTemplate is not a function
```

fopsdev:
could be a generic issue with cerebral as well

fopsdev:
found the issue:
```
set(state`${path.join('.')}.media`, input`media`),
```
fails because the valueTemplate should maybe not contain ${} - expressions
if i hardcode it like:
```
set(state`useragent.media`, input`media`),
```
then it works

PR #411, opened 2016-11-10
fix: populatePaths  

fopsdev:
fixes https://github.com/cerebral/cerebral/issues/410


packages/cerebral/src/operators/helpers/populatePath.js
```diff
@@ -5,6 +5,10 @@ export default function populatePath (context, strings, values) {
       return currentPath + string + context.state.compute(valueTemplate)
     }
 
-    return currentPath + string + (valueTemplate ? valueTemplate(context).toValue() : '')
+    if (valueTemplate && typeof valueTemplate === 'function') {
+      return currentPath + string + valueTemplate(context).toValue()
+    }
+
+    return currentPath + string + (valueTemplate ? valueTemplate : '')
```
Guria: > standard: Use JavaScript Standard Style (http://standardjs.com)
>   /home/travis/build/cerebral/cerebral/packages/cerebral/src/operators/helpers/populatePath.js:12:52: Unnecessary use of conditional expression for default assignment.

Use `valueTemplate || ''`

fopsdev: ok thanks,  i've used ... + (valueTemplate || '') now. (in () because of + gets executed before || and we just wanna make sure that valueTemplate has a value)
need bigger linting symbols ;)

PR #415, opened 2016-11-10
feat(tutorial): move tutorial and improve scripts

Guria:


Guria:
now it works like `npm run start:ch 01`

packages/tutorial/.gitignore
```diff
@@ -0,0 +1,2 @@
+public/index.html
+src
```
henri-hulski: @Guria Should it not be /src?
Like this all `parts/xx/src/` are excluded from github.

henri-hulski: Ok.

PR #444, opened 2016-11-14
feat(firebase): add sendPasswordRestEmail funtionality

bannerintheUK:


packages/cerebral-provider-firebase/src/sendPasswordRestEmail.js
```diff
@@ -0,0 +1,14 @@
+import firebase from 'firebase'
+
+export default function sendPasswordResetEmail (email) {
```
gaspard: File name does not match function name. "sendPasswordResetEmail" (missing "e" in Reset)...

gaspard: Maybe the function also needs to be declared in index.js ?

https://github.com/cerebral/cerebral/blob/master/packages/cerebral-provider-firebase/src/index.js#L66

PR #446, opened 2016-11-15
Fix controller init

gaspard:
Initial model state should be "no changes" or on first action run, the changes are checked and debugger may wrongly throw "you are changing 'foo' in strict mode". Problem is module initial state set.


gaspard:
Fixed typo in test. Should be {} instead of [].

PR #447, opened 2016-11-15
fix(debugger): fix model update issue

christianalfoni:


Guria:
Your merge conflict fix is wrong since you haven't used my corrections to make lint pass.

PR #451, opened 2016-11-16
chore(docs): restructure tutorial

Guria:


gaspard:
Maybe we could make chapter titles as # links (so that we can link there). Also keeping the numbers in chapters would help the feeling of "step 1, step 2". If possible, I would also add some margin above each of these chapter titles to help "breath" through the text:

```
01 Define State
...
//
// More space here
//
02 Render State
...
```

Guria:
@gaspard 
* header links is planned
* will try numbers
* will add more space before subsequent h2

PR #454, opened 2016-11-16
Tutorial feedback1

christianalfoni:
Added inquirer to load chapters, also updated docs

### DIFF
--- a/packages/tutorial/scripts/run.js
+++ b/packages/tutorial/scripts/run.js
@@ -1,0 +1,49 @@
+/* global rm, cp, sed */
+var path = require('path')
+var paths = require('./paths')
+var inquirer = require('inquirer')
+var exec = require('child_process').exec
+var tutorial = exec('react-scripts start')
+var runningInquirer = false
+
+function prepare (part) {
+  rm('-rf', paths.src)
+  cp('-R', path.join(paths.parts, part, 'src'), paths.root)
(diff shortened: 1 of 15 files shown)


packages/tutorial/scripts/run.js
```diff
@@ -0,0 +1,49 @@
+/* global rm, cp, sed */
+var path = require('path')
+var paths = require('./paths')
+var inquirer = require('inquirer')
+var exec = require('child_process').exec
+var tutorial = exec('react-scripts start')
+var runningInquirer = false
+
+function prepare (part) {
```
Guria: you can just require it from prepare.js?

packages/tutorial/scripts/run.js
```diff
@@ -0,0 +1,49 @@
...
+  cp(paths.partsIndex, paths.public)
+  sed('-i', '%N%', part, paths.publicIndex)
+}
+
+function chooseChapter (chapter) {
+  inquirer.prompt([{
+    type: 'list',
+    name: 'chapter',
+    message: 'Choose next chapter',
+    choices: ['01', '02', '03', '04', '05', '06', '07', '08', '09', '10', '11'],
```
Guria:
```js
ls('-d', paths.parts + '/*')
  .filter(function (item) { return test('-d', item) })
  .map(function (dir) { return dir.split('/').slice(-1)[0] })
```

PR #462, opened 2016-11-19
feat(forms): add requiredMessage, defaultRequiredMessage and custom global props

henri-hulski:
BREAKING CHANGE:

Rename 'errorMessages' to 'validationMessages'


packages/cerebral-forms/README.md
```diff
@@ -62,11 +64,14 @@ export default {
       lastName: {
         value: '',
         // Combine rules using an object
-        errorMessages: [{
+        validationRules: [{
```
edgesoft: Think it was a typo

christianalfoni:
@henri-hulski @edgesoft Great stuff! Though I want to challenge one thing here. I do not see the point of adding global props. It adds complexity to the API without making things simpler... cause it does not actually do anything? It is just a state? So you can just as easily have that outside of the form. For example:

```js
import {form} from 'cerebral-forms'

export default function MyAction({state}) {
  state.set('some.new', {
    form: form({
      name: {
        value: '',
      }
    }),
    showFormErrors: false
  })
}
```

Please correct me if I am wrong here. But yeah... :)

edgesoft:
@christianalfoni @henri-hulski Think @christianalfoni is right here. What do you think @henri-hulski

henri-hulski:
I have this now and it really pollutes the name space. 
It would be much cleaner if I could have it inside the form. 

The only thing I change to archive this is to testing if the field is an object. 
```js
// If not an actual field but a global prop or just a namespace
  if (field !== Object(field) || !('value' in field)) {
    return field
  }
```
So I  it's not a big deal.

christianalfoni:
Hm, I see... well, I suppose we could define `Form` as: "Looks for objects with a value key and converts them to a field. Any other keys stays as is".... It feels a tiny bit icky still, cause we have to explain it. 

What if you rather to this?

```js
import {form} from 'cerebral-forms'

export default function MyAction({state}) {
  state.set('some.new.form', {
    fields: form({
      name: {
        value: '',
      }
    }),
    showErrors: false
  })
}
```

Cause this is really just about naming and nesting.

henri-hulski:
@christianalfoni Hm.  More or less what I have now. 
```js
state: {
   login: {
     form: form({}),
     showErrors: false
   }
}
```
But it still feels more natural and clean when it's inside the form. 

BTW `defaultRequireMessage` is also a global prop. So if we want this we get global props for free. 

And I already explained them. Maybe not accurate enough.

henri-hulski:
Maybe we should call them global option as props could get confused with react props.

henri-hulski:
@christianalfoni the check for the value key was there before. I only added `field !== Object(field) || `. Like this `!('value' in field)` doesn't fall with uncaught TypeError when field is not an object.

As we can add custom state to the fields it seems just natural to me to be able to add custom state to the form also. The limitation is that the value of the custom property cannot be an object.
Objects without value key where possible before as a namespace, I think for nested forms.

christianalfoni:
@henri-hulski Hm, sorry, but my intuition tells me that we are introducing too much new API and things to learn. I have been thinking a bit and what you want is some custom behaviour to your forms. You can very easily do that with a factory:

```js
const MyAppForm = (form) => {
  const newForm = Form(form)
  // Some special props
  newForm.showErrors = false
  // Add some defaults
  Object.keys(newForm).forEach(field) => {
    if (!newForm[field].requiredMessage) {
      newForm[field].requiredMessage = 'Some default message'
    }
  })
  return newForm
}

// So the Form stays the same, you have just added some extra behaviour
{
  form: MyAppForm({name: {value: ''})
}
```

My reasoning is that the current Form API probably solves 99% of forms, as most forms are not very complicated. Then it is better for edge cases to use factories or write some extra code, rather than introducing new things to learn for those 99% other cases :)

It is definitely not that I think it is a bad idea, but it just does not "fit like a glove". Of course if we get more feedback from others needing the exact same functionality we can look more into it... but factories are a wonderful thing for creating custom behaviour :)

henri-hulski:
@christianalfoni I not really get what you want. 
Should I remove `defaultRequireMessage` or only the documentation about custom props.

BTW did you read the diff of utils/configureField.js? 
There are the only code changes I made. 
The other changes are related to renaming `errorMessages`, tests and documentation. 

For custom props there is no special code change at all. Only some documentation. 

It works because I added `field !== Object(field) || `, but this is also necessary for `defaultRequireMessage`.

christianalfoni:
@henri-hulski Hi again, hehe, sorry... when it comes to forms I am like a "women who has been hurt too many times", related to the work on Formsy :-)

So what I suggest is:

1. We keep your snippet and change documentation to:

"You can add any properties to a form where properties containing an object with a "value" property are identified as fields"...

2. We do not implement special props that affects the fields (Like `defaultRequiredMessage`. The reason is:

1. Since you can nest forms it would be expected that these default things are inherited down the nesting, which requires more implementation, causing unnecessary complexity as there are other solutions

2. It is very narrow. With point 1. it becomes very complex implementation for just one prop

But instead, we can document a very flexible solution, allowing people to create any kinds of defaults:

```js
import {Form, getFormFields} from 'cerebral-forms'

const MyFormFactory (form) {
  const newForm = Form(form)
  const fields = getFormFields(newForm)

  fields.forEach((field) => {
    field.someProp = field.someProp || 'Some default'
  })

 return newForm
}
```

Does this seem reasonable? Sorry again for being so unclear on this :)

henri-hulski:
@christianalfoni Yeah the default factory looks awesome. 
So I'm agree. :)

Guria:
What status of this PR?

christianalfoni:
@henri-hulski Maybe you want to rebase this pull? Just remove the defaultRequired thingy and add some docs on the factory? Let me know I can help on this as well :)

henri-hulski:
Sure I will do it.

Guria:
@henri-hulski ok, don't forget to update forms-demo as well

henri-hulski:
Done.

PR #465, opened 2016-11-19
refactor(Operators): rename string to toString

christianalfoni:
It was decided to rename this for clarity

### DIFF
--- a/packages/cerebral/src/operators/index.js
+++ b/packages/cerebral/src/operators/index.js
@@ -5,7 +5,7 @@ export {default as equals} from './equals'
 
 export {default as state} from './state'
 export {default as input} from './input'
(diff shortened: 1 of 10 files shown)


Guria:
I have followed an conversation in Discord well. Could you provide motivation to make this change?

christianalfoni:
@Guria Jup, reason is that `state` and `input` operates as targets. Like: `state` means I want to use this template related to state, or I want to use this template related to input. This is not the case for `string`, what you want there is just convert the template to a string, you do not have a target. So it has been confusing... `toString` is just more explicit in terms of not being a target, but just convert the template to a string.

christianalfoni:
https://github.com/cerebral/cerebral/pull/473

PR #464, opened 2016-11-19
fix: errorMessages throws error when empty

edgesoft:
Also added failing test for validateForm

PR #466, opened 2016-11-19
chore: simple forms-demo

edgesoft:
- Simple forms demo


packages/forms-demo/.gitignore
```diff
@@ -0,0 +1,15 @@
...
+# testing
+coverage
+
+# production
+build
+
+# misc
+.DS_Store
+.env
+npm-debug.log
```
henri-hulski: We don't need .gitinore here as we have it at the root of the monorepo.

edgesoft: @henri-hulski Thanks! Have removed it

Guria:
@edgesoft is it ready or will be updated after #462 and #464 ?

edgesoft:
@Guria It's ready

packages/forms-demo/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "name": "forms-demo",
+  "version": "0.1.0",
+  "private": true,
+  "devDependencies": {
+    "react-scripts": "0.7.0"
+  },
+  "dependencies": {
+    "aphrodite": "^1.1.0",
+    "cerebral": "^2.0.0-0-alpha.5a3d129e",
+    "cerebral-forms": "^1.0.0-0-alpha.5a3d129e",
```
Guria: just define "next" here

edgesoft:
@Guria But will get error when #462 is merged.

packages/forms-demo/src/components/Fields/Input/index.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
+import {connect} from 'cerebral/react'
+import {css} from 'aphrodite'
```
Guria: Haven't explored this lib yet. Are this really makes much sense for demo?
Isn't just css not enough?

edgesoft: @Guria It's very straightforward so don't think it would be an issue.

packages/forms-demo/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "forms-demo",
+  "version": "0.1.0",
+  "private": true,
+  "devDependencies": {
+    "react-scripts": "0.7.0"
+  },
+  "dependencies": {
+    "aphrodite": "^1.1.0",
+    "cerebral": "cerebral@next",
```
Guria: wrong again :)
just use  "cerebral": "next"

PR #471, opened 2016-11-20
docs: update --save to next

edgesoft:


Guria:
I am not sure about this change. There is pros and cons on both variants.

`npm i package@next --save-dev` - will stick to exact version in package.json, but it will also secure a setup to be stable. user would have to rerun this command to get new "next" when he ready.

Defining "next" in package.json will result to newest next on each npm i and it could break user setup.

edgesoft:
Ok, don't have any opinion here. Should I close it?

Guria:
I still don't sure which option is better here. Maybe we should just add note about having to rerun first command to get updated to latest next.

edgesoft:
Someting like this.

**NOTE** If you need to update to a later @next release dot the npm install again.

Guria:
Yes

Guria:
Hm. I thought to keep `npm install cerebral-module-useragent@next --save` instruction and have a note to rerun it again to be updated.
But maybe wait for @christianalfoni opinion here as well?

christianalfoni:
Yeah, there are smaller changes now so I think the note should just say: "Run npm i again to update your next packages. There are still changes to API so please check latest updates in case something  has changed".

Guria:
@christianalfoni question not about a note but about initial way to setup. See options [here](https://github.com/cerebral/cerebral/pull/471#issuecomment-261798572)

christianalfoni:
Oh so `npm i package@next` does not add `next` in package.json? But latest version?

If that is the way it works I would vote for using `npm i package@next`, to keep it stable. They decide themselves when to move to next version by installing again.

edgesoft:
Yes, thing npm i package@next should be there and a note on how to update i new features or fixes where made.

Guria:
@edgesoft Let's update it to:
```
This is still alpha but to test the bleeding edge.
  
`npm install cerebral-forms@next --save`

**NOTE**: If you need to update to a later @next release than execute above command again.
```

edgesoft:
Updated to the last suggestion

Issue #412, opened 2016-11-10
revisit migration guide before release


edgesoft:
- [x] Operators change is not in migration. 
- [x] controller.getSignals() is also missing so this needs to be adressed
- [x] Collect state operations that has been deleted and document it when we have settled on what should be removed or reimplemented

edgesoft:
From discord chat.

- Components (2 people - TO / SA)
    - Change import to “cerebral/react”
    - Move signals to Cerebral deps
- Modules (1 people - WA)
    - Export object instead
- Computed (1 people - CH)
    - Do not call computed
    - If passing in props, use “someComputed.props({})” instead
- Operators (2 people - CH, WA)
    - Refactor actions to use only operators
    - state, input, string template tags
    - set(statefoo.bar, inputsomething)
- Actions (2 people)
    - No more output, use path
    - No more services, use firebase, http
    - Remove async/outputs
    - Return promises
    - Return paths

edgesoft:
#472

edgesoft:
#479

PR #479, opened 2016-11-21
docs(website): adding missing part to migration

edgesoft:
fixes: #412


edgesoft:
Have updated migration with @christianalfoni suggestions.

PR #481, opened 2016-11-21
fix(tutorial): make sure source folder is not hidden

christianalfoni:
When `src` is part of  `gitignore` it might be hidden by IDEs, it should not do that, very confusing. The SRC folder is built up with chapter 01 when starting up anyways.

### DIFF
--- a/packages/tutorial/.gitignore
+++ b/packages/tutorial/.gitignore
@@ -1,2 +1,1 @@
 /public/index.html
-/src


henri-hulski:
@christianalfoni Now the src folder will be included in every pull request. Could also get confusing.

christianalfoni:
@henri-hulski Completely agree :) Though I consider this less bad than the other one. More important to take care of beginners than people are at the level of contributions. But maybe there is a third option here... just dunno :)

henri-hulski:
Hmm in Atom for example you can make ignored files visible by unchecking the “Hide VCS Ignored Files” setting in the tree-view package. I think other IDEs has a similar switch probably even easier to find. 😉

henri-hulski:
BTW  This is also helpful for examining the node_modules folder.

christianalfoni:
Yeah, it has the switch, but by default it hides gitignore files. We can not start the tutorial with:

- For this idea, do this
- For this idea, do this

Configuring an IDE to go through a tutorial is a really bad start ;-)

henri-hulski:
Agree. 😊

christianalfoni:
Though, still spinning... maybe we will come up with best of both worlds somehow :)

Guria:
We were about to prepare zip to download

christianalfoni:
Yeah, maybe that is better... this clone stuff scares people who are not on Github.... you know... I vote for ZIP. What you guys think? Btw we need somewhere to put electron apps as well. Any ideas? Just use google drive?

PR #482, opened 2016-11-21
docs(website): add standalone debugger downloads

christianalfoni:
Okay, got them built and uploaded to Google Drive for now... completely open to put them somewhere else :) But at least now people can try it out!

### DIFF
--- a/packages/cerebral-debugger/package.json
+++ b/packages/cerebral-debugger/package.json
@@ -9,7 +9,10 @@
     "build:chromeExtension": "NODE_ENV=production webpack --config webpack.chromeExtension.config",
     "build:electron": "NODE_ENV=production webpack --config webpack.electron.config && npm run prepare:electron && npm run package:electron",
     "prepare:electron": "rimraf electron/dist electron/node_modules && copyfiles 'node_modules/ws/**/*' 'node_modules/ultron/**/*' 'node_modules/options/**/*' electron",
-    "package:electron": "electron-packager ./electron CerebralDebugger --platform=darwin --out ./electron/dist"
+    "package:electron": "npm run package:electron:mac && npm run package:electron:win && npm run package:electron:linux",
+    "package:electron:mac": "electron-packager ./electron CerebralDebugger --platform=darwin --out ./electron/dist",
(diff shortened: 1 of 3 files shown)


christianalfoni:
Hm... we could have these on github as well, though they are 40MB each. Meaning the install of monorepo would be very large.

Maybe create a separate repo for this? Or any other ideas?

henri-hulski:
Yeah this is the caveat I thought about. If it's so big no way to put it in the repo. 
The best would be to add it as a binary file to the release. In the meantime either leave it on Google or create a separate repo.

henri-hulski:
Maybe the easiest would be to just make a preview release for each every  new version of the debugger. 
Not sure if there will be a permanent url but otherwise we can add an URL to a redirect on the website.

Guria:
Github releases supports to attach binaries

Guria:
Just downloaded debugger and have several questions:
* "Please refresh your app to use the debugger..." is completely useless. We should provide better instructions here.
* which port does it starts? can it be configured? 8080 is not a good choice for default port.
* what will happen when several instances and/or apps will connect to the same debugger host?

henri-hulski:
@Guria Yep, but if I'm not mistaken every release will have a different URL. So to have a permanent URL we need a redirect somewhere so it's easier to bookmark. 
But not really sure if this is necessary.

Guria:
@henri-hulski https://github.com/cerebral/cerebral/releases/latest is pretty permanent

Guria:
I will merge this for now, but we must address issues described above

henri-hulski:
@Guria Aah cool. So something like 
https://github.com/cerebral/cerebral/releases/download/latest/tutorial.zip

christianalfoni:
@Guria @henri-hulski Okay, so we can attach debugger to repo somehow? How will this work? And we need to version them.

@Guria I will address the issues you stated, completely agree :)

henri-hulski:
@christianalfoni Take a look what @Guria did [here](https://github.com/cerebral/cerebral/releases/tag/untagged-e5b773432376958077d1).

christianalfoni:
Oh... cool... how did you do that @Guria ?

PR #483, opened 2016-11-21
docs(tutorial): add tutorial as zip and point to new debugger page

christianalfoni:
Depends on : https://github.com/cerebral/cerebral/pull/482


henri-hulski:
I would propose to use Github for the zip files. 
We can even put it in the root of the package and there should be a raw url to point to it. 

Alternatively we could create zips automatically in Github releases similar to the Atom downloads. 

But at least for now the former makes more sense.

henri-hulski:
The URL would be something like 
https://raw.githubusercontent.com/cerebral/cerebral/master/packages/tutorial/tutorial.zip

christianalfoni:
Yeah, github makes a lot more sense, totally agree. Let me change that :)

christianalfoni:
Okay, this is updated now, will do the same for electron apps

Guria:
No way. Let's avoid artifacts under vcs

christianalfoni:
@Guria I am completely blank on this... not sure what is best approach. You have any idea of where we can/should store it? :)

PR #486, opened 2016-11-22
refactor(forms): merge runValidation and validate

henri-hulski:
Move runValidation from validateField to validate.
As runValidation becomes the default export rename file to runValidation.js.
Reuse runValidation in validateForm and configureField.


henri-hulski:
Added also a fix to check first if the key is an object before testing if it contains a 'value'  key.

Guria:
Congrats, You have just won 1000 commit on repo 🎆

henri-hulski:
That's great!  🙌🎉🍻

PR #490, opened 2016-11-23
test(function-tree): ensure inputs are passed correctly with empty paths

christianalfoni:
Just a test to ensure that an edge case in Cerebral 1 does not happen in Cerebral 2

### DIFF
--- a/packages/function-tree/src/FunctionTree.test.js
+++ b/packages/function-tree/src/FunctionTree.test.js
@@ -240,4 +240,23 @@ describe('FunctionTree', () => {
     assert(branchStartCount, 2)
     assert(branchEndCount, 2)
   })
+  it('should pass correct input on empty paths', () => {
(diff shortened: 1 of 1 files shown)


packages/function-tree/src/FunctionTree.test.js
```diff
@@ -240,4 +240,25 @@ describe('FunctionTree', () => {
     assert(branchStartCount, 2)
     assert(branchEndCount, 2)
   })
+  it('should pass correct input on empty paths', () => {
+    let branchStartCount = 0
+    let branchEndCount = 0
```
henri-hulski: Lint errors: branchStartCount and branchEndCount are assigned a value but never used.

christianalfoni: DOH! Will rebase soon ;-)

PR #494, opened 2016-11-24
[wip] Add codemods for moving from Cerebral 1.x to 2.x

NervosaX:
Addresses https://github.com/cerebral/cerebral/issues/318

### DIFF
--- a/codemods/1.x-2.x/__testfixtures__/computed-decorator.output.js
+++ b/codemods/1.x-2.x/__testfixtures__/computed-decorator.output.js
@@ -1,0 +1,15 @@
+/* eslint-disable */
+import {connect as Cerebral} from 'cerebral/react';
+import {limitedList} from "./limitedList";
+import {anotherComputed} from "./anotherComputed";
+
+@Cerebral({
+  anotherComputed: anotherComputed,
+  limitedList: limitedList.props({ foo: "bar" }),
+  myState: "path.to.state"
+})
+export default class List {
+  render() {
+    return <div />;
+  }
+}
--- a/codemods/1.x-2.x/__testfixtures__/controller.output.js
+++ b/codemods/1.x-2.x/__testfixtures__/controller.output.js
@@ -1,0 +1,24 @@
+/* eslint-disable */
+import {Controller} from 'cerebral'
+
+import Home from './modules/Home'
+import Admin from './modules/Admin'
+import doSomething from './chains/doSomething'
+
+const controller = Controller({
+  state: {
+    // You can add some initial state here if you want
+    foo: 'bar'
+  },
+
+  modules: {
+    home: Home(),
+    admin: Admin()
+  },
+
+  signals: {
+    buttonClicked: doSomething
+  }
+})
+
+export default controller
--- a/codemods/1.x-2.x/__testfixtures__/operators-flow.input.js
+++ b/codemods/1.x-2.x/__testfixtures__/operators-flow.input.js
@@ -1,0 +1,17 @@
+/* eslint-disable */
+import {
+  when
+} from 'cerebral/operators';
+
+export default [
+  // Conditional truthy check of state or input
+  when('state:foo.isAwesome'), {
+    true: [],
+    false: []
+  },
+
+  when('input:foo.isAwesome'), {
+    true: [],
+    false: []
+  }
+]
--- a/codemods/1.x-2.x/__testfixtures__/operators-flow.output.js
+++ b/codemods/1.x-2.x/__testfixtures__/operators-flow.output.js
@@ -1,0 +1,19 @@
+/* eslint-disable */
+import {
+  when
+} from 'cerebral/operators';
+
+import { state, props } from "cerebral/tags";
+
+export default [
+  // Conditional truthy check of state or input
+  when(state`foo.isAwesome`), {
+    true: [],
+    false: []
+  },
+
+  when(props`foo.isAwesome`), {
+    true: [],
+    false: []
+  }
+]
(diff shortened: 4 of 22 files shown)


codemods/1.x-2.x/__testfixtures__/controller.output.js
```diff
@@ -0,0 +1,7 @@
+import {Controller} from 'cerebral'
+
+const controller = Controller({
+  // You can add some initial state here if you want
+})
+
+export default controller
```
gaspard: Should be:

```js

  state: {
    // You can add some initial state here if you want
  }
})

export default controller
```

codemods/1.x-2.x/__testfixtures__/operators-flow.input.js
```diff
@@ -0,0 +1,29 @@
...
+} from 'cerebral/operators'
+
+export default [
+  // Conditional truthy check of state or input
+  when('state:foo.isAwesome'), {
+    true: [],
+    false: []
+  },
+
+  when('foo.isAwesome'), {
```
gaspard: I think this was not supported, it was `state:foo.isAwesome`...

codemods/1.x-2.x/__testfixtures__/operators-state.output.js
```diff
@@ -0,0 +1,29 @@
+import {
+  set,
+  copy,
```
gaspard: No more `copy`... (not needed with operator tags)

codemods/1.x-2.x/__testfixtures__/operators-state.output.js
```diff
@@ -0,0 +1,29 @@
...
+  unset,
+  toggle,
+  state,
+  input,
+  output
+} from 'cerebral/operators'
+
+export default [
+  // Copy
+  set(state`foo.bar`, input`bar`),
```
gaspard: This is

```js
set(input`bar`, state`foo.bar`)
```
```

File: codemods/1.x-2.x/__testfixtures__/operators-state.output.js [hunk 5]
```diff
@@ -0,0 +1,29 @@
...
+  set(input`foo`, output`bar`),
+  set(input`foo`, state`foo.bar`),
```
gaspard: Should be

```js
set(state`foo.bar`, input`foo`)
```

codemods/1.x-2.x/__testfixtures__/operators-state.output.js
```diff
@@ -0,0 +1,29 @@
...
+  set(state`foo.bar`, state`bar.baz`),
```
gaspard: Should be

```js
set(state`bar.baz`, state`foo.bar`)
```

codemods/1.x-2.x/__testfixtures__/operators-state.output.js
```diff
@@ -0,0 +1,29 @@
+import {
+  set,
+  copy,
+  unset,
+  toggle,
+  state,
+  input,
+  output
```
henri-hulski: Also no more `output`.

codemods/1.x-2.x/__testfixtures__/computed-decorator.output.js
```diff
@@ -0,0 +1,14 @@
+import {connect as Cerebral} from 'cerebral/react';
```
NervosaX: https://github.com/cerebral/cerebral/blob/391a142da89628938c0c31263525d7aa084f3a3e/packages/demo/src/components/Tasks/index.js

Looking at this demo, I'd say it's still call connect.

```
@connect()
class ....
```

is the same as

```
connect(
  class ...
)
```

I just rename it to `@Cerebral` :)

codemods/1.x-2.x/__testfixtures__/operators-flow.output.js
```diff
@@ -18,7 +14,7 @@ export default [
   },
 
   // Wait 200ms, then continue chain
-  ...wait(200, [
+  wait(200, [
```
gaspard: This is now `wait(200)` without any other parameter.

```js
wait(200),
doSomething,
```

NervosaX: Great, will fix it up.

codemods/1.x-2.x/__testfixtures__/operators-flow.output.js
```diff
@@ -14,7 +14,5 @@ export default [
   },
 
   // Wait 200ms, then continue chain
-  wait(200, [
-    doSomething
-  ])
+  wait(200)
```
gaspard: But you need the `doSomething` here, after wait from how I understand the output...

NervosaX: Ah, crap. You're right.

NervosaX: This one is proving hard. Moving the array starts messing with the whole piece of code's whitespace.

fvgoto:
If this PR is still worth it, I guess it should be updated according to
#660 rename `input` to `props`

NervosaX:
Thanks @fvgoto. I think it's worth it, I am still waiting for some of these changes to stabilise. I'll definitely be using this to upgrade my codebase once 2.0 drops, so at least it's worth it for me. :)

Guria:
ok, lets revisit this now with sequence/parallel, add some documentation and merge it in

NervosaX:
Okay, I'll have a look into the new changes, and see if I can find nice ways to apply these mods to a codebase. 

We will need to document what it can't do automatically as well (for example the issues I had with the  delay->wait codemod that ruins code style).

There's currently nothing for converting from services.

NervosaX:
@Guria Okay, I've fixed up the input tag -> props tag change.

I have yet to do the sequence/parallel change... that one might prove tricky to automate as well. Maybe we merge this in and work from master from here (others might feel free to contribute that way as well)

Guria:
Did some squash and integrations with monorepo.
Merging it as is now.
Would be great to get usage instructions on website migration,

PR #495, opened 2016-11-24
feat(forms): fixed default requiredMessage and global props factory example

yusufsafak:
Hi guys,
Form should be lowercase 
Related to cerebral/cerebral#462, you decided to keep global props outside the form. So I deleted this example and put it in factory. 
Thanks for the great library :)

### DIFF
--- a/packages/cerebral-forms/README.md
+++ b/packages/cerebral-forms/README.md
@@ -96,29 +96,21 @@ export default function MyAction({state}) {
 #### Set a default value for the whole form
 You can set a default value for a property using a factory:
 ```js
-import {Form, getFormFields} from 'cerebral-forms'
+import {form, getFormFields} from 'cerebral-forms'
 
-const MyFormFactory = (form) => {
-  const myForm = Form(form)
+const MyFormFactory = (formObject) => {
+  const myForm = form(formObject)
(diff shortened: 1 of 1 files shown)


edgesoft:
@yusufsafak Great catch thanks! Just a question. Any reason MyAction is removed?

henri-hulski:
@yusufsafak Nope. The decision was:

> We keep your snippet and change documentation to: "You can add any properties to a form where properties containing an object with a "value" property are identified as fields"...

So global props are allowed in the form.

yusufsafak:
Himm I think I understood wrong this:

> We keep your snippet and change documentation to: "You can add any properties to a form where properties containing an object with a "value" property are identified as fields"...

yusufsafak:
:) I will correct it now

henri-hulski:
Yeah it's not really clear. Feel free to reformulate this sentence.

henri-hulski:
@yusufsafak Sorry but you were actually right with removing the example. I forgot to clean it up. 
Like now it makes no sense. 
But I think I will restore it as it was before with a separate header as it is correct and I will also use it in the demo.

henri-hulski:
@yusufsafak So thanks for catching this! ☺

PR #506, opened 2016-11-25
fix(forms): Fix dependsOn failing test and add more tests

yusufsafak:


henri-hulski:
Super awesome!!!

packages/cerebral-forms/src/factories/validateField.js
```diff
@@ -18,7 +18,7 @@ export default function validateFieldFactory (pathTemplate) {
       dependentFields = [field.dependsOn]
     }
 
-    const depententOfValidationResult = dependentFields.reduce((currentValidationResult, stringPath) => {
+    const dependentOfValidationResult = dependentFields.reduce((currentValidationResult, stringPath) => {
```
yusufsafak: Why is dependsOn array here? Do you have any use case for it?

PR #507, opened 2016-11-25
feat(forms): add new initializeForm factory and nested form tests

yusufsafak:


henri-hulski:
Looks great! Just some linting must be fixed.

henri-hulski:
And documentation of initializeForm is missing.

yusufsafak:
Thanks for review. I fixed lint errors and add documentation of initializeForm :)

henri-hulski:
Are you sure that you need here
```js
        newField.defaultValue = newInitialValues || newField.defaultValue
        newField.value = newField.defaultValue
        newForm[key] = configureField(form, newField)
```
as in configureField we have
```js
  const value = field.value
  const defaultValue = field.defaultValue || value
```
Which looks more or less the same.

henri-hulski:
But ok maybe it makes sense.
I think the PR looks ready now.

PR #511, opened 2016-11-26
WIP: Save tasks in demo and Collection

gaspard:
There might be some glitches but the general idea on how collections could be used is in place. This needs some beta testing but it works...


packages/demo/src/common/Collection/signals/init.js
```diff
@@ -0,0 +1,33 @@
...
+import paths from '../paths'
+
+export default function init (moduleName, initState = {}) {
+  const {collectionPath, dynamicPaths, errorPath} = paths(moduleName)
+  return [
+    // prepare remote and local path
+    ...dynamicPaths,
+    // start listening
+    onChildAdded(
+      input`remoteCollectionPath`, `${moduleName}.updated`),
```
saitodisse: @gaspard what I usually do here is pass an option to get older values, like that:

```js
{
  orderByChild: 'updated_at',
  startAt: (new Date()).getTime(),
}
```

Example: https://github.com/saitodisse/scrap-cerebral-2/blob/master/src/shared_actions/firebase/firebase_listen.js#L1


Docs: https://github.com/cerebral/cerebral/tree/master/packages/cerebral-provider-firebase#onchildadded

That way we only get new items when they are added. Bellow we are calling `value(input`remoteCollectionPath`), {` so all item will be downloaded. That's why I put the `updated_at` prop. I believe this is the cause for some error on initialization of `createOnChildAdded.js`

gaspard: BTW: startAt should be UTC date, no ?

gaspard: I think this should go inside the success of the value call so that we do not risk missing items.

gaspard:
OK Everyone, this new iteration seems done. I did some bug fixes and testing. Since I had to change the initial data, pleaser recreate login to test app.

packages/demo/src/common/Collection/computed/visibleKeys.js
```diff
@@ -0,0 +1,26 @@
...
+      items: `${collectionPath}.**`,
+      afilter: filterPath,
+      selectedKey: `${draftPath}.key`
+    },
+    ({items, afilter, selectedKey}) => {
+      const filter = afilter && afilter.toLowerCase()
+      const list = Object.keys(items).filter(key => (
+        !filter || items[key].name.toLowerCase().indexOf(filter) >= 0
+      ))
+      if (selectedKey && list.indexOf(selectedKey) < 0) {
```
yusufsafak: Hi Gaspard, when you have draft project on your page and you click the project selector (next to Start button), visibleProjectsByClient computed throws an error like `Uncaught TypeError: Cannot read property 'clientKey' of undefined(…)`. :(
I haven't check the code completely yet but when I remove this line, it does not show draft project. I understood that project list and project selector use visibleKeys computed. Also they are using the same filter. Maybe we should create a new computed for project selector. What do you think?

gaspard: Fixed. Thanks for this review.

Guria:
@gaspard is still wip?

gaspard:
@Guria I had to add visibility API because of some strange high CPU usage after running for a long time. I think this version is much more stable then current version in master so we could pull this in.

It will continue to evolve obviously because there are still some important features missing but this round is OK for me.

PR #512, opened 2016-11-26
Flowops

christianalfoni:
- Fixed wait/debounce operator for simpler syntax
- Several docs and tutorial fixes


Guria:
`refactor` commit type couldn't change api. Use `feat` instead next time

PR #514, opened 2016-11-26
fix: change priority hasChange (in submitted Props)

fopsdev:
if submitted properties changed do no further checking just return true so the comp gets rerendered


christianalfoni:
Okay, I did a rebase here with tests and changed around a little bit :)

fopsdev:
child is still rerendered as it should :) thanks @christianalfoni

PR #516, opened 2016-11-27
feat(Views): add dynamic paths based on state

christianalfoni:
@Guria @gaspard After playing around with it I figured it was the wrong way to go with operators here. It is not necessary, does not fit the signature as well and we would have an issue explaining why operators can be used with connect as well. But I think this looks way better anyways:

```js
connect((props, state) => {
  const currentUserKey = state('currentUserKey')

  return {
    user: `users.${currentUserKey}`
  }
}, ...)
```

### DIFF
--- a/packages/cerebral/src/viewFactories/Hoc.js
+++ b/packages/cerebral/src/viewFactories/Hoc.js
@@ -4,15 +4,36 @@ import {cleanPath, propsDiffer} from './../utils'
 export default (View) => {
   return function HOC (paths, signals, injectedProps, Component) {
     class CerebralComponent extends View.Component {
-      static getStatePaths (props) {
+      static getStatePaths (props, controller) {
         if (!paths) {
           return {}
         }
-        return typeof paths === 'function' ? paths(props) : paths
+
+        if (typeof paths === 'function') {
+          const dynamicPaths = {}
+          const getState = (path) => {
+            dynamicPaths[`__${path}`] = path
+
+            return controller.getState(path)
+          }
+          return {
+            static: paths(props, getState),
+            dynamic: dynamicPaths
+          }
+        }
+
+        return {
+          static: paths,
+          dynamic: {}
+        }
       }
-      constructor (props) {
-        super(props)
-        this.evaluatedPaths = CerebralComponent.getStatePaths(props)
+      constructor (props, context) {
+        super(props, context)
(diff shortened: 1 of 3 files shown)


Guria:
Don't sure about exposing state as function. Why just not use Computed here with explicit state dependency defined?

gaspard:
I agree that having something simpler is nice but am not a fan of `state` like this...

With computed, it looks like this and is very difficult syntax wise:

```js
connect({
  user: Computed({ key: 'currentUserKey'}, props => (
    Computed(({key}) => ({user: `users.${key}`}), ({user}) => user).props(props)
  ))
}), ...)
```

christianalfoni:
Hi there,

I was not very good at explaining why we need it, or something similar :)

A normal computed will not work:

```js
Computed({
  users: 'users',
  currentUser: 'currentUser'
}...)
```

The reason is that this computed till update whenever there is a change to `users`, which is not what we want. We need a way to do lookups inline.

The only way to handle this now is to pass in the `currentUser` from the parent component.

```js
connect((props) => ({
  user: `users.${props.currentUser}`
})
```

But then the component depends on the parent to get its state. I am not saying that is a bad thing, but there are scenarios where this is not apparent. Like you could use a computed instead... which makes all list items rerender, because the computed updated, because the  `users` updated.

But, maybe it is enough with documentation here instead. Cause all of this is related to lists. There already is documentation on lists related to it. And the other issue with this was modals who are going to display a specific user. Same issue. But again, the documentation could show how to do it.

Okay, I suggest this is premature implementation. Lets continue with documentation first as it is special scenarios :)

christianalfoni:
The other fixes is rebased into @fopsdev pull, https://github.com/cerebral/cerebral/pull/514

gaspard:
@christianalfoni I understand the optimisation here, hence my double Computed thing in the example. So maybe some kind of improvement on Computed would be easier (and would not add API to connect):

```js
connect({
  user: ComputedPath`users.${state`currentUser`}`
}), ...)
```

The idea here is that `ComputedPath` is a Computed factory.

christianalfoni:
Yeah, that is also an idea. Lets keep keep this in the back of our heads as we move on. Generally want to avoid bringing in new APIs and concepts, so I suggest we see how far documentation takes us first :) But very good experimentation of ideas here!

PR #521, opened 2016-11-29
refactor(forms): remove initializeForm

yusufsafak:


edgesoft:
@yusufsafak Why do we need to remove initializeForm

henri-hulski:
Cc @christianalfoni

christianalfoni:
It is the balance of "do it yourself" and creating APIs for everything :) Really want to emphasise that forms is completely "hackable". So instead of an API for this, you can:

```js
function updateForm ({input, state}) {
  state.set('form.field1.value', input.field1)
  state.set('form.field2.value', 'bar')
}
// chain
[
  updateForm,
  validateForm('form').
]
```

Balancing is super difficult to do, it is this balance of solving everything with the API or give the possibility to do what you want. For every factory that solves some scenario we go further away from making the developer solve the problems themselves... suddenly we are in a dark rabbit hole with 100 different factories for every possible use case because nobody knew that forms is completely "hackable" :) this is my being paranoid of course, hehe... but I think it is important to make it very difficult to add new official API as long as it can easily be done with current.

But documenting the hackability of forms could probably be improved, cause it really is a magical thing that you can change/add any state to the existing stuff :)

yusufsafak:
Hi @edgesoft @christianalfoni ,
I also found a another way to initialize form. Maybe you are already using that.
I keep my form as an object outside of the state. Whenever I need to fill my form with a new data, I create a new form using this form object and with data.
This way you can divide your form into small forms. You can use it anywhere. 

```
const myBaseForm = {
  name: {
    value: '',
    required: true
  },
  country: {
    value: '',
    defaultValue: 'Germany'
  }
}

const myDetailForm = {
  city: {
    value: ''
  }
}

const myCombinedForm = Object.assign({}, myBaseForm, myDetailForm)
```

Now you can use factory to fill the form. Something like:

```
const fillFormFactory = (dataTemplate, passedDefaultFormTemplate, passedFormTemplate) =>{
  const fillForm = (context) => {
    const {state, input} = context
    const defaultForm = (typeof passedDefaultFormTemplate === 'function' ? passedDefaultFormTemplate(context).value : passedDefaultFormTemplate) || input.defaultForm
    const formPath = (typeof passedFormTemplate === 'function' ? passedFormTemplate(context).value : passedFormTemplate) || input.form
    const data = (typeof dataTemplate === 'function' ? dataTemplate(context).value : dataTemplate) || input.serviceName || input.moduleName

    const newFormObject = Object.keys(defaultForm).reduce((newForm, key) => {
      newForm[key] = Object.assign({}, defaultForm[key])
      newForm[key].value = data[key]
      return newForm;
    }, {})
    state.set(`${formPath}`, form(newFormObject))
  }
  return fillForm
}


fillFormFactory(basedata, myBaseForm, 'client.baseForm')
fillFormFactory(detaildata, myDetailForm, 'client.detailForm')
fillFormFactory(combinedata (or baseData), myCombinedForm, 'client.detailForm')
```

This way you have always a clean default forms. You don't need resetForm / validateForm factory to initialize the form.
Maybe we can also document this :)
Thanks for the great discussion :)

@edgesoft  Do you need any help for form-demo?

edgesoft:
@yusufsafak Of course. See discord under forms.

Guria:
So, are we all ok with this? Can't see any approves here

PR #522, opened 2016-11-29
feat(binary): add support for passing binary files to signals via refs

gaspard:
API:

```js
import {binary} from 'cerebral'

// in component
const ref = binary.push(e.target.files)
// in signal
const files = binary.pop(ref) // this also removes the ref from binary
```

Usage is for upload or other operations dealing with binary data. It helps to not implement signal logic in Component simply because the value cannot be serialized.

This also supports a fixed ref so that `ref` can be sent on fieldChange multiple times without memory leak:

```js
import {binary} from 'cerebral'

// in component
const onFileChange = e => {
  // This can be executed multiple times before form validation triggers upload
  // without memory leak.
  const ref = binary.push(e.target.files, 'clientLogo')
  fieldValueChanged({key: 'clientLogo', value: ref})
}
```

### DIFF
--- a/packages/cerebral/src/binary.js
+++ b/packages/cerebral/src/binary.js
@@ -1,0 +1,15 @@
+let referenceRefCount = 0
+const storage = {}
+
+export default {
+  push (binary, fixedRef = null) {
+    const ref = `binary-ref-${fixedRef || ++referenceRefCount}`
+    storage[ref] = binary
+    return ref
+  },
+  pop (ref) {
+    const binary = storage[ref]
+    delete storage[ref]
+    return binary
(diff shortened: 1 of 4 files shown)


gaspard:
@Guria Yes and no. The way signals work by not allowing passing non-serializable values creates the need for this pattern. I agree that it could be app specific but there is a use case where it helps to have this shared: it lets you pass these reference across modules and providers.

A typical use case is handling the fileupload in firebase.upload by passing this reference.

gaspard:
Christian will work on another implementation with a better api and no memory-leak risk :-)

PR #528, opened 2016-11-30
feat(Controller): add ability to handle binary like values

christianalfoni:
Allrighty... after long and good discussions on handling binary like things (not serializable stuff) I think I found a really good solution.

We talked about identifying these on the input or use a special `.keyName` convention... but we already check the keys if they are serializable, so only thing now is to handle it differently when they are not (The payload to signal itself still has a check on being serializable though, like it needs to be a plain object):

```js
someSignal({
  file: event.target.files[0]
})
```

There is so little code it is silly, the only thing it does when it is not serializable is that it attaches a `toJSON` method to the object :) 

In actual state tree it looks like:
```js
{
  file: FileObject
}
```

And when serialized it looks like:
```js
{
  files: '[NON_SERIALIZABLE]'
}
```

Feel free to suggest some other naming there... but now we do not even have to mention binary. We just allow you to pass in whatever on the keys. Some is serializable some is not. We just note to people that time travel debugging does not work with "non serializable" state :)

### DIFF
--- a/packages/cerebral/src/Controller.js
+++ b/packages/cerebral/src/Controller.js
@@ -2,7 +2,7 @@ import DependencyStore from './DependencyStore'
 import FunctionTree from 'function-tree'
 import Module from './Module'
 import Model from './Model'
-import {ensurePath, isDeveloping, throwError, isSerializable, verifyStrictRender} from './utils'
+import {ensurePath, isDeveloping, throwError, isSerializable, verifyStrictRender, forceSerializable, isObject} from './utils'
 import VerifyInputProvider from './providers/VerifyInput'
 import StateProvider from './providers/State'
 import DebuggerProvider from './providers/Debugger'
@@ -145,31 +145,30 @@ class Controller extends EventEmitter {
     return this.model.get(ensurePath(path))
   }
   /*
-    Checks if payload is serializable
-  */
-  isSerializablePayload (payload) {
-    if (!isSerializable(payload)) {
-      return false
-    }
-
-    return Object.keys(payload).reduce((isSerializablePayload, key) => {
-      if (!isSerializable(payload[key])) {
-        return false
-      }
-
-      return isSerializablePayload
-    }, true)
-  }
-  /*
     Uses function tree to run the array and optional
     payload passed in. The payload will be checkd
   */
   runSignal (name, signal, payload = {}) {
-    if (this.devtools && this.devtools.enforceSerializable && !this.isSerializablePayload(payload)) {
-      throwError(`You passed an invalid payload to signal "${name}". Only serializable payloads can be passed to a signal`)
+    if (this.devtools && (!isObject(payload) || !isSerializable(payload))) {
+      console.warn(`You passed an invalid payload to signal "${name}". Only serializable payloads can be passed to a signal. The payload has been ignored. This is the object:`, payload)
+      payload = {}
     }
 
-    this.runTree(name, signal, payload || {})
+    if (this.devtools) {
+      payload = Object.keys(payload).reduce((currentPayload, key) => {
+        if (!isSerializable(payload, this.devtools.allowedTypes)) {
+          console.warn(`You passed an invalid payload to signal "${name}", on key "${key}". Only serializable values like Object, Array, String, Number and Boolean can be passed in. Also these special value types:`, this.devtools.allowedTypes)
+
+          return currentPayload
+        }
+
+        currentPayload[key] = forceSerializable(payload[key])
+
+        return currentPayload
+      }, {})
+    }
+
+    this.runTree(name, signal, payload)
   }
   /*
     Returns a function which binds the name/path of signal,
--- a/packages/cerebral/src/utils.js
+++ b/packages/cerebral/src/utils.js
@@ -25,18 +25,30 @@ export function isObject (obj) {
   return typeof obj === 'object' && obj !== null && !Array.isArray(obj)
 }
 
-export function isSerializable (value) {
+export function isSerializable (value, additionalTypes = []) {
+  const validType = additionalTypes.reduce((currentValid, type) => {
+    if (currentValid || value instanceof type) {
+      return true
+    }
+
+    return currentValid
+  }, false)
+
   if (
+    value !== undefined &&
     (
-      isObject(value) &&
-      Object.prototype.toString.call(value) === '[object Object]' &&
-      value.constructor === Object
-    ) ||
-    typeof value === 'number' ||
-    typeof value === 'string' ||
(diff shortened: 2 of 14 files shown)


christianalfoni:
It might be an idea here actually to discard "non serializable" payload objects... not throwing errors. People have a tendency to not handle normal clicks etc. correctly. That is not where people go wrong... like "where is my event object?", it is that they are passing something in that they did not realise. 

So yeah... just ignoring those payloads I think would be good! Please comment :)

packages/cerebral/src/utils.js
```diff
@@ -107,3 +111,19 @@ export function debounce (func, wait) {
...
 export const noop = () => {}
+
+export const forceSerializable = (payload) => {
+  return Object.keys(payload).reduce((currentPayload, key) => {
+    currentPayload[key] = payload[key]
+
+    if (!isSerializable(currentPayload[key])) {
+      Object.defineProperty(currentPayload[key], 'toJSON', {
+        value () {
+          return '[NON_SERIALIZABLE]'
```
gaspard: Maybe the value can be created only once.

packages/cerebral/src/Controller.js
```diff
@@ -188,8 +172,8 @@ class Controller extends EventEmitter {
       throwError(`There is no signal at path "${path}"`)
     }
 
-    return function (payload) {
-      this.runSignal(path, signal, payload)
+    return function (payload, binaries) {
```
gaspard: Not sure why this is here... :-)

gaspard:
For me ignoring events passed in signals is not so good. It's something annoying at first but it makes the code more readable:

```js
<button onClick={() => mySignal()} /> // Makes it obvious what is passed in sighal.
```

gaspard:
I like this solution most. It has many advantages:

- It is the least intrusive 
- it does not require any new API
- it works as expected to newcomers

packages/cerebral/src/utils.js
christianalfoni: It is only created once? :) It happens before signal runs... or do I misunderstand?

packages/cerebral/src/Controller.js
christianalfoni: Ops, merge issue, will fix!

christianalfoni:
@gaspard About wrong payload object, I was thinking we throw it away, BUT warn about it:

"You passed in an invalid payload object on signal"

This allows us to also log out the object that was invalid.

In devtools settings we have an option: `{warnInvalidPayload: true}`, which of course is true by default... basically, using warning and log object, rather than throwing error :)

Guria:
@christianalfoni please document it as well :)

christianalfoni:
Oh, of course :-) let me do another rebase!

christianalfoni:
Okay, lets see. I got some thinking down trying to calm down the little misses with stomach ache :)

We do want to warn when invalid values are put into the state tree. The current change gave you ability to just put whatever in, which is not really what we want. To allow both "special values" and still warn about wrong usage, these changes were made:

- Now Cerebral only checks serializability when devtools is active. Removed the `enforceSerializability` option, it is always on, but now it does not run in production

- If not sending in a valid base object as payload to signal you get a warning and the payload is ignored. This will avoid app just stopping, but rather educate. Typically people do not know they pass in the event object on clicks, now you just get a warning and nothing is added to payload. This also fixes issue where you pass in something serializable, but is not base object

- Added `allowedTypes` as option to devtools. By default **File**, **FileList** and **Blob** is there, but you can extend to whatever you want... allowing more special types. Please add more base types if you think it is relevant

- Any allowed types not having a `toJSON` ability gets one added based on the constructor name. To allow debugger to show it

- Updated docs

Okay, I think this has "all the things" :)

packages/docs/content/in-depth/03_state.en.md
```diff
@@ -50,4 +50,13 @@ const conroller = Controller({
...
 
-And that is basically all there is to state in Cerebral. If you know how to use work with basic JavaScript types, you can work with state in Cerebral.
+Using the plain value types of JavaScript gives Cerebral all its power. That said, there are times you need to store things like files or blobs. Cerebral supports this as well. The reason files are supported, but not **Date** for example, is because a date can be mutated directly. For example:
+
+```js
+const date = new Date()
+date.setHours(8)
+```
+
+There is no way for Cerebral to keep track of this, and therefor is not supported. But do not worry about this. You can still use the Date object in your app, only that when you store it as state you use either a string or a number.
```
Guria: therefore

packages/docs/content/in-depth/03_state.en.md
```diff
@@ -50,4 +50,13 @@ const conroller = Controller({
...
+
+And that is basically all there is to state in Cerebral. If you know how to work with basic JavaScript types, you can work with state in Cerebral.
```
Guria: Hm. Aren't we have to use special api to deal with state?

packages/docs/content/in-depth/04_signals.en.md
```diff
@@ -56,3 +56,5 @@ connect({
   }
 )
 ```
+
+The payload passed to a component is typically the main value types of JavaScript. Object, Array, String, Number or Boolean. It is also possible to pass in some special value types, like files. For a full list of supported value types, check the state API documentation.
```
Guria: link to state API here?

packages/docs/content/api/02_state.en.md
```diff
@@ -19,6 +19,15 @@ const controller = Controller({
...
+When building an application you often need to keep things like files and blobs in your state for further processing. Cerebral supports these kinds of values because they have no API to be mutated, or can use existing state API. This is the list of supported types:
+
+- **File**
+- **FilesList**
+- **Blob**
+
+If you want to force Cerebral to support other types as well, you can do that with devtools options.
+
 ### Get state
 The only way to get state in your application is by connecting it to a component or grabbing it in an action.
```
Guria: Could you please also document state.compute here?

packages/docs/content/in-depth/03_state.en.md
christianalfoni: Yeah, bad wording, changed it :)

packages/docs/content/in-depth/04_signals.en.md
christianalfoni: Jup, done!

packages/docs/content/api/02_state.en.md
christianalfoni: For sure :)

christianalfoni:
Okay, updated your comments @Guria and made some small improvements myself as well :)

PR #530, opened 2016-12-01
feat(demos): add onSubmit, onBlur validation and reset

yusufsafak:
@edgesoft Sorry, I had no time to push it early today :( 
I created a simple task list according to your summary. 

**Simple demo**
- [x]  should be able to validate directly when typing (as it does now)
- [x]  should be able to validate only onBlur
- [x]  should be able to validate only onSubmit.
- [x]  should be able to reset the form 

**Checkout demo**
- [ ] masked inputs like credit cards
- [ ] show how to dynamically add forms and fields ( adding products .. )

**Sign up demo** (confirm password)
- [ ] using dependsOn

**Fileupload demo**
- [ ] using https://github.com/cerebral/cerebral/pull/528
- [ ] support multiple uploads
- [ ] progress bar

**Dependent DropDown List**

### DIFF
--- a/packages/forms-demo/src/components/Fields/Input/index.js
+++ b/packages/forms-demo/src/components/Fields/Input/index.js
@@ -5,8 +5,8 @@ import styles from './styles'
 
 export default connect((props) => (
   {
-    'field': `${props.path}.**`,
-    'settings': 'app.settings.**'
+    field: `${props.path}.**`,
+    settings: 'app.settings.**'
   }),
   {
     fieldChanged: 'simple.fieldChanged'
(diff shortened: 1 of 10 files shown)


packages/forms-demo/src/components/Fields/Input/index.js
```diff
@@ -15,22 +15,30 @@ export default connect((props) => (
...
-          {errorMessage}
+          {showErrors && errorMessage}
         </div>
       )
     }
     return (
       <div style={{marginTop: 10, fontSize: 14}}>
         {name} {field.isRequired ? '*' : ''}<br />
-        <input onChange={(e) => onChange(e)} type={'text'} className={css(styles.input)} />
+        <input onChange={(e) => onChange(e)} onBlur={(e) => onBlur(e)} value={field.value} type={'text'} className={css(styles.input)} />
```
yusufsafak: You should add value properties into input tag otherwise resetForm does not work. Now I found the bug :)

packages/forms-demo/src/components/Menu/index.js
```diff
@@ -27,7 +27,10 @@ export default connect(
         </div>
         <div className={css(styles.settingsContainer)}>
           {Object.keys(settings).map((settingKey, index) => {
-            return <SettingsCheckbox key={index} path={`app.settings.${settingKey}`} />
+            if (settings[settingKey] === Object(settings[settingKey])) {
+              return <SettingsCheckbox key={index} path={`app.settings.${settingKey}`} />
```
yusufsafak: @edgesoft  Except "Disable submit when form is invalid" I think, the other settings option should be a radio button, because the user must select only one choice. What do you think?

edgesoft: @yusufsafak Yeah it could! I began to implement it as clicked some checkbox disable others in array. ```unToggleFieldsWhenChecked: ['app.settings.validateOnChange'] ```

Guria:
is it ready to merge?

edgesoft:
@Guria Yes, it's ready to merge. We will bring in the other features later.

yusufsafak:
@Guria  yes it is ready 
@edgesoft  Today I will try to look into `Sign up demo` 
Do you have any good scenario for multiple fields dependsOn?

edgesoft:
@yusufsafak Not that much that comes into mind more than password, repeat and email.

PR #532, opened 2016-12-02
feat(Operators): add signal operator to pass a signal to action/factory

gaspard:
Usage example:

```js
firebase.upload('remote.path', state`$draft.imageFile`,
  {progress: signal`clients.$draft.imageProgress`}
), {
  success: [],
  error: []
}
```


Guria:
@christianalfoni so are we ok with signals passed to payload?

christianalfoni:
Perfectly okay to action factories, cause already doing that in Firebase
for example... for listeners. Just have to be explicit in docs about you
should not pass signals in and trigger them directly, that would be bad
practice :-)

PR #534, opened 2016-12-02
Uploadfordemo

gaspard:
Add support for image upload in demo (uses latest firebase implementation for upload).


Guria:
need rebase

PR #536, opened 2016-12-02
feat(Controller): flush async state changes

christianalfoni:
To handle async state changes we need to run an async flush on any state change. To avoid doing unnecessary flushes the async flush will of course be reset for every new state change. There is also a general optimization here where flushes which are not forced and has no changes, does not cause a flush to happen (avoid running lots of unnecessary code).

So even though more flushes, code is faster and handles async.

You might say, why not only have async flushes? React and inputs :)

### DIFF
--- a/packages/cerebral/src/providers/State.js
+++ b/packages/cerebral/src/providers/State.js
@@ -18,6 +18,7 @@ function StateProviderFactory () {
 
   function createProvider (context) {
     const model = context.controller.model
+    let asyncTimeout = null
 
     return methods.reduce((currentStateContext, methodKey) => {
       if (methodKey === 'compute') {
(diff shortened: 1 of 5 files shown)


gaspard:
I tested this with upload:

```js
firebase.put(
input`remoteItemImagePath`,
input`imageFile`, {
  progress: state`${input`itemPath`}.$imageProgress`
})
```

Works fine and the UI updates nicely. The debugger shows many state changes, it all makes sense.

packages/cerebral/src/providers/State.js
```diff
@@ -26,6 +27,11 @@ function StateProviderFactory () {
         currentStateContext[methodKey] = (...args) => {
           const path = ensurePath(args.shift())
 
+          if (methodKey !== 'get') {
+            clearTimeout(asyncTimeout)
+            setTimeout(() => context.controller.flush())
```
gaspard: This is very clever to trigger only on async. 👍

Guria: Hmm. Aren't it should assign to asyncTimeout here?

christianalfoni: Oh, shit... yes, you are right, I will rebase :)

PR #537, opened 2016-12-03
feat(Views): add props function and fix warnings

christianalfoni:
- now throws error when a state path definition is undefined, typically when importing computeds wrong
- Should only warn once now for big components
- Now warns when strict render is active
- Added:

```js
connect({}, {}, (ownProps, stateProps, signalProps) => {
  return {} // Overrides everything
}
```

Have not documented this yet, can do it in another pull. Got some other doc updates to do as well

### DIFF
--- a/packages/cerebral/src/viewFactories/Hoc.js
+++ b/packages/cerebral/src/viewFactories/Hoc.js
@@ -1,5 +1,5 @@
 import {Computed} from './../Computed'
-import {cleanPath, propsDiffer} from './../utils'
+import {cleanPath, propsDiffer, throwError} from './../utils'
 
 export default (View) => {
   return function HOC (paths, signals, injectedProps, Component) {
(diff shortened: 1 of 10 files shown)


packages/cerebral/src/viewFactories/Hoc.js
```diff
@@ -99,25 +108,24 @@ export default (View) => {
           if (
             this.context.cerebral.controller.devtools &&
             this.context.cerebral.controller.devtools.bigComponentsWarning &&
+            !this.hasWarnedBigComponent &&
             Object.keys(extractedSignals).length >= this.context.cerebral.controller.devtools.bigComponentsWarning.signals
           ) {
-            console.warn(`Component named ${Component.displayName || Component.name} has a lot of signals, consider refactoring. Adjust this option in devtools`)
+            console.warn(`Component named ${Component.displayName || Component.name} has a lot of signals, consider refactoring or adjust this option in devtools`)
```
henri-hulski: Should `hasWarnedBigComponent` not be set to `True`?

christianalfoni: Just testing you ;-)

henri-hulski: ;-)

PR #538, opened 2016-12-03
docs(tutorial): add details about alpha install and rename folder

christianalfoni:
Just a note on tutorial about alpha install and renamed `parts` folder, as people starts editing inside there


Guria:
@christianalfoni maybe just put a readme in parts folder?

christianalfoni:
Do not think it will help... people look at "chapter 01" and then goes to "parts/01". There is no incentive to check out a readme in parts folder... but "DO_NOT_TOUCH" is a big incentive to think twice ;-)

Happy with other suggestions, though do not think readme is enough

Guria:
@christianalfoni why not put clear instructions in start script which will also tell about that all changes will be lost on chapter switch?

christianalfoni:
It is already there, but people miss it... even though we write things, people are too eager I think :) There is not reason why this folder should be named parts though... it has no meaning to the user. It just helps I think :)

PR #540, opened 2016-12-04
feat(debugger): update to latest cerebral version

christianalfoni:
Published new debugger versions, also uploaded to google drive.

Will look into automatic publish on next release


Guria:
no need to merge it :)

christianalfoni:
Oh? What about manifest version update?

PR #541, opened 2016-12-05
Function tree debugger

christianalfoni:
This branch has been used to build a POC and demo off functiontree by Fricke. It is a good version 1.

- Debugger copied from Cerebral and refactored to work directly with function tree
- Devtools added and refactred for function-tree purpose, with some custom functionality for serialization.. cause in Node you have lots more complex objects passed around


Guria:
@christianalfoni now rebase it :)

Guria:
@christianalfoni is it hard to make debugger universal?

christianalfoni:
@Guria 

I thought about implementing function tree debugging into Cerebral debugger, but it is very problematic to make a good user experience, cause you can have any kind of combination of the two. Also many uncertainties in terms of how the pure function-tree debugger should work. That said, we might be able to merge it in later when we know more how it should work :) Better to start out separate and then merge, than merge first and then try to figure out how it should work.

Chrome extension debugger is not ready yet. I need to figure out how to handle the WebSocket import, which should only happen on Node. I think it will be necessary due to function-tree in combination with Redux etc.

Okay, hopefully it is up and running now. We will start using it in our project as well, so will give this some more attention :)

Fricke demo went great, they will start using function-tree on their BOT service now :)

PR #542, opened 2016-12-06
fix(demos): fix image upload

gaspard:
A small change in firebase provider for upload was not reflected in the demo. This fixes it.


Guria:
@gaspard need rebase to solve latest lint issues

PR #553, opened 2016-12-13
Added yarn for installing Cerebral 2 edge

hipertracker:
Added simpler yarn. It's already used by default in create-react-app, so it makes sense to use it also here.


henri-hulski:
Yarn is optional in CRA. 
And Yarn didn't work for me (yarnpkg/yarn#746). 
It also has some more issues: facebookincubator/create-react-app#1189. 

So I think we should at least document both npm and yarn and make yarn optional.

PR #554, opened 2016-12-14
fix(tutorial): npm install did not work (tested on windows)

fopsdev:
adding "next" - version for the cerebral stuff directly into the package.json solved the install

issue

@todo: the alpha instructions needs to be updated if this gets pulled int

### DIFF
--- a/packages/tutorial/package.json
+++ b/packages/tutorial/package.json
@@ -15,9 +15,9 @@
     "shelljs": "^0.7.5"
   },
   "dependencies": {
-    "cerebral": "2.0.0-0",
-    "cerebral-provider-http": "1.0.0-0",
-    "cerebral-router": "1.0.0-0",
+    "cerebral": "next",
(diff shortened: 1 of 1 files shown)


henri-hulski:
We had that before.  But lerna is not able to create cross-links when using "next". 
Instead it installs all internal dependencies from npm. 
@Guria fixed that at [e6bb6bed](https://github.com/cerebral/cerebral/commit/e6bb6bed3c3817766b867145cf0f592604be9583).

fopsdev:
ah ok, but then we need to adjust the existing versions in package.json because 
cerebral": "2.0.0-0" cannot be found anymore and therefore npm install is failing completely (i used as well a npm show cerebral to doublecheck that)
so maybe we should put the last version something like "2.0.0.0.alpha..." in there and retry

Guria:
need to setup build step which will write actual exact version into tutorial package.json and publish it as zip in github releases. on my list.

PR #569, opened 2016-12-18
Info about strict render should not be a warning

menberg:
As it's just a regular console information, it shouldn't be treated as a warning IMHO. In my build setup, I'm getting a notification each time there are new warning messages on the node server – which is a bit annoying if it's just a regular info like in this case.

### DIFF
--- a/packages/cerebral/src/Controller.js
+++ b/packages/cerebral/src/Controller.js
@@ -63,7 +63,7 @@ class Controller extends EventEmitter {
     if (this.devtools) {
       this.devtools.init(this)
     } else if (
-      process.env.NODE_ENV !== 'production' &&
+      isDeveloping() &&
       typeof navigator !== 'undefined' &&
       /Chrome/.test(navigator.userAgent)
     ) {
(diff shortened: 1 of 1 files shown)


menberg:
By the way, should the info message not only be printed while in development? I get the message also on production, which feels less than ideal...

Guria:
Yes, you are right. Could you add production build check as well?

packages/cerebral/src/Controller.js
```diff
@@ -70,7 +70,10 @@ class Controller extends EventEmitter {
       console.warn('You are not using the Cerebral devtools. It is highly recommended to use it in combination with the debugger: https://cerebral.github.io/cerebral-website/install/02_debugger.html')
     }
 
-    if (this.options.strictRender) {
+    if (
+      process.env.NODE_ENV !== 'production' &&
```
Guria: there is isDeveloping utility in util.js
https://github.com/cerebral/cerebral/blob/master/packages/cerebral/src/utils.js#L72

Guria:
Thanks

PR #572, opened 2016-12-19
feat(Devtools): add collapsable paths and outputs

christianalfoni:


packages/cerebral-debugger/package.json
```diff
@@ -17,8 +17,8 @@
   "repository": "https://github.com/cerebral/cerebral",
   "author": "Christian Alfoni",
   "dependencies": {
-    "cerebral": "2.0.0-0",
-    "cerebral-module-useragent": "1.0.0-0",
+    "cerebral": "2.0.0-alpha.4-alpha.ed1e4568",
```
Guria: doesn't look to be right versions here

christianalfoni: I am not able to build Electron apps without installing dependencies directly to project. I get Babel errors :(

Guria: But it definetely old version. use the same command as we propose on website

PR #576, opened 2016-12-22
feat(http): return headers on response and fix bug

henri-hulski:


packages/cerebral-provider-http/src/utils.js
```diff
@@ -64,3 +64,20 @@ export function convertObjectWithTemplates (obj, context) {
     return convertedObject
   }, {})
 }
+
+export function parseHeaders (rawHeaders) {
```
christianalfoni: Is it necessary to parse the headers? Can you not just grab them? Not quite sure if HTTP should be responsible for it :)

henri-hulski: > The getAllResponseHeaders( ) method returns all HTTP headers from the response of the XMLHttprequest. 
> HTTP headers are returned as a single string with each header line separated by a CR/LF pair and with each header name and header value separated by a ": " pair.

So I don't know if this is a bit useful.
Axios e.g. also parses the headers.

christianalfoni: Ah, they do? Sorry, I thought there was a native method for grabbing them. Okay, but looks good :)

henri-hulski: For axios see https://github.com/mzabriskie/axios/blob/master/lib/helpers/parseHeaders.js.

christianalfoni:
Oh... sorry, I thought I was merging something else, oh well... it is approved :) Sorry @Guria

PR #578, opened 2016-12-24
More view tests and hoc.js bugs

yusufsafak:
@christianalfoni  When I was trying to improve test coverage of hoc.js, I found some bugs. I fixed small ones, but I don't have time to fix one of the bug related to computed.
Computeds are not automatically removed before a component is unmounted (componentWillUnmount). There is a problem at https://github.com/cerebral/cerebral/blob/master/packages/cerebral/src/viewFactories/Hoc.js#L70

### DIFF
--- a/packages/cerebral/src/viewFactories/react.test.js
+++ b/packages/cerebral/src/viewFactories/react.test.js
@@ -157,10 +157,10 @@ describe('React', () => {
           <TestComponent />
         </Container>
       ))
-      const TestComponentRef = TestUtils.findRenderedComponentWithType(tree,TestComponent)
+      const TestComponentRef = TestUtils.findRenderedComponentWithType(tree, TestComponent)
       assert.equal(TestComponentRef._isUnmounting, undefined)
       assert.equal(TestUtils.findRenderedDOMComponentWithTag(tree, 'div').innerHTML, 'bar')
-      ReactDOM.unmountComponentAtNode(TestUtils.findRenderedDOMComponentWithTag(tree, 'div').parentNode);
+      ReactDOM.unmountComponentAtNode(TestUtils.findRenderedDOMComponentWithTag(tree, 'div').parentNode)
       assert.equal(TestComponentRef._isUnmounting, true)
       assert.deepEqual(controller.componentDependencyStore.getAllUniqueEntities(), [])
     })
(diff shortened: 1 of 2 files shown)


packages/cerebral/src/viewFactories/react.test.js
```diff
@@ -136,10 +137,182 @@ describe('React', () => {
...
+        return (
+          <div>{props.foo}</div>
+        )
+      })
+      const tree = TestUtils.renderIntoDocument((
+        <Container controller={controller}>
+          <TestComponent />
+        </Container>
+      ))
+      // how to get hasWarnedBigComponent or console.log??
```
yusufsafak: How can we test hasWarnedBigComponent? should we test console outputs?

christianalfoni:
@yusufsafak Great stuff!

About warning, you can override `console.warn` temporarily for the test and check if it is called :)

About the computed bug, you have some more info on that? I do not see where there is an issue?

Guria:
A bit of linting issues should be fixed

yusufsafak:
Sorry guys for late reply. I was in a hurry to go holiday :) I pushed it so late and I couldn't explain it well. 
I am on mobile phone now and there is no good connection:( I just commented out two console tests to fix the lint errors. 
For the computed bug you can check this test https://github.com/cerebral/cerebral/blob/cb852e91d1b10dc098972530dec2a9a92c4d6d0c/packages/cerebral/src/viewFactories/react.test.js#L225 and coverage. Unmounting the component has never touch this line https://coveralls.io/builds/9429704/source?filename=packages%2Fcerebral%2Fsrc%2FviewFactories%2FHoc.js#L63 because there is no computed instance in depsmap. https://github.com/cerebral/cerebral/blob/master/packages/cerebral/src/viewFactories/Hoc.js#L70. It has only computed's depsmap.

christianalfoni:
Looks great! Let me have a check on those commented out tests

Guria:
@christianalfoni any news?

Guria:
@christianalfoni revisit this please

christianalfoni:
@Guria I looked into the branch and there has been so many changes to HOC that I think we need to throw this PR. A lot of the tests was related to previous signature, Computed etc. :(

Really sorry @yusufsafak!

yusufsafak:
No problem!
Thanks for the new Computed and the other changes. They are really good :)

PR #579, opened 2016-12-25
WIP: Tags

christianalfoni:
- [x] refactored tags
- [x] refactored todomvc
- [x] refactored firebase
- [x] refactored http
- [x] refactored tutorial
- [x] Updated docs
- [x] refactored demo

### How it works
Operator tags has now been moved out into its own "TAGS" concept. This allows us to use the tags also in computeds and connect.

```js
import {state, input, signal, props, string} from 'cerebral/tags'
```

Making tags its own concept makes the API of Cerebral more consistent, but more importantly empowers dependencies. For example:

```js
connect({
  item: state`items.${state`currentItemKey`}`
}, ...)
```

And you can combine this any way you want, for example dynamic signals based on state:

```js
connect({
  clicked: signal`${state`currentModule`}.clicked`
}, ...)
```

Read more in the updated docs :)

#### model.compute
The `model.compute` method is removed. The reason is that computeds does not really have anything to do with the model itself, comuteds only needs access to state. Due to how tags now work I decided to remove the method and rather create "tag getters" like connect does.

#### force recompute
I took this opportunity to flip how "force recompute" works. Up until now when using `state.compute` inside an action it will use the cached value. It is rather difficult to reason about this, cause state changes made before the `state.compute` usage will not bust the cache. To make it more predictable it will by default always bust the cache with `state.compute`, but you can tell it to use the cache with `state.compute(someComputed, true)` to optimize, if needed.


christianalfoni:
Okay, so I did a refactor of Tags "under the hood". Now it is class based. Which means you can check for tags like this:

```js
import {Tag} from 'cerebral/tags'

something instanceof Tag
```

Also now you can instantly check `tag.type`. Docs are updated as well, so can read that under "in depth" :)

gaspard:
@christianalfoni I think we should refactor `Tag` internals to make it easier to create custom tags. I have this now but it will not longer work and it's a pity:

```js
import {elapsedSeconds as computeElapsedSeconds} from '../../../helpers/dateTime'

export default function (startedAtTemplate, endedAtTemplate) {
  function elapsedSeconds (context) {
    const startedAt = typeof startedAtTemplate === 'function'
      ? startedAtTemplate(context).value : startedAtTemplate
    const endedAt = typeof endedAtTag === 'function'
      ? endedAtTemplate(context).value : endedAtTemplate

    return {value: computeElapsedSeconds(startedAt, endedAt)}
  }

  elapsedSeconds.displayName = 'elapsedSeconds'

  return elapsedSeconds
}
```

gaspard:
OK, I refactored demo to not use the operator helpers for now.

I found some bugs in firebase regarding `signal` handling in onChildAdded, onChildChanged, etc.

gaspard:
OK, I managed to fix all firebase issues I had with the demo. The code feels ugly at the moment (onChildAdded and similar factories) but I think @christianalfoni will refactor this anyway so it's just some kind of pointer.

FYI, firebase.onChildAdded requires a signal path to avoid double init (not sure we need this check). The code is in helpers, listenTo.

christianalfoni:
Okay, I believe everything is now fixed here.

I also fixed forms and forms demo.

Two new changes:

1. Now Cerebral has a `resolveArg`provider. This allows you to resolve arguments in factories.

```js
function myFactory (someArg) {
  function myAction ({resolveArg}) {
    // Is it a tag?
    const isTag = resolveArg.isTag(someArg)
    // Is it a specific tag?
    const isTag = resolveArg.isTag(someArg, 'state', 'input')    
    // Resolves the path if it is a tag, throws error if not tag
    const isTag = resolveArg.path(someArg)
    // Resolves the value of the arg, it being a tag or any other value
    const isTag = resolveArg.value(someArg)

    // if arg is function the context of current action is passed in, allowing
    // you to use plain functions where tags are allowed, where the returned
    // value of the function is the value resolved example:
    // set(state`foo`, someValueBasedOnContext), or a factory
    // set(state`foo`, someValueFactory(state`bar`))
  }
}
```

2. Forms can not use `input` as a factory parameter, because input holds path to state, not the value itself. It create an inconsistency. So you always have to `isValidForm(state`${input`path`}`)`

christianalfoni:
Okay, ran this tag on our project and it is working :)

Guria:
@christianalfoni do we really need isStateDependency flag? Why not just `tag.type === 'state'` ?

christianalfoni:
@Guria Because other tags might also express state dependencies, like:

```js
connect({
  form: form`path.to.form`
})
```

:)

packages/cerebral/src/tags/Tag.js
```diff
@@ -0,0 +1,99 @@
...
+    } else {
+      return this.getPath(getters)
+    }
+  }
+  /*
+    Extracts value from object using a path
+  */
+  extractValueWithPath (obj, path) {
+    return path.split('.').reduce((currentValue, key) => {
+      // if (currentValue[key] === undefined) {
```
Guria: commented code left

christianalfoni: Ops, how did that happen? Fixed now

christianalfoni:
Okay, about the tag discussion. I actually think we should keep it "as is". My reasoning:

- The `context.resolveArg` is not a helper function, it actually uses the context to resolve the argument (access to state, input and signals). The alternative `getValue(context, argA)`, is giving the helper a context... so yeah, I think it is better to keep actions imperative in that regard

- We want to resolve more than a tag... we want to handle "non-serializable" arguments, which means there might be other kinds of special values we want to resolve later. Importing general helpers does not fit in with current imports

- It is low level, for library authors... not a huuuge deal ;-)

FWeinb:
I just watched the hangout and thought about this too. Is there a problem to pass in a object of "helper" functions as the second parameter so you would access it like:

```js
function myActionFactory(argA){
    function myAction(context,{getValue,getPath}){
      const value = getValue(argA);
       ....
    }
    return myAction;
}

I like to access these helpers explicitly using `getValue`and `getPath`.

As you said this is low-level stuff so no one needs to care about a second parameter. But I am probably missing something here.

**Comment by gaspard** [comment]
OK, to summarise, we have:

**== A. import helper ==**

```js
import {getValue} from 'cerebral/tags'

function elapsed (startedAt, endedAt) {
  return function (context) {
    return getValue(context, endedAt) - getValue(context, startedAt)
  }
}
```

**Pros**

- no changes in context
- special import makes it more "dev specific"
- not clear from where this should be imported

**Cons**

- have to explicitly pass context argument
- cannot use destructuring with context

**== B. resolver in context ==**

```js
function elapsed (fromTime, toTime) {
  return function ({resolve}) {
    console.log(resolve.path(fromTime), resolve.path(toTime))
    return resolve.value(toTime) - resolve.value(fromTime)
  }
}
```

**Pros**

- nothing to import in context, easier to "just add tag support"
- can use destructuring with context
- slightly easier to implement

**Cons**

- adds a pure helper function in context (not very orthogonal with input, state, etc)

**==== Conclusion, personal opinion ====**

I like the "resolve" or "resolveArgument" thing in context (solution B). I think it encourages people to add support for tags and helper functions and I think this is a very good way to add abstraction in signals.

I vote B...

henri-hulski:
@gaspard I also like the resolve solution. 

And +1 for using "resolve"  instead of "resolveArg". Looks much cleaner for me.

christianalfoni:
Thanks for the list @gaspard , nice summary and yes, that is my reasoning as well :) I like `resolve`, better but it is not explicit enough I think. You may resolve many things... maybe we should call it: `resolveFactoryArgument`... because that is exactly what it is. It is there to help lib developers use and internals of Cerebral to resolve factory arguments

Guria:
I think is better to mention tag in name `resolveTag` or `resolveAsTag`

christianalfoni:
@Guria But it is not for resolving tags :) It is generally to resolve arguments, them being just a string, array, function or tag... it is to avoid writing IFs and typeof everywhere... you usually just want the value of the thing, it being a value already, a function or a tag. We save soooo much refactoring ahead making changes to this :)

christianalfoni:
@Guria @gaspard @henri-hulski What about:

```js
function elapsed (fromTime, toTime) {
  return function ({resolve}) {
    const resolvedFromTime = resolve.tagValue(fromTime) || resolve.functionValue(fromTime) || fromTime
    const resolvedToTime = resolve.tagValue(toTime) || resolve.functionValue(toTime) || toTime

    return resolvedToTime - resolvedFromTime
  }
}
```

This way API would be:

- resolve.tagValue
- resolve.tagPath
- resolve.functionValue

This requires more syntax, but more explicit as well. You can also fine tune what arg you support. Problem though is if the resolved value is undefined or null... where it would probably do the wrong thing.

I still like `resolveFactoryArgument` better... after some use and we come up with something better it is so simple to just deprecate it... because it is just sone methods. Doing instanceof, typeof etc... way harder to handle as we move on.

gaspard:
@christianalfoni I think it is best if we do not have to figure out what thing this is from the "outside", this way if there are new ways to pass values we simply enhance the resolver and we do not need to update code everywhere. It also avoids situations where a dev thought about supporting one type of value passing but not another and then user experience is weird because things do not work the same way everywhere.

I think "resolveFactoryArgument" is too much. I also no longer consider functions that create functions as "factories" (this term is linked to object oriented programming for me). It is actually some kind of currying. So it would be "resolveCurriedArguments" for me but then,"resolveArgument" is good and simple. So "resolveArgument" for me :-).

christianalfoni:
@gaspard @Guria @henri-hulski 

Okay, cool, I also vote for `resolveArgument`. We need to conclude on this to move on. And as @gaspard confirmed here now, the safest way to move forward with this is to have it on `context.resolveArgument`, cause that can be modified without changing API. And since: `someHelper(context, arg)` basically is: `someHelper.call(context, arg)`, it is not a pure helper. It is handling the `arg` argument with its context.... `context` is an actions `this`.

So I vote for just renaming and move on :-) As we also talked about, this is low level stuff!

Guria:
Ok. I am in. Personally for me `resolveArg` is explicit enough in compare with `resolveArgument`. But it is not very strong opinion but just personal preference.

henri-hulski:
Agree. Same opinion as @Guria.

gaspard:
I am for avoiding abbreviations. They are like a virus and tend to spread and they do not add any real value. I vote for "resolveArgument" :-)

Guria:
@christianalfoni now your move

christianalfoni:
Let us keep `resolveArg`, for two reasons... the first being lazy, I already refactored this twice :) The second being that `arg` is more common than `argument`. `argv` from node, `args` typically in all function spread.

Even in our examples: `argA`, `argB`... :)

packages/cerebral/src/tags/Tag.js
```diff
@@ -0,0 +1,100 @@
...
+    } else {
+      return this.getPath(getters)
+    }
+  }
+  /*
+    Extracts value from object using a path
+  */
+  extractValueWithPath (obj, path) {
+    return path.split('.').reduce((currentValue, key) => {
+      if (currentValue[key] === undefined) {
```
Guria: @gaspard please tell us why you [commented out this part](https://github.com/cerebral/cerebral/pull/579/commits/a4b0bd1d8847cc435d266570bcdd10d871ae90ec). What it stands for? Can we proceed now and merge or smth wrong with this piece?

gaspard: The problem is when testing a value with code like this:

```js
when(input`foo`)
```

If foo is undefined, the code raises an issue. This could be fixed by having a special get for truthy/existence values, like "resolveArg.truthy()" and this code would not throw. Since we do not have this for now, throwing breaks code without any way to work around it.

christianalfoni: Ah, it is supposed to throw when you define an invalid path...  not an invalid value... we have to check the index there as well. Should throw on:

```js
input`foo.bar` // Undefined path, throw

input`foo` // should not throw as it is an undefined value
```

PR #580, opened 2016-12-26
(http): update docs and add response headers to reject path

henri-hulski:


henri-hulski:
I added response headers also to reject path.
This makes sense e.g. for status code 401 which requires a WWW-Authenticate header field.

> **401 Unauthorized** Similar to 403 Forbidden, but specifically for use when authentication is required and has failed or has not yet been provided. The response must include a WWW-Authenticate header field containing a challenge applicable to the requested resource.

packages/cerebral-provider-http/README.md
```diff
@@ -170,12 +170,15 @@ function someGetAction ({http}) {
     .then((response) => {
       response.status // Status code of response
       response.result // Parsed response text
+      // The response headers are returned as an object with lowercase header
+      // names as keys. Values belonging to the same key are seperatet by ', '.
```
Guria: seperatet -> separated

henri-hulski: @Guria Fixed.

Guria:
Thanks :)

Issue #560, opened 2016-12-15
Computed state layers


maxfrigge:
As proposed by @garth one could consider using computed layers on top of the state. 

Let's collect thoughts here.

maxfrigge:
Thoughts I had were:

- could decreases visibility of dependencies
- would eliminate the risk of circular dependencies

garth:
From the point of view of the debugger, it would be nice to see the final state (state + computed layers) and the step down through the layers until you reach the original state.

Here's just a rough idea of what connect could look like (just to get the conversation started)

```js
connect([
  { foo: 'foo.path' },
  { foo: augmentFoo() },
  { foo: sortAugmentedFoo() }
], {
  signal: 'foo.signal'
}, (stateProps, signalProps, ownProps) => ({
  originalFoo: stateProps[0].foo,
  foo: stateProps[2].foo
}), (props) => {
  ... 
})```

**Comment by Guria** [comment]
Can't get what it really gives us. Could you give example on how you will use it

**Comment by maxfrigge** [comment]
@garth So you wanted to use those layers on a per component base and not for the globale state object?

**Comment by garth** [comment]
@maxfrigge yes, I think that the control should stay with connect so that you maintain visibility of what is going on (no magic), also each component may need to compose computeds in different ways which may conflict in a global approach.

@Guria maybe it's best to think about what we are trying to solve:

### Level 0
I have an unordered set of people
```js
{
  '1': { fistName: 'c', lastName: 'd', dob: new Date() },
  '2': { fistName: 'a', lastName: 'b', dob: new Date() }
}
```

### Level 1
Then I have a computed that will add `fullName` and `age`
```js
{
  '1': { fistName: 'c', lastName: 'd', dob: new Date(), fullName: 'c d', age: 0 },
  '2': { fistName: 'a', lastName: 'b', dob: new Date(), fullName: 'a b', age: 0 }
}
```

### Level 2
Then I have a computed that will sort a set by given properties (eg `['fullName']`)
```js
[
  { fistName: 'a', lastName: 'b', dob: new Date(), fullName: 'a b', age: 0 }
  { fistName: 'c', lastName: 'd', dob: new Date(), fullName: 'c d', age: 0 }
]
```

### There are a number of things that I want to achieve

1. If I add/remove a person the other Level 1 computeds should not re-run except for the newly added item. The level 2 computeds should re-run.
2. If I change level 2 computeds (eg change sort direction) level 1 computeds should not re-run.
[...]

PR #582, opened 2016-12-28
fix(Computed): computed too many recalcs

fopsdev:
computed code gets evaluated too many times

### DIFF
--- a/packages/cerebral/src/Computed.test.js
+++ b/packages/cerebral/src/Computed.test.js
@@ -99,10 +99,10 @@ describe('Computed', () => {
           foo
         }
       })
-      const value = computed.getValue(model)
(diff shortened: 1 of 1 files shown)


fopsdev:
this is the failing test.

fopsdev:
bump ;)

Guria:
https://github.com/cerebral/cerebral/issues/560#issuecomment-269723521

fopsdev:
@Guria imo this test makes sense anyway. so feel free to pull it in

Guria:
Pulling it in will break CI releases :(
We also need a fix as well.

cc @christianalfoni

christianalfoni:
Very soon moving to new computed approach, just wait for that?

PR #587, opened 2017-01-06
feat(firebase): delete user and throw errors correctly

christianalfoni:
- Can not return values from `catch` as they will be no longer propagate as errors, have to reject or throw again

- Wrapped a few methods with normal promises, cause Firebase promise is not native

### DIFF
--- a/packages/cerebral-provider-firebase/src/createTask.js
+++ b/packages/cerebral-provider-firebase/src/createTask.js
@@ -5,20 +5,27 @@ export default function createTask (options) {
     return new Promise((resolve, reject) => {
       const tasksPath = options.queuePath ? options.queuePath + '/tasks' : 'queue/tasks'
       const ref = firebase.database().ref(tasksPath)
-      const taskKey = ref.push(Object.assign({
-        _state: options.specPrefix ? `${options.specPrefix}_${name}` : name
-      }, payload))
 
-      const taskRef = firebase.database().ref(`${tasksPath}/${taskKey.key}`)
-      taskRef.on('value', data => {
-        if (!data.val()) {
-          taskRef.off()
-          resolve()
-        }
-        if (data.val() && data.val()._error_details) {
-          reject(data.val()._error_details)
-        }
-      })
+      return firebase.auth().currentUser.getToken()
+        .then((_token) => {
+          const taskKey = ref.push(Object.assign({
+            _state: options.specPrefix ? `${options.specPrefix}_${name}` : name,
+            _token
+          }, payload))
+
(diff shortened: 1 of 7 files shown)


packages/cerebral-provider-firebase/src/deleteUser.js
```diff
@@ -0,0 +1,9 @@
+import firebase from 'firebase'
+
+export default function deleteUser (password) {
+  return new Promise((resolve, reject) => {
+    const user = firebase.auth().currentUser
+
+    return user.delete().then(() => resolve(), (error) => reject({error}))
```
edgesoft: Can user be null here? According to docs "You can also get the currently signed-in user by using the currentUser property. If a user isn't signed in, currentUser is null:"

packages/cerebral-provider-firebase/src/sendPasswordResetEmail.js
```diff
@@ -0,0 +1,9 @@
+import firebase from 'firebase'
+
+export default function sendPasswordResetEmail (email) {
+  return new Promise((resolve, reject) => {
+    firebase.auth().sendPasswordResetEmail(email)
+      .then(() => resolve())
+      .catch((error) => reject({error}))
```
gaspard: This should be (catch does not work with firebase own Promises):

```js
 firebase.auth().sendPasswordResetEmail(email)
      .then(
        () => resolve(),
        (error) => reject({error})
      )
```

christianalfoni: Great catch, thanks! :)

christianalfoni:
Okay, this should now be fixed... We just realized Google has updated Firebase and there is a new approach to authentication on the backend side of things. We need to pass the token on every Task to identify the user on the server. A small detail, but necessary :)

packages/cerebral-provider-firebase/src/createTask.js
```diff
@@ -5,20 +5,25 @@ export default function createTask (options) {
...
+      return firebase.auth().currentUser.getToken()
+        .then((_token) => {
+          const taskKey = ref.push(Object.assign({
+            _state: options.specPrefix ? `${options.specPrefix}_${name}` : name,
+            _token
+          }, payload))
+
+          const taskRef = firebase.database().ref(`${tasksPath}/${taskKey.key}`)
+          taskRef.on('value', data => {
+            if (!data.val()) {
```
gaspard: maybe add:

```js
const value = data.val()
```

Not sure about the cost of "data.val()"... But maybe it's very small.

PR #588, opened 2017-01-06
fix(forms): Fix a bug and README

henri-hulski:


henri-hulski:
@Guria why have you closed this? 
It contains actually an important fix. 
At the moment when I have a global prop with value 'null' it throughs. This fix it.

Guria:
Ops. sorry. I think I was about to merge it. Strange

PR #589, opened 2017-01-08
feat(router): add getSignalUrl back into router

idream3:
Using this feature throughout our app, it is needed for dynamic route display.

Usage:
```
const getUrl = controller.router.getSignalUrl;
```


Guria:
Router would be updated to support new tags in 2.0. I am also plan to return getSignalUrl as well.

idream3:
Thanks, sounds good. Do you have an ETA for 2.0?  I am upgrading our app to
Cerebral 2, but rely heavily on getSignalUrl.

Guria:
@idream3 hope to update router on this week

Guria:
Sorry about failed estimations. Now when computed stuff is about to lend it seems turn for router update

edgesoft:
Think this is valid. If you use baseUrl you need to make you own function to support it so this is great.

PR #594, opened 2017-01-11
fix(provider-http): Use resolveArg instead of getters in convertObjectWithTemplates

FWeinb:


Guria:
https://travis-ci.org/cerebral/cerebral/builds/190981287#L201

PR #595, opened 2017-01-12
feat(Module): preserve module instance

Guria:


packages/cerebral/src/Module.js
```diff
@@ -1,31 +1,35 @@
...
-    this.modules = {}
+  constructor (controller, path, moduleDescription) {
+    const moduleStub = {
+      controller,
+      path,
+      name: path.slice().pop()
+    }
 
-    const moduleDescription = typeof module === 'function' ? module(this) : module
+    const module = typeof moduleDescription === 'function' ? module(moduleStub) : moduleDescription
```
Guria: Probably it's better to split moduleStub into separate arg here

packages/cerebral/src/Module.js
```diff
@@ -1,31 +1,35 @@
...
-      this.modules[moduleKey] = new Module(this.path.concat(moduleKey), modules[moduleKey], this.controller)
-    })
+    module.signals = module.signals || {}
+
+    module.modules = Object.keys(module.modules || {}).reduce((registered, moduleKey) => {
+      registered[moduleKey] = new Module(controller, path.concat(moduleKey), module.modules[moduleKey])
+      return registered
+    }, {})
+
+    ;(module.prototype || module).getProviders = this.getProviders
```
Guria: getProviders could be moved out to utility

packages/cerebral/src/Module.js
```diff
@@ -1,35 +1,24 @@
...
 
-    const moduleDescription = typeof module === 'function' ? module(this) : module
+    const module = typeof moduleDescription === 'function' ? moduleDescription(moduleStub) : moduleDescription
 
-    this.initialState = moduleDescription.state || {}
-    if (moduleDescription.state) {
-      this.controller.getModel().set(this.path, moduleDescription.state)
-    }
+    module.initialState = module.state || {}
+    controller.getModel().set(path, module.initialState)
```
Guria: @christianalfoni 
I see no usage of module.initialState.
Could we replace it with just:
```js
controller.getModel().set(path, module.state || {})
```

packages/cerebral/src/Module.js
```diff
@@ -1,35 +1,24 @@
...
-    this.modules = {}
+  constructor (controller, path, moduleDescription) {
+    const moduleStub = {
+      controller,
+      path: path.join('.'),
+      name: path.slice().pop()
+    }
 
-    const moduleDescription = typeof module === 'function' ? module(this) : module
+    const module = typeof moduleDescription === 'function' ? moduleDescription(moduleStub) : moduleDescription
```
christianalfoni: Do not understand that this works? You have to use `new moduleDescription(moduleStub)` for classes to work. So check if moduleDescription is a a basic object, if not use `new`... does not matter if it is a class or a plain function as you return a complex object which will be used even if you do a "new" on it...

Guria: `moduleDescription(moduleStub)` returnes either moduleDescription like object or any class instance with special supported properties (state, signals, provider, modules).
Then we just return recieved object/instance from constructor to make "module" instance.
Just see a test and
https://github.com/cerebral/cerebral/blob/3e311f0/packages/cerebral-router/src/index.js#L33
https://github.com/cerebral/cerebral/blob/3e311f0/packages/cerebral-router/src/router.js#L5

packages/cerebral/src/Module.js
```diff
@@ -1,35 +1,24 @@
...
-  /*
-    Used by controller to extract all the providers of all modules
-  */
-  getProviders () {
-    return (this.provider ? [this.provider] : []).concat(Object.keys(this.modules)
-      .reduce((nestedProviders, moduleKey) => {
-        return nestedProviders.concat(this.modules[moduleKey].getProviders())
-      }, [])
-    )
+    module.signals = module.signals || {}
```
Guria: Also seems redundant. Would like to remove it and just secure controller.getSignal instead

packages/cerebral/src/Module.js
christianalfoni: Yeh, that makes sense :)

packages/cerebral/src/Module.js
christianalfoni: Ah, so you did not want to define the module as a class... Just return class instance instead of object... okay, cool. You needed chnages for this? That should have been possible already? Anyways, nice cleanup! :)

Guria: Yes, changes was needed. Because there was creating new object instance by copy some properties from description object: https://github.com/cerebral/cerebral/pull/595/files#diff-da938b16e674eaf4bf8d7cd785d662caL15.

In fact main trick here is return existing object from constructor.

PR #596, opened 2017-01-12
fix(firebase): pass payload also on onValue and onChildChanged

christianalfoni:
- Can now pass payload on other listeners as well
- `onValue` does not trigger initial value for more predictable behaviour... it is a listener for change, not grab initial value.. that is why you have `value`

### DIFF
--- a/packages/cerebral-provider-firebase/src/createOnChildChanged.js
+++ b/packages/cerebral-provider-firebase/src/createOnChildChanged.js
@@ -11,10 +11,20 @@ export default function createOnChildChanged (controller) {
       'child_changed',
       signal,
       (data) => {
-        controller.getSignal(signal)(Object.assign({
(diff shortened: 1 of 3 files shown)


packages/cerebral-provider-firebase/src/createOnChildChanged.js
```diff
@@ -11,10 +11,20 @@ export default function createOnChildChanged (controller) {
...
-        }, options.payload))
+        }
+        let payload = initialPayload
+
+        if (options.payload) {
+          payload = Object.keys(options.payload).reduce((payload, key) => {
+            payload[key] = options.payload[key]
+
+            return payload
+          }, initialPayload)
```
edgesoft: Maybe set {key, value} here directly?

PR #598, opened 2017-01-13
fix(tags): Improve error messages and remove getters from getTags

FWeinb:


packages/cerebral/src/tags/Tag.js
```diff
@@ -15,18 +15,14 @@ export default class Tag {
     in components
   */
   getTags (getters) {
```
Guria: seems argument could be removed as well. and in getNestedTags as well

FWeinb: Ups., will do that.

Guria:
Thanks a bunch

PR #599, opened 2017-01-13
WIP: Improve production bundle

FWeinb:
There where some opportunities to save same memory and and possible minor performance gains. This make the code a little bit more complicated to read. This saved approx. 512 Byte.


packages/cerebral/src/Controller.js
```diff
@@ -2,7 +2,7 @@ import DependencyStore from './DependencyStore'
 import FunctionTree from 'function-tree'
 import Module from './Module'
 import Model from './Model'
-import {ensurePath, isDeveloping, throwError, isSerializable, verifyStrictRender, forceSerializable, isObject, getProviders} from './utils'
```
Guria: isDeveloping could be a constant exposed from utils instead of function. It should do a trick with code elimination and still keep sources readable.

packages/cerebral/src/Controller.js
```diff
@@ -24,13 +24,21 @@ class Controller extends EventEmitter {
     this.componentDependencyStore = new DependencyStore()
     this.options = options
     this.flush = this.flush.bind(this)
-    this.devtools = devtools
-    this.model = new Model({}, this.devtools)
+
+    // Only allow devtools in development
+    if (process.env.NODE_ENV !== 'production') {
```
Guria: Hmm. Looks like it would break our demos. Because they are built with production environment set, but we still want it to support debugger.

packages/cerebral/src/Controller.js
```diff
@@ -41,11 +49,11 @@ class Controller extends EventEmitter {
         this.router.provider
       ] : []
     ).concat((
-      this.devtools ? [
+      process.env.NODE_ENV !== 'production' ? [
```
Guria: Does it enough to eliminate unused DebuggerProvider import as well?

packages/cerebral/src/Controller.js
```diff
@@ -71,21 +79,16 @@ class Controller extends EventEmitter {
...
-      console.warn('You are not using the Cerebral devtools. It is highly recommended to use it in combination with the debugger: https://cerebral.github.io/cerebral-website/install/02_debugger.html')
-    }
-
-    if (
-      isDeveloping() &&
-      this.options.strictRender
-    ) {
-      console.info('We are just notifying you that STRICT RENDER is on')
+    if (process.env.NODE_ENV !== 'production') {
+      if (this.devtools) {
```
Guria: Condition could be combined to avoid extra nesting

FWeinb: I just tested combining these expression but than uglify wont remove the branch.

packages/cerebral/src/Controller.js
FWeinb: I think we need to evaluate if it is necessary to include devtools support in production or not. In the docs devtools are short circuited like this:
`devtools: process.env.NODE_ENV === 'production' ? () => {} : Devtools()`

Guria: Exactly. Now it's leaved user to decide condition on when to include devtools. Proposed change will not leave user a choice to have devtools in production build.

packages/cerebral/src/Controller.js
Guria: Got it

FWeinb:
I am open for ideas to approach this. There are many code paths that, in production, aren't needed. It would be great to have the ability to remove these during build.

The first thing that should be discussed is if devtools should work in production mode or not.

Guria:
cc @christianalfoni

christianalfoni:
Hm, yeah... @Guria: is it only about demos? We could run them in a "non-production" environment or just override NODE_ENV when building them?

Guria:
@christianalfoni I am not sure if it possible with creact-react-app that we using for demos. As I remember it force production env when building files.

edgesoft:
Also create-react-app still uses webpack 1 so no tree shaking out of the box.

Guria:
seems pr became a bit outdated. let's return to this question when we are closer to rc. I will add an action item to todo

PR #601, opened 2017-01-15
fix(Model): fix issue where same path is updated within same flush

christianalfoni:


Guria:
@christianalfoni lint issues

christianalfoni:
@Guria I am still not able to lint on my machine... is there some other linter we can use that just has the same set of rules? Not getting any response here yet: https://github.com/feross/standard/issues/746

Anyways, checked travis and fixed issues

PR #605, opened 2017-01-17
docs(migration): Updated migration guide after migrating real world app

garth:


docs/docs/content/migration/01_start_migration.en.md
```diff
@@ -389,17 +427,18 @@ export default connect({
 )
 ```
 
-In 1.x you had all signals in props. In Cerebral 2.x we favor to be explicit about what signals you need.
+In 1.x you would first connect state, then siganls as seperate arguments in connect. In Cerebral 2.x these are combined, tags are used to differentiate between state, signals and props.
```
henri-hulski: Typo: siganls

docs/docs/content/migration/01_start_migration.en.md
```diff
@@ -104,35 +108,38 @@ export default {
...
+  modules: {
+    subModule: SubModule()
+  },
   signals: {
     newItemTitleSubmitted: submitNewItemTitle
   }
 }
 ```
 
 ### Operators
```
Guria: Would be great to add copy -> set and when -> equals changes

garth: Yes, this was just my first pass. Will also need to update again when computed functions are changed.

PR #607, opened 2017-01-17
WIP: DependencyTracker

christianalfoni:
- [x] Removed computed
- [x] Removed function as optional argument type for connect (no need for it anymore)
- [x] Implemented dependency tracker
- [x]  Optimized HOC code
- [x]  Strict is only render mode
- [x]  Shows warning when depending on Array/Object with no child interest
- [x] Docs
- [x] Update demos
- [x] Update tutorial
- [x] Update TodoMVC
- [ ] Bring back deprecated computed (if necessary)

The final signature is:

```js
connect({
  dynamic: (getProp, getState) => {
    return getProp('foo') + getState('bar')
  }
})
```

We also need to track use of props, which is why it is a function, to optimize when the dependency function should be called again. It is just more consistent as well.

I dropped "getSignal" cause there is no use case for it. Any dynamic nature of grabbing a signal can most surely be handled by tags. Also creates more consistency with `state.get`, where there is not signals.

So now you can have complete control of how component should depend on state. Grab some state dependency, based on the value create some new dependencies and so on. Everything will run optimally :)


edgesoft:
Will have a more detailed look tomorrow, but it looks good 👍

edgesoft:
@christianalfoni Remove WIP

docs/docs/content/tutorial/10_compute_state.en.md
```diff
@@ -19,23 +19,18 @@ function setStarsSum({state}) {
 
 This is a perfectly okay approach for our simple scenario, but computing state like this can be tedious in large applications. We might want to use this state multiple places in our application and we want to make sure it is the same wherever we use it.
 
-In Cerebral we can automatically compute state using a **Computed**.
+In Cerebral we can automatically compute state by using a function. The function can be used by both **connect** and **state.get** inside actions. Let us look at the signature of the function:
 
 ```js
-import {Computed} from 'cerebral'
-import {state} from 'cerebral/tags'
+function starsSum (props, getState) {
```
Guria: getProps?

packages/cerebral/src/Controller.js
```diff
@@ -33,6 +31,10 @@ class Controller extends EventEmitter {
     }, this)
     this.router = router ? router(this) : null
 
+    if (options.strictRender) {
+      console.warn('DEPRECATION - No need to use strictRender option anymore, it is the only render mode now')
```
Guria: Users of non strict mode would be unwarned about change.
Would be automatic "strict" warnings would be enough?
Maybe just warn everyone until release?

packages/cerebral/src/Module.js
```diff
@@ -22,6 +22,24 @@ class Module {
     })
   }
   /*
+    Converts arrays to actually runable signals
+  */
+  registerSignals () {
```
Guria: Looks like you forgot to rebase onto latest Module changes

Guria:
Need rebase and fix bunch of [lint issues](https://travis-ci.org/cerebral/cerebral/jobs/192851838#L831-L857)

packages/cerebral-provider-firebase/src/createOnValue.js
```diff
@@ -4,16 +4,32 @@ import {
...
       createRef(path, options),
       path,
       'value',
       signal,
       (data) => {
-        controller.getSignal(signal)({
+        if (!hasEmittedInitialValue) {
+          hasEmittedInitialValue = true
+          return
+        }
```
gaspard: I don't understand this code. Shouldn't it be:

```js
if (hasEmittedInitialValue) {
  return
}
hasEmittedInitialValue = true
```

packages/cerebral-provider-firebase/src/createOnValue.js
```diff
@@ -4,16 +4,32 @@ import {
...
-        })
+        }
+        let payload = initialPayload
+
+        if (options.payload) {
+          payload = Object.keys(options.payload).reduce((payload, key) => {
+            payload[key] = options.payload[key]
+
+            return payload
+          }, initialPayload)
```
gaspard: Reduce does not make sense for me here and makes the code harder to understand. Why not simply use:

```js
const payload = Object.assign({}, {value: data.val()}, options.payload)
```

christianalfoni:
No idea what happened to rebase... I rebased, merged in, but could not continue rebase... had to commit and then continue... oh man my GIT skills suck

packages/cerebral/src/DependencyTracker.test.js
```diff
@@ -0,0 +1,46 @@
...
+      return getState('foo') + getProp('bar')
+    }
+    const tracker = new DependencyTracker(func)
+    tracker.run(() => 'foo', () => 'bar')
+    assert.equal(tracker.value, 'barfoo')
+    assert.deepEqual(tracker.stateTrackMap, {foo: true})
+    assert.deepEqual(tracker.propsTrackMap, {bar: true})
+  })
+  it('should be able to match changes map', () => {
+    const func = (getProp, getState) => {
```
fmal: looks like `func` can be removed since it's not used in the test 😃

packages/cerebral-provider-firebase/src/createOnValue.js
christianalfoni: AH, hehe... yeah, seems a bit weird maybe. But the thing is that it should skip the first value :)  Cause you do not care about first value, you care about changes to the value. To grab current value you use `value` :)

packages/cerebral-provider-firebase/src/createOnValue.js
christianalfoni: Indeed, silly me... assign much better, fixed in next commit :)

christianalfoni:
Allrighty... I have adjusted to comments and fixed a couple of snags. Also created an extra test for warning about wrong usage of "strict".

SO! We had a long discussion on this change. It is very hard for me to identify the total scope of what we want to do, so please excuse me if I miss something :)

Computed has definitely served a purpose in Cerebral. It basically allowed us to do two things:

1. Define more complex state dependencies
```js
Computed({
  items: 'items',
  currentItem: 'currentItemKey'
}, ({items, currentItem}) => {
  return items[currentItem]
})
```

2. Compute values
```js
Computed({
  items: 'items',
  filter: 'filter'
}, ({items, currentItem}) => {
  return items.filter((item) => item.someFilter === filter)
})
```

Now, the first scenario is already been solved by TAGS, cleaning up quite a bit of computeds I would believe:

```js
connect({
  currentItem: state`items.${state`currentItemKey`}`
}, ...)
```

The second example can not be handled by Tags though, which means we still need some concept of it. That said, there is still a third scenario that computed has not been able to solve:

```js
Computed((props) => ({
  user: `users.${props.userKey}`,
  projects: 'projects'
}), ({user, projects}) => {
  return user.projectIds.map((projectId) => projects[projectId])
}
```

Now, the problem here is that we depend on `projects`, meaning that any change to any project would make this computed run again. With the new change we can:

```js
connect({
  projects(getState, getProp) {
    const user = getState(`users.${getProp('userKey')}`)

    return user.projectIds.map((projectId) => getState(`projects.${projectId}`) 
  }
})
```

We depend on the exact projects of the user. This is a super important optimization that affects any kind of relational data. 

BUT! This API does not solve everything. The ability to compose these functions for example has not been a focus. It might also seem strange that we have to call `getProp` and `getState` as functions... it just has to be that way, because that is the only way to track their usage :) But of course it does not "look as smooth".

So what I want to conclude with here is that this is a good replacement for Computed. These functions are more powerful because you create specific dependencies, and that is the most important thing to solve I think. When it comes to creating values the API is not as "nice" as Computed, but with the new Tags the need for computed is reduced.

This is a low level implementation I am sure people will find good abstractions over it as well. For example with tags:

```js
connect({
  user: User(props`userId`)
}, ...)

function User (idTag) {
  return (state, props) => {
    // By passing the state and props arg into the tag "getValue"
    // it is able to resolve the tag, whatever it is composed by.
    // we name them "state" and "props" to match with tag type
    const id = idTag.getValue({state, props})

    return getState(`users.${id}`)
  }
}
```

So yeah... I vote for bringing this in as a low level powerful dependency identifier, as a replacement for Computed. As you can see, we can build new powerful stuff over this :)

BUT PLEASE! If I am missing something important, please shout out and big thanks for bringing so much experience and suggestions on how to solve things :)

edgesoft:
@christianalfoni Looks good to me 👍

fopsdev:
Like it so far :)
Wondering how to solve "dependant computeds"
In my current app i'm using 2 computeds. lets call them _filter_ and _sort_
_filter_ just returns a filtered list originating from the connected list in state
_sorting_ consumes _filter_ and does sorting and paging and also returns a list which gets consumed by a simple ui displaying this information
so now why did i split this into 2 computeds? well its because the _filter_ result also gets used by another ui component (its a generic paging control in my case)
now since (if i understood this comment from @christianalfoni  correctly) we cannot create dependent computeds anymore, whats the recommended way of doing this? As a beginner of JS i now would be tempted to save the results in some global namespaced variables? would appreciate some advice here :)

Guria:
@fopsdev Good point. Nested computeds is also a proble I want to be solved out of the box. But I also would like to warn you not to stre any kind of derived data. So writing back results into state from computeds is not solution at all.

fopsdev:
I'm also thinking on how to solve that.
One idea is to put those two computeds into one and connect to all the state dependencies of both  so it gets recalculated for sure. But then it would be really great to know what exactly changed so we could do some more fine grained stuff inside the computed and that brings us back to something like: to:https://github.com/cerebral/cerebral/issues/608

Guria:
@fopsdev 
```js
function add (a, b) {
  return a + b
}

function multiply (a, b) {
  return a * b
}

function computedAdd (getState, getProps) {
  return add(getState('foo'), getState('bar'))
}

function computedMultiplyResultOfAdd (getState, getProps) {
  return multiply(computedAdd(getState, getProps), getState('baz'))
}
```
Idea here is just to split state extraction and actual computation.
Requires a bit of manual managing, but still possible.
And of course we can build general solution for such compositions on top of it,

fopsdev:
@Guria
Nice! :) So for caching something like:
```js
function computedAdd (getState, getProps) {
  const result = add(getState('foo'), getState('bar'))  	
  return () =>  result
}

function computedMultiplyResultOfAdd (getState, getProps) {
	const baz = getState('baz')
	const add = computedAdd(getState, getProps)
	return () => multiply(add(), baz )
}
```
could maybe work... not sure ... hmmm how would you solve caching with your example @Guria ?

christianalfoni:
Hi guys,

Okay, I think I have found a route of compatability here. One thing I realised creating the `User` example above is that we should change signature to:

```js
function compute (get) {}
function compute ({state, props}) {}
```

Let me give some examples of what we are able to do with this:

### Factories in connect
```js
function User (idTag) {
  return (get) => {
    const userId = idTag.getValue(get)

    return get.state(`users.${userId}`)
  }
}

connect({
  user: User(props`userId`)
})
```

### Can use same signature in signals
```js
function ElapsedTime (dateATag, dateBTag) {
  return (get) => {
    const dateA = dateATag.getValue(get)
    const dateB = dateBTag.getValue(get)

    return dateA - dateB
  }
}

[
  set(state`elapsedTime`, ElapsedTime(state`someDate`, state`someOtherDate`))
]
```

### Composing
Now... I think a bit of the confusion around composing is that we are not handling values, we are handling state, props and input in general. It is just more complicated to compose that way. If we only had values going in and values going out... no problem :) I think @Guria shows exactly how composition can be done with this approach, where the refactored example would be:

```js
function add (a, b) {
  return a + b
}

function multiply (a, b) {
  return a * b
}

function computedAdd ({state}) {
  return add(state('foo'), state('bar'))
}

function computedMultiplyResultOfAdd (get) {
  return multiply(computedAdd(get), get.state('baz'))
}
```

So yeah... I suggest we make the signature more compatible with signals and tags. It is done in a giffy ;-)

henri-hulski:
@christianalfoni Cool! Yeah for me this signature looks much better and more consistent.

Guria:
Finally I feel this solution to be pretty solid.
Anyway I still sugget to expose general helper here:

```js
connect({
  fullName: Computed(state`user.name`, state`user.surname`, (name, surname) => [name, surname].join(' '))
})
```

christianalfoni:
@Guria Yeah, that makes more sense to me now as well. Cause now it can also be used in:

```js
[
  set(state`foo.bar`, Computed(state`firstName`, state`lastName`, (firstName, lastName) => firstName + lastName))
]
```

FWeinb:
@Guria This is similar to what I am suggesting in #611 but making the call to `Computed` optional.

christianalfoni:
yeah, @FWeinb I think ideas are merging here. Only reason we need `compute` around it is to make it compatible with signal usage as well :)

Guria:
@FWeinb Yes, I know. But I like explicit Computed call more:
* it's explicit
* it's a general factory concept
* signature looks same as reselect

FWeinb:
@christianalfoni @Guria 
Great! I would go for allowing `compute([])` and `compute(...args)`. So a module could export an array that has the state dependencies already.

christianalfoni:
@FWeinb Though exporting an array you could just spread them? :) `compute(...exportedarray)`?

Guria:
@christianalfoni and don't foget to support nested computeds within helper:
```js
Computed(Computed(state`foo`, state`bar`, (foo, bar) => foo + bar), state`baz`, (nested, baz) => nested * baz)
```

henri-hulski:
+1 for using compute instead of computed.

henri-hulski:
The helper looks also really great now  
So for me this is the way to go.

christianalfoni:
@Guria Cool!

gaspard:
With idTag.getValue(get) aren't we losing support for pure function helpers ? Why not use resolveArg ?

christianalfoni:
@gaspard You can still do that in an action, but there is no "context" in components, so there is no `resolveArg`... looking at making this more consistent now

christianalfoni:
Okay, so to summarize all layers of computed usage:

```js
// CONNECT
function User(userId) {
  return compute(userId, (userId, get) => {
    return get.state(`users.${userId}`)
  })
}

connect({
  user: User(props`userId`)
})

// SIGNAL
function IsAwesome(module) {
  return compute(module, (module, get) => {
    return get.state(`modules.${module}.isAwesome`)
  })
}

[
  set(state`isAwesome`, IsAwesome(input`module`))
]


// ACTION
function IsAwesome(module) {
  return compute(module, (module, get) => {
    return get.state(`modules.${module}.isAwesome`)
  })
}

function setAwesome ({input, state, resolveArg}) {
  state.set('isAwesome', resolveArg.value(IsAwesome(input.module)))
}
```

When we see this working in all layers I am considering some last changes.

1. We do not treat function as a generic thing. You always "compute". This returns a class that can be identified and run appropriately

2. We rename `resolveArg` to `resolve`, as we now do not necessarily resolve an argument

So typical creation of a computed now would be:

```js
import {compute} from 'cerebral'

function UserWithProjects (userId) {
  return compute(userId, (userId, get) => {
    const user = get.state(`users.${userId}`)

    return {
      ...user,
      projects: user.projectIds.map((id) => get.state(`projects.${id}`))
    }
  })
}
```

This can be used as:

```js
// Connect
connect({
  user: UserWithProject(props`userId`)
})

// Signal
[
  // Would typically not do this, but just as example
  set(state`path`, UserWithProjects(input`userId`))
]

// Action
function someAction ({input, resolve}) {
  const userWithProject = resolve.value(UserWithProject('some-id-123'))
  const otherUserWithProject = resolve.value(UserWithProject(input.userId))
}
```

So `compute` works like this:

```js
// A function indicate a computing step, it will always receive the `get` argument.
// The usage of state/props is tracked in connect
compute((get) => {
  get.state('some.path')
  get.props('some.other.path')
})

// Any tags or other values used will be resolved and passed to the compute step
compute(state`foo`, input`bar`, (foo, bar, get) => {
  foo // "bar"
  bar // "baz"
})

// You can pass an existing computed as well
compute(someOtherComputed, (otherComputed, get) => {
  otherComputed // "computed value"
})

// You can compose multiple
connect({
  list: compute(filteredList, onlyAwesome)
}
```

Now there is consistency in how compute is used across all layers. We have composability and connect is able to track however complex the computeds might be :)

docs/docs/content/install/02_debugger.en.md
```diff
@@ -48,9 +48,11 @@ const controller = Controller({
     verifyStrictRender: true,
     // Throw error when overwriting existing input property
     preventInputPropReplacement: false,
+    // Turn off the warning of not pointing to children of objects/arrays when declearing state dependencies
+    preventDependencyWarning: false,
     // Shows a warning when you have components with number of state dependencies
-    // or signals above the set number
-    bigComponentsWarning: {state: 5, signals: 5},
+    // or signals above the set number  
+    bigComponentsWarning: 5,
```
Guria: is it stil defaults to 5? I thought we decided to increase it.

packages/cerebral/src/index.js
```diff
@@ -1 +1,6 @@
+import Compute from './Compute'
+
 export {default as Controller} from './Controller'
+export function compute (...args) {
```
Guria: Why factory was placed here? Why not to place it in Compute.js?

docs/docs/content/in-depth/12_computed.en.md
```diff
@@ -4,18 +4,19 @@ title: Computed
...
-export default Computed({
-  limit: state`items.limit`,
-  items: state`items.list`
-}, ({limit, items}) => {
-  return items.filter((item, index) => index < limit)
-})
+export default compute(
+  state`items.limit`,
+  state`item.list`,
+  (limit, list) => {
```
fmal: the param name should be `items` instead of `list` to make the example 100% correct 😃

packages/cerebral/src/Compute.js
```diff
@@ -0,0 +1,30 @@
...
+    this.value = null
+  }
+  getValue(get) {
+    const result = this.args.reduce((details, arg, index) => {
+      if (arg instanceof Compute || arg instanceof Tag) {
+        details.results.push(arg.getValue(get))
+
+        return details
+      } else if (typeof arg === 'function') {
+        details.results.push(arg(...details.results.slice(details.previousFuncIndex, index), get))
```
fmal: shouldn't `previousFuncIndex` be altered here `details.previousFuncIndex = index` ?

demos/todomvc/src/components/App/index.js
```diff
@@ -8,7 +8,7 @@ import Recorder from '../Recorder'
 import visibleTodosRefs from '../../computed/visibleTodosRefs'
 
 export default connect({
-  todos: state`app.todos`,
+  todos: state`app.todos.**`,
```
henri-hulski: Would not `todos: state`app.todos.*`,` be enough as we use here only `Object.keys(props.todos).length`?

henri-hulski:
@christianalfoni Looks great! :ok_hand:

fmal:
@christianalfoni so as far i understand the state dependencies passed to `compute` like this:

```js
compute(
  state`app.todos.**`,
  state`app.filter`,
  (todos, filter) => {}
)
```

will not actually be tracked? In order to optimize the former compute one has to declare it like this:

```js
compute(({ state }) => {
 const todos = state('app.todos.**')
 const filter = state('app.filter')
})
```
then it will run only when either `todos` or `filter` changed?

Analogically, would 

```js
function User (passedUserKey) {
  return compute(
    passedUserKey,
    (userKey, {state}) => {
      return state(`users.${userKey}.**`)
    }
  )
}

connect({ user: User(props`userKey`) })
```
be less efficient/optimized than:

```js
function User (passedUserKey) {
  return compute(({ state, props }) => {
    return state(`users.${passedUserKey}.**`)
  })
}
```

aside from the fact that the latter isn't capable of resolving `passedUserKey` if it's a tag?

fmal:
Also, what would be the recommended approach for testing the "new" computeds?

```js
export const limitList = (limit, list) => {
  return items.filter((item, index) => index < limit)    
}

export default compute(
  state`items.limit`,
  state`item.list`,
  limitList
);
```

and then importing `limitList` in the test and simply passing fake list? 

Another thing is that when you use `state` or `props` getter inside computed's function you need to mock the getters. With the previous implementation of computeds there was no such requirement, but i guess that's the necessary trade of 😃

saulshanabrook:
Has using the [Proxy object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) been discussed to implement this instead of relaying on the `getState` and `getProp` functions? It allows you to track property access dynamically. It is [supported in all main browsers](http://caniuse.com/#feat=proxy) and has a [Babel polyfill as well](https://www.npmjs.com/package/babel-plugin-proxy). 

I believe this would get around some of the verbosity that [@christianalfoni explained](https://github.com/cerebral/cerebral/pull/607#issuecomment-273920457):

> BUT! This API does not solve everything. The ability to compose these functions for example has not been a focus. It might also seem strange that we have to call getProp and getState as functions... it just has to be that way, because that is the only way to track their usage :) But of course it does not "look as smooth".

FWeinb:
@saulshanabrook 
The Proxy spec can't be polyfilled in a very performant way. The polyfill you are linking to is wrapping every property access in a function call which in itself is bad for performance. I am the maintainer of [cerebral-proxy-tags](https://github.com/FWeinb/cerebral-proxy-tags). This module is implements `Tags` using proxy object. While developing it I evaluated the use of the polyfill and came to the conclusion that the performance overhead is not worth it (keep in mind that you would use that polyfill in IE which is slow).

But this is a great suggestion for the future. I would suggest that IE is still a major browser and not supporting it in cerebral would be bad in my opinion.

docs/docs/content/api/04_tags.en.md
```diff
@@ -3,99 +3,18 @@ title: Tags
...
-```js
-import {string, state} from 'cerebral/tags'
-import showMessage from '../factories/showMessage'
-
-export default [
-  showMessage(string`Hi there ${state`app.user.name`}!`)
-]
-```
+- **state** : Used in signals and connect to target state
+- **input** : Used in signals to target state
```
christianalfoni: Fixed in new website branch

docs/docs/content/install/02_debugger.en.md
christianalfoni: Ops, fixed

packages/cerebral/src/index.js
christianalfoni: Good point! fixed

docs/docs/content/in-depth/12_computed.en.md
christianalfoni: Fixed in newDocs branch :)

packages/cerebral/src/Compute.js
christianalfoni: Yes indeed! Strange, I wrote a test for this specifically... will review and fix! Nicely spotted!

christianalfoni: Okay, fixed, needed one more func in my test ;-)

demos/todomvc/src/components/App/index.js
christianalfoni: Yes indeed @henri-hulski , nice catch :)

christianalfoni:
@fmal 

It will actually track those tags as well :) So no worries there... the reason is that the tags actually uses the same {state, props} that is passed into the function :)

And yes, you would have to test like that... basically same as `connect` :)

christianalfoni:
@FWeinb @saulshanabrook Really fun to see discussions on proxy continuing and I hope IE is left in the ditch sooner than later, so that we could explore a "proxy only" approach.

I am not quite sure I agree that signature is more consistent. The thing is that Tags are declarative, if used with:

```js
const bar = get.state`foo.bar`
```

they suddenly become imperative. So imperative when working inside functions, like actions as well, but declarative in the rest... signals, connect etc. :)

It would though be amazing to have a proxy so for example:

```js
import {state, input, props} from 'cerebral/proxies'

// signals
[
  set(state.foo, input.foo)
]

// connect
connect({
  foo: state.items[props.itemKey]
})

// Computed
compute(state.currentUserId, (userId, get) => {
  const user = get(state.users[userId])

  return user
})
``` 

Like THAT is freakin` awesome. So to have a compatible signature it would have to be:

```js
// Computed
compute(state`currentUserId`, (userId, get) => {
  const user = get(state`users${userId}`)

  return user
})
```

We could do that... but it kinda goes against actions. Where you go:

```js
function myAction ({state}) {
  const user = state.get('users.foo')
}
```

This consistency is very hard to balance. Currently it is "inside actions imperative, elsewhere declarative" :) Not saying it is right, it is just consistent.

christianalfoni:
Anyone want to investigate creating a Babel loader for this? :)

Do not even have to use proxies... could just convert

```js
import {state} from 'cerebral/proxies'

[
  set(state.foo, 'bar')
]

// Converts to
import {state} from 'cerebral/proxies'

[
  set(state('foo'), 'bar')
]
```

Like... no need for proxies even

FWeinb:
@christianalfoni
With my proxy-tags repo. The first two examples signals and computed are working like that. 
I can second that using these tags in an imperative manner could be confusing.

christianalfoni:
Okay... so one final thing here:

We can change:

```js
compute(state`currentUserId`, (userId, get) => {
  const user = get.state(`users.${userId}`)

  return user
})
```

```js
compute(state`currentUserId`, (userId, get) => {
  const user = get(state`users${userId}`)

  return user
})
```

So **compute** is all about tags... I think it keeps consistency since compute is all about tags in the first place.

`state` in action context is its own thing... far away from compute :)

christianalfoni:
Okay, so the signature is changed and I also added automatic tracking of objects/arrays adding `**` to path. You can override with `*`, if you rather want immediate child interest. Now there is no asterix introduction at all, only this `*` optimization. Lower threshold and cleaner API :)

christianalfoni:
Okay, updated demos, though Firebase Demo needs some more work... do not know the project well enough. It is running, but some things not working. Hopefully @gaspard or @saitodisse can have a look at it when it is brought in :)

demos/demo/src/common/Collection/computed/visibleKeys.js
```diff
@@ -1,26 +1,26 @@
...
-      items: state`${collectionPath}.**`,
-      afilter: state`${filterPath}`,
-      selectedKey: state`${draftPath}.key`
-    },
-    ({items, afilter, selectedKey}) => {
+  return compute(
+    state`${collectionPath}`,
+    state`${filterPath}`,
+    state`${draftPath}.key`,
+    (items, afilter, selectedKey) => {
```
Guria: what stands for afilter?

demos/demo/src/components/Email/index.js
```diff
@@ -1,16 +1,12 @@
...
-      <a className='level-item' target='_blank'
-        href={email && `mailto:${email}`}>
-        <span className='icon is-small'>
-          <i className='fa fa-envelope' style={email ? null : {color: '#ccc'}} />
-        </span>
-      </a>
-    )
-  }
-)
+export default function Phone ({email}) {
```
Guria: function Email ?

demos/todomvc/src/computed/counts.js
```diff
@@ -1,21 +1,22 @@
...
+import {compute} from 'cerebral'
 import {state} from 'cerebral/tags'
 
-export default Computed({
-  todos: state`app.todos.**`
-}, props => {
-  return Object.keys(props.todos).reduce((counts, ref) => {
-    let todo = props.todos[ref]
+export default compute(
+  state`app.todos.**`,
```
Guria: ** ?

demos/todomvc/src/computed/isAllChecked.js
```diff
@@ -1,12 +1,13 @@
...
-  visibleTodosRefs: visibleTodosRefs,
-  todos: state`app.todos.**`
-}, props => {
-  return props.visibleTodosRefs.filter((ref) => {
-    return !props.todos[ref].completed
-  }).length === 0 && props.visibleTodosRefs.length !== 0
-})
+export default compute(
+  computeVisibleTodosRefs,
+  state`app.todos.**`,
```
Guria: ** ?

demos/todomvc/src/computed/visibleTodosRefs.js
```diff
@@ -1,17 +1,18 @@
...
 import {state} from 'cerebral/tags'
 
-export default Computed({
-  todos: state`app.todos.**`,
-  filter: state`app.filter`
-}, props => {
-  return Object.keys(props.todos).filter((ref) => {
-    const todo = props.todos[ref]
+export default compute(
+  state`app.todos.**`,
```
Guria: ** ?

christianalfoni:
@Guria Would suggest we do a Firebase Demo fix in a new PR... it is not running correctly either, but I do not have much insight into it and I do not think it runs in master now either. So should not block PR :)

docs/docs/content/install/02_debugger.en.md
```diff
@@ -48,9 +48,11 @@ const controller = Controller({
     verifyStrictRender: true,
     // Throw error when overwriting existing input property
     preventInputPropReplacement: false,
+    // Turn off the warning of not pointing to children of objects/arrays when declearing state dependencies
+    preventDependencyWarning: false,
     // Shows a warning when you have components with number of state dependencies
-    // or signals above the set number
-    bigComponentsWarning: {state: 5, signals: 5},
+    // or signals above the set number
```
Guria: extra spaces

docs/docs/content/install/02_debugger.en.md
Guria: hmm, isee still 5

packages/cerebral/src/viewFactories/connect.js
```diff
@@ -1,17 +1,13 @@
-function connect (HOC, paths, passedSignals, injectedProps, passedComponent) {
+function connect (HOC, dependencies, injectedProps, passedComponent) {
```
Guria: injectedProps -> mergeProps ?

packages/cerebral/src/viewFactories/Hoc.js
```diff
@@ -1,198 +1,283 @@
...
       }
-      createTagGetters () {
-        return {
-          state: this.context.cerebral.controller.getState.bind(this.context.cerebral.controller),
-          props: this.props,
-          signal: this.context.cerebral.controller.getSignal.bind(this.context.cerebral.controller)
-        }
+
+      if (this.mergeProps) {
+        return this.mergeProps(dependenciesProps, props)
```
Guria: how about getters we discussed?

packages/cerebral/src/utils.js
```diff
@@ -142,4 +147,70 @@ export function getProviders (module) {
...
+
+    let matches = []
+    if (changes[key] === true) {
+      if (dependencyMap[pathKey]) {
+        matches = matches.concat(dependencyMap[pathKey])
+      }
+      if (dependencyMap[pathKey + '.*']) {
+        matches = matches.concat(dependencyMap[pathKey + '.*'])
+      }
+      if (dependencyMap[pathKey + '.**']) {
```
Guria: is .** still used or it just used as sinonym?

packages/cerebral/src/viewFactories/Hoc.js
```diff
@@ -1,198 +1,283 @@
...
-          this.forceUpdate()
-        }
+
+      this.forceUpdate()
+    }
+  }
+
+  return function HOC (dependencies, mergeProps, Component) {
+    if (typeof dependencies === 'function') {
+      throw new Error('You can not use a function to define dependencies. Use tags or a function on the specific property you want to dynamically create')
```
Guria: throwError

packages/cerebral/src/viewFactories/Hoc.js
```diff
@@ -1,198 +1,283 @@
...
+    }
+
+    if (!dependencies) {
+      throw new Error('There is no reason to connect a component that has no dependencies')
```
Guria: throwError

docs/docs/content/in-depth/12_computed.en.md
```diff
@@ -62,3 +67,105 @@ export default connect({
...
+```
+
+You would define **User** as:
+
+```js
+function User (passedUserKey) {
+  return compute(
+    passedUserKey,
+    (userKey, {state}) => {
+      const user = state(`users.${userKey}.**`)
```
edgesoft: Should ** go away?

docs/docs/content/in-depth/12_computed.en.md
```diff
@@ -62,3 +67,105 @@ export default connect({
...
+Now your component depends on the user and any changes to its children. It also depends on each specific project of the user and any changes to their children.
+
+But you could split this up. You could keep the **User** as:
+
+```js
+function User (passedUserKey) {
+  return compute(
+    passedUserKey,
+    (userKey, {state}) => {
+      return state(`users.${userKey}.**`)
```
edgesoft: Should ** go away?

docs/docs/content/in-depth/12_computed.en.md
```diff
@@ -62,3 +67,105 @@ export default connect({
...
+}
+```
+
+And rather add a new compute to grab the projects:
+
+```js
+compute((user, {state}) => {
+  return {
+    ...user,
+    projects: user.projectIds.map((id) => state(`projects.${id}.**`))
```
edgesoft: Should ** go away?

packages/cerebral/src/viewFactories/Hoc.js
```diff
@@ -1,198 +1,283 @@
...
+      this.dependencyTrackersDependencyMaps = this.getDependencyTrackersDependencyMaps(props)
+      this.tagsDependencyMap = this.getTagsDependencyMap(props)
+    }
+    /*
+      Register the component to the dependency store with its
+      state tracker and tags state dependencies
+    */
+    componentWillMount () {
+      if (!this.context.cerebral.controller) {
+        throw new Error('Can not find Cerebral controller, did you remember to use the Container component? Read more at: https://cerebral.github.io/api/05_connect.html')
```
Guria: throwError

docs/docs/content/tutorial/09_compute_state.en.md
```diff
@@ -19,31 +19,28 @@ function setStarsSum({state}) {
...
-In Cerebral we can automatically compute state using a **Computed**.
+In Cerebral we can automatically compute state by using **compute**. It is basically a function that takes any number of arguments to produce a value. Let us look at how it works with our scenario:
 
 ```js
-import {Computed} from 'cerebral'
-import {state} from 'cerebral/tags'
+import {compute} from 'cerebral'
+
+export default compute(function starsCount (get) {
+  const repos = get.state('repos.**')
```
edgesoft: Should ** go away or is this valid here?

packages/cerebral/src/viewFactories/Hoc.js
```diff
@@ -1,198 +1,283 @@
...
+        this.tagsDependencyMap
+      ))
+    }
+    /*
+      A getter for StateTracker and tags to grab state from Cerebral
+    */
+    stateGetter (path) {
+      const value = this.context.cerebral.controller.getState(path)
+
+      return value
```
Guria: return this.context.cerebral.controller.getState(path)

packages/cerebral/src/utils.js
edgesoft: Should we add a deprecation warning

demos/demo/src/common/Collection/computed/visibleKeys.js
christianalfoni: Let us fix Firebare Demo in separate branch... to not block this one, other maintainers on it and I do not believe current master Firebase Demo runs correctly anyways

demos/demo/src/components/Email/index.js
christianalfoni: Let us fix Firebare Demo in separate branch... to not block this one, other maintainers on it and I do not believe current master Firebase Demo runs correctly anyways

demos/todomvc/src/computed/counts.js
christianalfoni: Ops!fixed

demos/todomvc/src/computed/isAllChecked.js
christianalfoni: Ops, fixed

demos/todomvc/src/computed/visibleTodosRefs.js
christianalfoni: ops, fixed

docs/docs/content/install/02_debugger.en.md
christianalfoni: Updated in `newDocs` branch

packages/cerebral/src/viewFactories/connect.js
christianalfoni: Ah, yes, lets be consistent and call it mergeProps. Fixed!

packages/cerebral/src/utils.js
christianalfoni: It is used internally, also shown in debugger. Just not exposed as a required API of users

docs/docs/content/in-depth/12_computed.en.md
christianalfoni: Yes, though docs are updated in `newDocs` branch

docs/docs/content/in-depth/12_computed.en.md
christianalfoni: Yes, though docs are updated in `newDocs` branch

docs/docs/content/in-depth/12_computed.en.md
christianalfoni: Yes, though docs are updated in `newDocs` branch

docs/docs/content/tutorial/09_compute_state.en.md
christianalfoni: Yes, though docs are updated in `newDocs` branch

packages/cerebral/src/utils.js
christianalfoni: It is still used internally :)

docs/docs/content/in-depth/12_computed.en.md
christianalfoni: Yes, though docs are updated in `newDocs` branch

docs/docs/content/in-depth/12_computed.en.md
christianalfoni: Yes, though docs are updated in `newDocs` branch

docs/docs/content/in-depth/12_computed.en.md
christianalfoni: Yes, though docs are updated in `newDocs` branch

docs/docs/content/tutorial/09_compute_state.en.md
christianalfoni: Yes, though docs are updated in `newDocs` branch

packages/cerebral/src/utils.js
christianalfoni: Needed internally and does not affect load... just fine to use it. Think we can handle this simply by communicating it :)

Guria:
https://github.com/cerebral/cerebral/blob/ec9fc14dffcbe9fe53f343c8840cf4e110390798/packages/cerebral/src/devtools/index.js#L31

christianalfoni:
@Guria Fixed that one now... just testing your mergeProps get and then will push in changes :)

packages/cerebral/src/viewFactories/Hoc.js
christianalfoni: fixed in new commit

packages/cerebral/src/viewFactories/Hoc.js
christianalfoni: fixed in new commit

packages/cerebral/src/viewFactories/Hoc.js
christianalfoni: fixed in new commit

packages/cerebral/src/viewFactories/Hoc.js
christianalfoni: Fixed in new commit

christianalfoni:
Okay, NOW I believe we are up to date also with all latest comments. Thanks so much for reviewing :D

Guria:
are you sure that all docs changes migrated to newDocs?
could we clean up this pr from doc edits to avoid merge conflicts?

christianalfoni:
@Guria That sounds like a good idea! Yes, everything is moved over. So removing doc changes here is safe... is that something to sprinkle Aleksey GIT magic on, or can I do something? :)

Guria:
ok I will do it :)

PR #616, opened 2017-01-23
feat(Operators): add tags toString to show displayname of operators

christianalfoni:


packages/cerebral/src/operators/merge.js
```diff
@@ -17,7 +17,9 @@ export default function (target, ...values) {
     }))
   }
 
-  merge.displayName = 'operator.merge'
+  merge.displayName = `operator.merge(${String(target)}, ${values.map((value) => {
+    return String(value)
```
edgesoft: Humm maybe we merge a payload that is rather large from the backend.

packages/cerebral/src/operators/set.js
```diff
@@ -19,7 +19,7 @@ export default function (target, value) {
     }
   }
 
-  set.displayName = 'operator.set'
+  set.displayName = `operator.set(${String(target)}, ${String(value)})`
```
edgesoft: Could be a large payload.

packages/cerebral/src/operators/merge.js
christianalfoni: It is not the value, it is just the string representation of the tag... so basically displays 

```js
"state`foo.bar`" 
```

not the value :)

packages/cerebral/src/operators/set.js
christianalfoni: Same as above ;-)

edgesoft: Ok, cool my mistake

packages/cerebral/src/operators/merge.js
edgesoft: Ok, cool my mistake

Guria:
/home/travis/build/cerebral/cerebral/packages/cerebral/src/tags/index.test.js:77:34: Unexpected template string expression.

christianalfoni:
Fixing it now :)

PR #619, opened 2017-01-24
fix(operators): fix display name format for debugger output

garth:


christianalfoni:
Maaan, how could I avoid catching that? :)

PR #621, opened 2017-01-24
refactor(Controller): extend controller from function tree

christianalfoni:
Tests needed to be refactored due to how controller async throws errors to not block function tree execution and messaging debugger


Guria:
How about controller.runTree = wrapped controller with deprecation?

PR #632, opened 2017-01-28
WIP: .* optimization failed within nested computed

Guria:


demos/todomvc/src/components/App/index.js
```diff
@@ -6,6 +6,11 @@ import TodosList from '../List'
 import TodosFooter from '../Footer'
 import Recorder from '../Recorder'
 import visibleTodosRefs from '../../computed/visibleTodosRefs'
+import isAllChecked from '../../computed/isAllChecked'
+
+const AllChecked = connect({
```
Guria: component that depends only on nested computed with '.*' optimization and get() calls to track particular tasks `completed` fields

Guria:
solved in #633

PR #633, opened 2017-01-28
fix(Views): update compute depmaps on state change

christianalfoni:


packages/cerebral/src/viewFactories/react.test.js
```diff
@@ -621,6 +621,41 @@ describe('React', () => {
         controller.getSignal('changeState')()
         assert.equal(TestUtils.findRenderedDOMComponentWithTag(tree, 'div').innerHTML, 'bar2')
       })
+      it('should update depdency map when compute is rerun', () => {
```
Guria: depdency -> dependency

PR #655, opened 2017-02-02
Add test helpers

garth:
- [x] add runCompute
- [x] add runAction
- [x] add runSignal
- [ ] ~~add renderComponent~~ This is view layer specific
- [x] prepare docs

## runCompute

The `runCompute` test helper accepts a `compute` and `state` arguments and returns the compute output.

```js
var result = runCompute(compute, state)
```

### Example

```js
import {compute} from 'cerebral'
import {props, state} from 'cerebral/tags'
import {runCompute} from 'cerebral/test'

import Multiply from './Multiply'

it('should multiply by the specified number', () => {
  const multiply = Multiply(state`number`, props`number`)
  const result = runCompute(multiply, {
    state: { number: 5 },
    props: { number: 2 }
  })
  assert.equal(result, 10)
})
```

## runAction

The `runAction` test helper accepts a `action` and `fixture` arguments and returns a promise.

```js
runAction(action, fixture).then((result) => {})
```

The optional `fixture` argument should be an object that contains any of the following:

```js
{
  state: {}, // test state
  props: {}, // props passed to the action
  // any other options that can be passed to the
  // cerebral controller, including router, providers...
}
```

The `result` object passed when the promise resolves contains `state`, `controller`, `props` and `output` properties.

```
{
  state,
  controller,
  props: {
    // props data received by action
  },
  output: {
    // action output data
  }
}
```

### Example

```js
import {compute} from 'cerebral'
import {runAction} from 'cerebral/test'

import Increment from './Increment'

it('should increment numbers in state', () => {
  const increment = Increment(state`number`)

  return runAction(increment, { state: { number: 1 } })
    .then(({state}) => assert.equal(state.number, 2))
})
```

## runSignal

The `runSignal` test helper accepts a `signal` (chain of actions) and `fixture` arguments and returns a promise.

```js
runSignal(signal, fixture. options).then((result) => {})
```

The optional `fixture` argument should be an object that contains any of the following:

```js
{
  state: {}, // test state
  props: {}, // props passed to the signal
  // any other options that can be passed to the
  // cerebral controller, including router, providers...
}
```

The optional `options` argument contain the the following options:

`recordActions: true|false|'byName'`

When `recordActions: true` is specified each action will record its props/output against its index within the signal action chain. When `recordActions: 'byName'` is specified each action will record its output against an named property in the result.

The `result` object passed when the promise resolves contains `state`, `controller` and an object for each named action in the signal chain with the same name as the actions with `props` and `output` properties.

```
{
  state,
  controller,
  '2': {
    props: {
      // props data
    },
    output: {
      // action output data
    }
  },
  '1': {
    props: {
      // props data
    },
    output: {
      // action output data
    }
  }
}
```

### Example

```js
import {compute} from 'cerebral'
import {runSignal} from 'cerebral/test'

// the buttonClicked signal has two actions: validateForm and updateIsValid
import buttonClicked from './buttonClick'

it('should handle button clicks', () => {
  const fixture = {
    state: { isValid: false },
    props: { buttonName: 'submit' }
  }
  return runSignal(buttonClicked, fixture, {recordActions: 'byName'})
    .then(({validateForm, updateIsValid, state}) => {
      assert.equal(validateForm.props.buttonName, 'submit')
      assert.equal(updateIsValid.props.isValid, true)
      assert.equal(state.isValid, true)
    })
})
```


maxfrigge:
Looks great!

What if you use an action with the same name twice in a signal - would they overwrite each other? Wouldn't it be enough to validate the state and final output when testing on the signal level?

garth:
Using a name based hash makes asserting action input/output very simple.

If you care about individual action input/output you should probably use unique names.

Returned state at the end is probably the most useful for assertions.

garth:
dependent on https://github.com/cerebral/cerebral/pull/658

maxfrigge:
Guess it will be OK 90% of the time, but could at some point lead to unexpected behaviour when not being aware of it. Maybe it's enough to log a warning.

garth:
@maxfrigge Agreed, I think that it should console.warn when two actions have the same name.

The other option was to build the same action tree, but it's much harder to test. Named actions is just simple and if you don't care then this will never be an issue anyways.

I think that the state output is still the most useful at the signal level, maybe this action input/output feature could even be off by default and you would need to turn in on via a flag?

christianalfoni:
@garth I think this looks great! BUT! actions also has indexes. Might be an idea to expose these indexes instead (as keys on object) and rather have the name as a property

maxfrigge:
@garth disabling by default sounds good. I would assume that, when testing each action anyway, the state after signal as well as "side effects" will be of most value. I haven't yet written much tests for signals though.

garth:
@christianalfoni I considered using the index, but it is error prone. If you add an action then the indexes will change, so you'd need to assert the action name as well maybe.  I just figured that if you an action where you really need to test the input/output, then you will probable be ok with giving it a unique name.

I will turn this feature off by default and only activate via an option, also I will log a warning if there is a name clash.

christianalfoni:
@garth Cool, yeah... makes sense. Why test an operator :)

garth:
@Guria rebased to @christianalfoni props PR, so should be ok to merge after that.

PR #657, opened 2017-02-02
docs(tutorial): tutorial script will not remove unnecessary files

saitodisse:
## Need reviewers for all platforms (Linux, Windows and Mac) 

This fixes webpack watching after tutorial reloads.
On Linux after the script removes the files the page stops to refresh.

## fix

Now will delete files only if needed.

#### steps to reproduce the Issue without this PR

_If your are on linux you can test this strange behavior_:

```sh
# go to tutorial folder
cd docs/tutorial
npm start
```

- Keep switching chapters and suddenly stops to refresh the page, even if we change the files by hand.


christianalfoni:
Okay, I can confirm this works on mac :)

PR #658, opened 2017-02-03
feat(compute): add resolve.compute(compute, props) to signal context

garth:
Enables computes that require props to be called from within actions

### DIFF
--- a/packages/cerebral/src/utils.js
+++ b/packages/cerebral/src/utils.js
@@ -238,6 +238,13 @@ export function createResolver (getters) {
 
       return arg
     },
+    compute (arg, props = {}) {
+      if (arg instanceof Compute) {
+        return arg.getValue(Object.assign({}, getters, { props }))
+      }
+
+      return arg
+    },
     path (arg) {
       if (arg instanceof Tag) {
         return arg.getPath(getters)


garth:
Test coverage will come back up when https://github.com/cerebral/cerebral/pull/655 is merged (it uses this method).

Guria:
I am only curious why have both `resolve.compute` and `resolve.value`?
It really bother me to provide 2 almost identical methods with only slight difference:

--- | compute | value
--- | ---- | ---- 
accept Tags | - | +
accept Computed | + | +
allows to pass props | + | -

christianalfoni:
@Guria Because there is a slight difference :) You can not pass props to tags or plain values, and use of `.value` indicates that you may receive multiple types of values... while `resolve.compute` is explicit. You expect a computed to be passed in and you can give it props.

Imagine a scenario where you would receive a TAG instead a computed on `.value`, passing props. The second argument would be ignored and no error thrown on wrong usage. With `resolve.compute` we throw error if wrong type passed in :)

Guria:
Updated table with note that both compute and value accept plain values.
And i see no reason not to pass props to tags as well:

```js
resolve.value(state`categories.${props`itemKey`}`, { itemKey: input.categoryId })
```

garth:
@Guria good point about tags!

christianalfoni:
@Guria @garth There is something "off" with this thing. `props` is currently conceptually props passed in by component. In the example with TAG above:

```js
resolve.value(state`categories.${props`itemKey`}`, { itemKey: input.categoryId })
```
states that we are doing something wrong I think, because:

```js
resolve.value(state`categories.${input.cateogryId}`)
```
Is the explicit straight forward way of doing it.

I am not saying the suggested API is bad... but then `props` has to conceptually be something else. Not related to components at all. we have to document it as: 

"Tags and computeds supports props. When evaluating them with `resolve` in actions you can pass in these props, but when connected to a component the component props is used as source for props"... something like that.

Guria:
Updated tag support in table :). resolve.compute will just return it unprocessed. Very confusing in my opinion. Let's just extend resolve.value and get rid of compute

christianalfoni:
Yeah, we can do that, but have to fix docs and also how we talk about `props`. The tag `props` is not "the props of component", it is the props passed into the tag/computed... when connected to a component happens to be the components props :)

garth:
I still have the opinion that props are input are conceptually the same thing, just a different context. If they were named the same then this would be a non-issue.

Guria:
@christianalfoni the reason to have
```js
resolve.value(state`categories.${props`itemKey`}`, { itemKey: input.categoryId })
```
instead of 
```js
resolve.value(state`categories.${input.cateogryId}`)
```
is very simple.

I want to have ```state`categories.${props`itemKey`}` ``` exported from my constants and being able to use everywhere. Otherwise, I have to use factory here ``` itemKey => state`categories.${itemKey}` ``` which looks redundant for me.
But you suggest to have 2 dublicating APIs with very slight, strange difference which we will have to document somehow and provide adequte reasoning.

Guria:
> when connected to a component happens to be the components props :)

when connected to a component components props used to pass into computed / tag

henri-hulski:
Maybe we can use here input instead of props.

christianalfoni:
@garth But that would mean in a signal you would have to output whatever you want the next action to have as input to its computed :)

It just so much mental load to track the input to an action that is magically passed into a computed it is resolving. The same thing happened to connect now... we magically pass in props from the component into the computed making it too implicit:

```js
connect({
  user: computedUser // How does it know which user? Magically from the component props
})
// Or in action
function myAction ({resolve}) {
  resolve.value(computedUser) // Where did the userKey come from? Magically from the input
}
```

I would rather actually suggest removing automatic props passing to computed completely and make `props` tag only purpose to extract component props:

```js
connect({
  user: computedUser(props`userKey`) // Now we know where it comes from
})
```

That means if you were to use this computed in an action you would:

```js
resolve.value(computedUser(input.userKey))
resolve.value(computedUser(state.get('app.currentUserKey'))
```

So to summarize:
I state that we have taken this too far. There is too much crashing concepts and implicitness. We should take some steps back and make things more explicit with single purpose:

### 1. The props tag is for extracting component props, it only works in CONNECT

```js
connect({
  user: state`users.${props`userKey`}`,
  computedUser: computedUser(props`userKey`)
})
```

### 2. Computed can only extract state
export default function (userKeyTag) {
  return compute(userKeyTag, (userKey, get) => {
    return get(state`users.${userKey}`)
  })
}

Now it is completely explicit and no confusion what `input` and `props` mean. `input` is only for signals and `props` is only for connect. I see now that we created a bridge for `props` through computed... In hindsight I think this was the wrong choice

garth:
@christianalfoni I'm not saying that you shouldn't be able to add to props. I think that I may have mentioned in the past that I think that compute and action are also the same thing, except that action may have side effects.

If you can use a compute in place of action then you can define additional props for that action when necessary, also you can grab state in the same way as you do for components.

Would make everything more consistent IMO.

Guria:
@christianalfoni 
hmm, but
```js
export default function (userKeyTag) {
  return compute(userKeyTag, (userKey, get) => {
    return get(stateusers.${userKey})
  })
}
```
is just computed factory and props tag would be still resolved by computed itself.
And nothing would stop to define it inline and make it implicit again.
```js
compute(props`itemKey`, (item, get) => {}`)
```
It works like that right now. Am I right?

christianalfoni:
@Guria @garth Yeah... good point Aleksey, I was a bit too hasty with a solution :)

I guess explicitness vs implicitness is up to the developer, as you state here.

But we are definitely doing something wrong here and after some thinking I am warming up to @garth suggestion :) :

Remove concept of `input` and use `props` all over. With this change the debugger should probably change its naming of "input" and "output" to "props" and "output". The computed becomes a concise bridge between signals and connect, the props are either from component or signal. Less to think about

Then the signature of resolve.value would be: `resolve.value(tagComputedValue, overrideProps)`

bfitch:
I think this is a great clarifying change. 👍  Remove all references to `input` and add "just how components receive `props`, in Cerebral, actions receive them as well". It is one concept: passing data to a function.

christianalfoni:
Okay, this is concluded, have prepared a branch with completed refactor, also to related packages. Just fixing docs now :)

PR #662, opened 2017-02-06
fix(website): import input from tags

fvgoto:
`input` should be imported from `cerebral/tags` instead of `cerebral/operators`


Guria:
@fvgoto we just renamed input to props in #660 
please update pr to meet new name

PR #663, opened 2017-02-06
fix(forms): import input from tags

fvgoto:
`input` should be imported from `cerebral/tags` instead of `cerebral/operators`

### DIFF
--- a/packages/cerebral-forms/README.md
+++ b/packages/cerebral-forms/README.md
@@ -200,7 +200,7 @@ export default connect({
 An **action** factory you can use to validate any field in any chain.
 
 ```js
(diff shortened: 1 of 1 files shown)


Guria:
same as in #662 
need to replace input to props, please

fvgoto:
NB: Please check, if it is still wise to talk about "props" in the same file at [packages/cerebral-forms/README.md#L94](https://github.com/fvgoto/cerebral/blob/8c33cc52ac84021047896062594e66d6838cc95f/packages/cerebral-forms/README.md#custom-global-props):
```
#### Custom global props
You can add custom props to the root to the form state.
```
or if it might create confusion with the `props` tag.

PR #664, opened 2017-02-06
fix(firebase): import input,state... from tags

fvgoto:
`input`, `state` and that like should be imported from `cerebral/tags` instead of `cerebral/operators`.
review: double-check for `toString` (I guess, it's: [packages/cerebral/src/tags/Tag.js#L102](https://github.com/cerebral/cerebral/blob/master/packages/cerebral/src/tags/Tag.js#L102))

### DIFF
--- a/packages/cerebral-provider-firebase/README.md
+++ b/packages/cerebral-provider-firebase/README.md
@@ -135,11 +135,11 @@ function someAction({ firebase, path}) {
 
 *factory*
 ```javascript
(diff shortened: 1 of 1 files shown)


fvgoto:
Sorry, my fault: commit message is bad - should read `fix(firebase)` instead of `fix(website)`

Guria:
same as #662 and #663 
need to replace input to props

PR #666, opened 2017-02-07
fix(http) date encoding

Reltik:
This pr fix enables urlEncode to handle parameters where the type is Date. 

Issue: #667


gaspard:
Would be nice to have a test case...

edgesoft:
Maybe we could extend this an have an array of types that could be serialized.

christianalfoni:
Hi there,

I have to challenge this a bit. The reason is just the principle of "doing one thing" and not adding documentation and API just to make things easier, cause it does not make the tool simpler :)

Is there any reason why you could not do this?

```js
http.get('/someurl', {
  date: someDate.toISOString()
})
```

Sorry if I misunderstand :)

Guria:
I also have a feeling that adding some implicit "helpers" are could be harmful.

PR #668, opened 2017-02-07
WIP - feat(Views): improve strict dependency to handle parent replacement

christianalfoni:
Okay, so how does this work? :)

Up until now dependencies has been stored as:

```js
{
  'foo.bar': [ComponentA],
  'foo.baz.**': [ComponentB]
}
```

This was really efficient as the change map:

```js
{
  foo: {
    bar: true
  }
}
```
Could very easily be traversed and look at specific keys in the dependency map. Super fast and straight forward.

The problem occurs when you have dependencies where a parent is completely removed. For example:

```js
connect({
  documentTitle: state`currentDocument.title`
}, ...)
```
If you would do a replacement of currentDocument here:

```js
set(state`currentDocument`, {title: 'wuuut'}), 
```

The above connected component would not update. The reason is that the only way to find this dependency is to iterate all dependencies to see if `currentDocument` is part of them. This is a lot of dependencies to iterate and we loose the optimization.

### The solution
Now dependencies are stored like this:

```js
{
  foo: {
    children: {
      bar: {
        entities: [ComponentA]
      },
      bar: {
        entities: [ComponentB]
      }
    }
  }
}
```

So there is definitely a memory overhead here, but should not be that much as many state deps share paths. Anyways, now it is possible to figure out just as quickly how changes to state affects paths, even in the scenario where we replace a parent. The reason is that now a parent has a `children` property we can further point to.

This is how changes now look:
```js
[{
  path: ['foo', 'bar'],
  forceChildPathUpdates: true
}]
```

It is rather an array of changes. The `forceChildPathUpdates` comes from the model and is set to true where state changes are in risk of replacing paths... like in `set` or `merge`. 

All tests are running green and it would be awesome to test this in an app. I believe I have to make a fix to send correct data format to debugger, but that is quick fix.

### DIFF
--- a/packages/cerebral/src/DependencyStore.js
+++ b/packages/cerebral/src/DependencyStore.js
@@ -86,9 +86,8 @@ class DependencyStore {
             }
           }
         }
-        if (children[childKey].entities) {
-          return traverseChildren(children[childKey].children)
-        }
+
+        return traverseChildren(children[childKey].children)
       }
     }
 
--- a/packages/cerebral/src/DependencyStore.js
+++ b/packages/cerebral/src/DependencyStore.js
@@ -9,7 +9,23 @@ class DependencyStore {
   */
   addEntity (entity, depsMap) {
     for (const depsMapKey in depsMap) {
-      this.map[depsMapKey] = this.map[depsMapKey] ? this.map[depsMapKey].concat(entity) : [entity]
+      const path = depsMapKey.split('.')
+
+      path.reduce((currentMapLevel, key, index) => {
+        if (!currentMapLevel[key]) {
+          currentMapLevel[key] = {}
+        }
+
+        if (index < path.length - 1) {
+          currentMapLevel[key].children = currentMapLevel[key].children || {}
+
+          return currentMapLevel[key].children
+        }
+
+        currentMapLevel[key].entities = currentMapLevel[key].entities ? currentMapLevel[key].entities.concat(entity) : [entity]
+
+        return currentMapLevel
+      }, this.map)
     }
   }
   /*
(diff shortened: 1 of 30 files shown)


garth:
@christianalfoni I can't imagine that memory usage for this change could ever be an issue, but the benefits are very clear. I will be happy to test.

christianalfoni:
Awesome, thanks :) Did a rebase here now because I completely forgot to actually use the new `forceChildPathUpdates` to handle the scenario I went out to fix :)

But all good now!

christianalfoni:
Okay, I did a "run-through" on a simple project and it behaves as expected now. As soon as I get my ass back to work I will test this on our app.

garth:
@christianalfoni Tested this PR agains my code base and it looks very good. Component updates are much more reliable now. 👍

christianalfoni:
Okay, I can confirm our project works as expected as well. Awesome :D

I just have to refactor the debugger to handle the new dependency structure. Will get at it later today. So need to publish new debugger with this as well.

fopsdev:
well, i maybe have a case here...

fopsdev:
ok, just from my first observations:

i have a connect that connects to:
field: state`${props`path`}`   (it's pointing to the cerebral-form field object)
and after this field.value changes it doesn't get updated anymore (was working before)

fopsdev:
image

not sure it's related. but when switching to the component tab in remote debugger i get a call stack size error...

christianalfoni:
@fopsdev Let me look into this... second is missing debugger refactor :) Will get to that when everything works with rendering

packages/cerebral/src/DependencyStore.js
```diff
@@ -42,15 +77,23 @@ class DependencyStore {
...
+    function traverseChildren (children) {
+      for (const childKey in children) {
+        if (children[childKey].entities) {
+          for (let y = 0; y < children[childKey].entities.length; y++) {
+            if (entities.indexOf(children[childKey].entities[y]) === -1) {
+              entities.push(children[childKey].entities[y])
+            }
+          }
+        }
+        if (children[childKey].entities) {
```
fmal: any reason for repeating `if (children[childKey].entities)` condition here? 😃

christianalfoni: Hah, good catch, fixed ;-)

Guria:
@christianalfoni there is failing tests now

fopsdev:
@christianalfoni pulled it in and it works

PR #677, opened 2017-02-13
feat(test): Added a RunSignal factory

garth:
The factory allows runSignal to be called many times without resetting
state in between calls.

#### Example

```js
import {RunSignal} from 'cerebral/test'

it('should accumulate a count', () => {
  const runMathSignal = RunSignal({
    modules: {
      math: math()
    }
  })
  return runMathSignal('math.plusOne').then(({state}) => {
    assert.equal(state.math.count, 1)
    return runMathSignal('math.plusTwo').then(({state}) => {
      assert.equal(state.math.count, 3)
    })
  })
})
```


docs/docs/docs/best_practices/testing.md
```diff
@@ -118,7 +116,9 @@ it('should increment numbers in state', () => {
 
 ## Signals
 
-The `runSignal` test helper accepts a `signal` (chain of actions) and `fixture` arguments and returns a promise.
+### runSignal
+
+The `runSignal` test helper accepts a `signal` (chain of actions or signal name) and `fixture` arguments and returns a promise. `runSignal` is designed to be called one time, to test calling multiple signals in a single test see the `RunSignal` factory below.
```
henri-hulski: I think should be "accepts **the** `signal` and `fixture` arguments".

docs/docs/docs/best_practices/testing.md
```diff
@@ -189,3 +189,32 @@ it('should handle button clicks', () => {
     })
 })
 ```
+
+### RunSignal factory
+
+The `RunSignal` factory is similar to run signal except that it will return a runSignal function that can be called many times without reseting the controller in between.
+
```
henri-hulski: reseting -> resetting

Issue #678, opened 2017-02-14
operators: using whole tree of a tag from root on


fvgoto:
Using the whole tree a tag represents should be possible.
Especially for `props` (former `input`).

Syntax propositions (I think I've seen first one somewhere...):
```
set( state`foo`, props`` )
set( state`foo`, props`.` )
set( state`foo`, props`*` )
```

UseCase:
Store the whole `props` at a state path.
```
// in a component:
  props.mySignal({ width: 3, height: 5 });
// action
  set( state`size`, props`` )
// should result in following state tree
  ...
  size: { width: 3, height: 5 },
```

I guess, it's the tag string parsing code that should be fixed for this...

fvgoto:
Links to #679

hipertracker:
It's not necessary. Too magic.

Guria:
closing as noop for now, but feel free to continue discussion here

PR #679, opened 2017-02-14
feat(tags): allow using whole tree of a tag (#678)

fvgoto:
Also allows for using superfluous '.' path separators (e.g. "foo..bar")

CodeStyle: Just tell me if you want a full-fledged if-else statement instead....

### DIFF
--- a/packages/cerebral/src/tags/Tag.js
+++ b/packages/cerebral/src/tags/Tag.js
@@ -71,7 +71,7 @@ export default class Tag {
         throwError(`A tag is extracting with path "${path}", but it is not valid`)
       }
 
-      return ''==key ? currentValue : currentValue[key];
+      return (key === '') ? currentValue : currentValue[key]
     }, obj)
   }
   /*
--- a/packages/cerebral/src/tags/Tag.js
+++ b/packages/cerebral/src/tags/Tag.js
@@ -71,7 +71,7 @@ export default class Tag {
         throwError(`A tag is extracting with path "${path}", but it is not valid`)
       }
 
-      return currentValue[key]
+      return ''==key ? currentValue : currentValue[key];
     }, obj)
   }
   /*


christianalfoni:
Thanks for the PR @fvgoto :)

I am a little bit unsure of this cause it would encourage bad practice. The reason I say that is:

```js
someSignal({
  propA,
  propB,
  propC
})
```

Should be passed in as:

```js
someSignal({
  somePayload: {
    propA,
    propB,
    propC
  }
})
```

The reason being that you are highly likely to add new things on the props from other actions. The "" would kinda go against this practice, if that makes sense?

Interesting catch on the `..` though. Should look into support that at least :)

But yeah, I am sorry to say I am skeptic to the root changing of props

fvgoto:
@christianalfoni: Oh yeah, great point about the action chain adding on-the-fly stuff to the props - you're right!

For the `..` thing:
- has not to be confound with '..' as parent accessor (as for fs directory structure). This here would accept `foo.....bar` (equals `foo.bar`) as well.
- It would (also) be bad style for sure to use such path strings
- It would imply thinking and auditing about nested string/tag replacements. If the nested yields empty, do we really want to accept the path (where one nesting level will be missing for sure)??

Fazit: I think we'll drop this issue/PR..

---
What about a misc section on the website for "Tips&Tricks" (for FAQ) where outcomes of such discussions could be nailed down in a terse way and helping the occasional/beginning user?
Candidates:
- "Using whole tag tree (e.g. ```props`` ```)? => Wrap your payload into a single property since action chains may add further stuff to `props`."
- "Hook up Computeds into state tree (as were historic Monkeys)? => No. Computeds are kind of filters and can only be used in components or actions."

Guria:
closed for now. feel free to continue discussion here or into related issue

PR #681, opened 2017-02-15
feat(Devtools): Fix time travel

christianalfoni:
- Added mutations tab to Debugger which lists all mutations, referencing the signal
- Double click to time travel
- Reacts to filtering
- Fixed huge memory leak related to resetting mutations in debugger
- Fixed time travel bug in devtools
- Emitting `remember` event with datetime to be used by router to figure out which routed signal to URL trigger (Router can track routed signals in devtools mode)

screen shot 2017-02-15 at 14 27 56


christianalfoni:
Just added a tiny bug fix

PR #684, opened 2017-02-17
fix(firebase): fix firebase signal callbacks (was using old api)

gaspard:


gaspard:
@christianalfoni "createOnChildAdded" no longer needs to be a factory since "controller" argument is not used. I'll clean this up and push the new changes.

gaspard:
I tested with demo and all works fine now (including file upload).

christianalfoni:
@gaspard Does this change require you to use tags inside actions?

```js
function someAction ({firebase}) {
  return firebase.onValue('some.path', signal`woop`)
}
```

?

gaspard:
@christianalfoni Yes.

I understand that we have to either expose two providers, one for operators and the other one for actions or simply use tags in actions.

Discussion continued on discord...

Guria:
@christianalfoni so what resolution on this one?

christianalfoni:
@gaspard Was this one updated after talk on Discord? :)

gaspard:
@christianalfoni Sorry for the delay. Haven't had much time to code. Will do this "operators" export today.

I will also make sure the regular provider does contain any Tag supporting code. I think you are right, if things are to be separated, they have to be so completely.

gaspard:
Will create a new pull request.

PR #693, opened 2017-02-23
RFC: cerebral-server concept

garth:
> This is an early (though working) concept for a "Cerebral Server". The basic idea is to allow you to use the same cerebral concepts on both client and server and make it easy to transfer from client to server or from server to client from within actions.

> I would not expect this PR to be merged as is, but ask for feedback. The readme is below.

# Cerebral Server
Run cerebral signals on both client and nodejs server.

## Server Setup

From your nodejs app you need to create a CerebralServer and then bind it to any existing http server object.

```js
import http from 'http';
import CerebralServer from 'cerebral-server';
import Devtools from 'cerebral/devtools';

const cerebralServer = CerebralServer({
  signals: {
    saveUsersStuff: [saveUsersStuff],
    notifyUser: [notifyUser]
  },
  providers: [
    context => {
      context.mongo = mongo; // 🤦
      return context;
    }
  ],
  devtools: process.env.NODE_ENV === 'production' ? null : Devtools({ remoteDebugger: 'localhost:8585' })
});

const server = http.createServer();
cerebralServer.bind({ server });
server.listen(3000)
```

### Debugging

If you configure devtools for remote debugging in the same way as you do for a cerebral controller (as in the above example) then the signals and actions will be displayed in the regular standalone cerebral debugger.

### Server Actions

Actions for CerebralServer are the same as on the client except that you do not have a access to `state` and you have two additional context methods `getClientSignal` and `returnToClient`. The `returnToClient` method should always be invoked once during a signal called by the client to allow the client action to continue.

```js
// saveUsersStuff
export default ({ props: { clientId, correlationId }, mongo, returnToClient }) => {
  // save client stuff in the db
  if (clientId && correlationId) {
    returnToClient(clientId, correlationId, { allGood: true });
  }
}

// notifyUser
export default ({ props: { clientId, message }, getClientSignal }) => {
  const signal = getClientSignal(clientId, 'notify');
  return signal({ message });
}
```

### Server Events

CerebralServer emits the following events:

```js
cerebralServer.on('connect', ({ clientId }) => {});
cerebralServer.on('disconnect', ({ clientId }) => {});
```

## Client Setup

```js
import { Controller } from 'cerebral';
import CerebralServerClient from 'cerebral-server/client';

  signals: {
    notify: [notify],
    saveMyStuff: [saveMyStuff]
  },
  modules: {
    server: CerebralServerClient()
  }
}

// call the connect signal and provide a unique clientId
controller.getSignal('server.connect')({ clientId: 'someUserName' });
```

### Client Actions

The CerebralServerClient module adds two additional context methods `getServerSignal` and `returnToServer`. The `returnToServer` method should always be invoked once during a signal called by the server to allow the server action to continue.

```js
// notify
export default ({ props: { correlationId, message }, foo, returnToServer }) => {
  // notify the user here
  if (correlationId) {
    returnToServer(correlationId, {});
  }
}

// saveMyStuff
({ props, getServerSignal }) => {
  const signal = getServerSignal('saveUsersStuff');
  return signal(props);
}
```


christianalfoni:
I think this is super interesting @garth :) And this is compatible with existing API of cerebral right? WIll give more feedback later, you asking that question first ;-)

garth:
@christianalfoni this is just a thin wrapper around function-tree. The server is < 100LOC and most of that is transport.

My main goal was to keep everything framework agnostic where possible. It should run standalone or alongside another server framework. Ideally the transport would even be a plugin.

At this moment this is just something that we can use as a discussion point. Maybe it's not something that we would ever want to support?

bfitch:
Wow. This is *very* cool. And kind of melting my brain! I think @maxfrigge mentioned it, but would this allow for Meteor style RPC (remote procedure calls)? Like, the client could invoke the server signal that inserts records in the DB directly, passing form values as `props`? And obviously, it can go the other way, with the server pushing data by invoking client signals. Something like that?

garth:
@bfitch, I'm in no way trying to copy Meteor, but yes that is exactly what it's for.

Context is a little different on the server and there's no state tree, but you can add your db of choice to the context and your all set to use it in the way you described if you so choose.

bfitch:
@garth sweet! I wouldn't want you to copy Meteor :) Just trying to connect it with an existing concept. 

So with this....we can totally invert normal client-server data flow from pull to push. The client state store can be kept in sync by performing the mutation "directly" on the server and having updates pushed to the client (or clients!). Kind of a cool mental shift and good for real time apps.

garth:
I think that we will close this for now, the feedback has been really great. Lots of ideas to work on going forward.

PR #698, opened 2017-02-24
feat(function-tree): Use primitives to identify types of execution

christianalfoni:
After a few iterations now this is the result:

```js
import {sequence, parallel} from 'cerebral' // or function-tree
```

## Sequence
So sequence is normal execution, one after the other:

```js
sequence(
  funcA () {},
  funcB () {}
)
```

Though instead of calling the `sequence` function, you can just use a plain array. That means `[]` is converted to `sequence()` inside function-tree.

```js
[
  funcA () {},
  funcB () {}
]
```

The reason you would want to use `sequence` is to be more explicit and/or name your sequence:

```js
sequence(
  'My sequence',
  funcA () {},
  funcB () {}
)
```

When naming the sequence it is picked up by the debugger.

## Parallel
Well, it is parallel execution. Meaning that "array inside array" is gone, you have to use this function to express parallel execution.

```js
parallel(
  'I can optionally name this as well',
  funcA () {},
  funcB () {}
)
```

When using `parallel` any group of actions (sequence) will be run as "one action". Meaning that:

```js
parallel(
  funcAAsync () {},
  [
      funcBAsync () {},
      funcC () {}
  ]
)
```

`funcC` would run after `funcBAsync` is done, it would not get merged in as a parallel action. This is a lot more predictable behaviour and in the debugger you will see that there is a `parallel` group of actions with a `sequence` group of actions inside it.

Here is a simple screenshot of how debugger looks with the identified grouping:

screen shot 2017-02-25 at 12 52 29

The really great bonus now is that compositions are identified and visualized in debugger. Because you typically import arrays from other chains etc. When not using spread these become their own `sequence` group of actions indicated in the debugger. And if you named them... even better :)


henri-hulski:
Is it now possible to use a single action without array?

packages/cerebral/src/operators/all.js
```diff
@@ -0,0 +1 @@
+export {all as default} from 'function-tree'
```
henri-hulski: We have still an all operator.

packages/cerebral/src/operators/debounce.test.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-env mocha */
 import Controller from '../Controller'
 import assert from 'assert'
-import {debounce} from './'
+import {all, debounce} from './'
```
henri-hulski: The all operator still used.

packages/cerebral/src/operators/wait.test.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-env mocha */
 import Controller from '../Controller'
 import assert from 'assert'
-import {wait} from './'
+import {all, wait} from './'
```
henri-hulski: Here also.

packages/cerebral/src/operators/all.js
christianalfoni: Thanks, fixed :)

packages/cerebral/src/operators/debounce.test.js
christianalfoni: Thanks, fixed

packages/cerebral/src/operators/wait.test.js
christianalfoni: Thanks, fixed

christianalfoni:
Okay, I just need to update function tree debugger to support these changes and it is ready to go :)

Guria:
I was away this days and not followed any discussions. I am positive with this change, though wondering about breaking for current alpha users.

christianalfoni:
@Guria Welcome back :D The only thing that breaks with this change is parallel execution. You have to identify where you do array in array and use `parallel` instead. Due to the nature of the change I could not see how to deprecate this.

If you use `spread` operator, that is still perfectly okay, you will just not get the "sequence" indication in the debugger, as it will think of it as "one sequence".

The static tree also look different now, so need to release new debugger at the same time as this. I suggest we release it on the next hangout to inform people and also make sure we shout out quite a bit about it the days before. So official release is next thursday.

Or is there a better way? :)

PR #701, opened 2017-02-24
RFC: Cerebral Service

garth:
This is a followup from the previous RFC #693. After lots of great feedback it was clear that a full blown server component has to make too many assumptions.

The idea to put cerebral on the server has been very well received though.

Ideally we should just be able to use the regular controller, but without the bits that are browser specific.

So presented here for your feedback is cerebral `Service`. `Service` is the same as the cerebral `Controller` except that it doesn't have state, tags (no `resolve`) or router support. Ideally the primary cerebral Controller could be based on this Service.

Since the Service is a cut down controller the docs are almost the same:

# Service

```js
import Service from 'cerebral/lib/Service'

const service = Service({
  // Defines the top level signals
  signals: {},

  // Defines the top level modules
  modules: {}
})

export default service
```

## Errors
Cerebral knows about any errors that happen during a signal execution, synchronous and asynchronous. By default Cerebral just throws these errors to the console, but you can take control if you want to pass them to error tracking services etc.

```js
service.on('error', function (error, execution, functionDetails) {})
```

### getSignal(path)
Returns signal from Cerebral

```js
const someSignal = service.getSignal('some.signal')
// Run signal
someSignal({foo: 'bar'})
```

### runSignal(name, definition, payload)
Allows you to run an arbitrary function tree definition

```js
service.runSignal('someSignal', [actionA, actionB], {foo: 'bar'})
```

### initialized
Triggers when Cerebral service has initialized.

```js
service.on('initialized', () => {})
```

### start
Triggered whenever Cerebral starts a signal execution.

```js
service.on('start', (execution, payload) => {})
```

### end
Triggered whenever Cerebral ends a signal execution.

```js
service.on('end', (execution, payload) => {})
```

### pathStart
Triggered whenever Cerebral starts execution a path in a signal

```js
service.on('pathStart', (execution, payload) => {})
```

### pathEnd
Triggered whenever Cerebral ends execution a path in a signal

```js
service.on('pathEnd', (execution, payload) => {})
```

### functionStart
Triggered whenever Cerebral starts executing an action.

```js
service.on('functionStart', (execution, functionDetails, payload) => {})
```

### functionEnd
Triggered whenever Cerebral ends executing an action.

```js
service.on('functionEnd', (execution, functionDetails, payload) => {})
```

### asyncFunction
Triggered whenever Cerebral executed an async action.

```js
service.on('asyncFunction', (execution, functionDetails, payload) => {})
```

### parallelStart
Triggered whenever Cerebral executes actions in parallel.

```js
service.on('parallelStart', (execution, payload, functionsToResolveCount) => {})
```

### parallelProgress
Triggered whenever Cerebral executes actions in parallel.

```js
service.on('parallelProgress', (execution, payload, functionsStillResolvingCount) => {})
```

### parallelEnd
Triggered whenever Cerebral executes actions in parallel.

```js
service.on('parallelEnd', (execution, payload, functionsExecutedCount) => {})
```

### DIFF
--- a/packages/cerebral/src/Service.js
+++ b/packages/cerebral/src/Service.js
@@ -1,0 +1,104 @@
+import {FunctionTree} from 'function-tree'
+import ServiceModule from './ServiceModule'
+import {ensurePath, isDeveloping, throwError, isSerializable, forceSerializable, isObject, getProviders} from './utils'
+import VerifyPropsProvider from './providers/VerifyProps'
+import DebuggerProvider from './providers/Debugger'
+import ControllerProvider from './providers/Controller'
+import PropsProvider from './providers/Props'
+
+class Service extends FunctionTree {
+  constructor ({signals = {}, providers = [], modules = {}, devtools = null, options = {}}) {
+    super()
+    this.options = options
+    this.devtools = devtools
+    this.module = new ServiceModule(this, [], {
+      signals,
+      modules
+    })
+
+    this.contextProviders = [
+      PropsProvider(),
+      ControllerProvider(this)
+    ].concat((
+      this.devtools ? [
+        DebuggerProvider()
+      ] : []
+    )).concat((
+      isDeveloping() ? [
+        VerifyPropsProvider
+      ] : []
+    )).concat(
+      providers.concat(getProviders(this.module))
+    )
+
+    if (this.devtools) {
+      this.devtools.init(this)
+      this.on('error', function throwErrorCallback (error) {
+        if (Array.isArray(this._events.error) && this._events.error.length > 2) {
+          this.removeListener('error', throwErrorCallback)
+        } else {
+          throw error
+        }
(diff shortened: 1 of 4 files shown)


christianalfoni:
Quick question... what is the difference between Service and just using plain function tree?

garth:
Same difference as function tree and Controller. Function tree is low level, Service/Controller group many signals/trees, manage providers and have debugging tools built in.

christianalfoni:
So idea is to extend Cerebral and Cerebral-server from this? :)

garth:
I think that the idea is to provide this instead of cerebral-server. Then no assumptions are made about how you use it on server (or client).

cerebraljs:
Should we close it now and continue discussion?

Guria:
Should we close it now and continue discusdion?

PR #708, opened 2017-02-28
Forms refactor

christianalfoni:
Look at docs for API

I chose not go with JSON schema, cause it is just too much adjusting and its scope is quite big and covers a lot more than our little tool here. Think it is overkill currently.


christianalfoni:
Can not find module `cerebral`? Does a new package requires some config?

garth:
Mostly looks very good. Only query would be about errorMessages.

Instead of doing error message definition this when the provider is setup, could we not have the compute set the rule name and args so that the component can handle?

```js
{
  myForm: {
    firstName: {
      value: '',
      validationRules: ['minLength:3']
    }
  }
}

export default connect({
  form: form(state`path.to.form`)
},
  function MyForm ({form}) {
    // Value of some field
    form.firstName.failedRule // { name: 'minLength', args: [3]  }
  }
)
```

christianalfoni:
Okay, updated to:

```js
// Add regexp as a rule, name of rule is "regexp", arg is the regexp"
{
  field: {
    value: 'foo',
    validationRules: [/foo/]
  }
}
```

And form computed now states:

```js
function MyComp ({form}) {
  form.failedRule.name // regexp
  form.failedRule.arg // /foo/
}
```

There is only one arg actually

packages/cerebral-provider-forms/README.md
```diff
@@ -0,0 +1,392 @@
+# cerebral-forms
```
Guria: readme must be rewritten carefully with new package name and imports

garth:
shouldn't the failed rule be on the field?
```js
function MyComp ({form}) {
  form.field.failedRule.name // regexp
  form.field.failedRule.arg // /foo/
}
```

christianalfoni:
@garth Sorry, typo by me, it is on the field :)

packages/cerebral-provider-forms/README.md
christianalfoni: Updated pointing to new path after PR of sequence stuff.. changing "tools" to "packages"

christianalfoni:
@yusufsafak Thanks for additions :)

christianalfoni:
@fopsdev That is exactly right :)

packages/cerebral-provider-forms/src/rules.js
```diff
@@ -0,0 +1,73 @@
...
+  isFloat (value) {
+    return rules.regexp(value, null, /^(?:[-+]?(?:\d+))?(?:\.\d*)?(?:[eE][\+\-]?(?:\d+))?$/)
+  },
+  isWords (value) {
+    return rules.regexp(value, null, /^[A-Z\s]+$/i)
+  },
+  isSpecialWords (value) {
+    return rules.regexp(value, null, /^[A-Z\s\u00C0-\u017F]+$/i)
+  },
+  isLength (value, form, length) {
```
edgesoft: form is not used. Switch to _ ?

christianalfoni:
Okay, so a suggestion is to rename this package to:

`cerebral-provider-validation`

Cause there is really nothing "forms" specifically related here. It is just generally validating a state structure with values. I like the idea, though a con is that people probably looks for `forms`. Also I can not imagine validating anything else than a form? :)

tiagocpontesp:
Options that need to be selected or handled depending on other things. 
I think the browsers now bring a bunch of [validations](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-maxlength) on input[type]s, so I guess the main advantage here is the dependencies? I'm a bit out of the loop here, but I wanna suggest leaving the more generic name for a more generic schema-based provider (which perhaps this one could be made to use internally).

garth:
@christianalfoni I would leave it as forms, there are helpers for updating values too. And as you suggest people are probably looking for forms more than anything else.

garth:
@christianalfoni I think that it should be possible to manually set form.field.errorMessage on the state and have the form() report the field as being invalid.

This is useful for async stuff that you may not use validation functions for.

henri-hulski:
@christianalfoni As an example what you can validate beside forms:
[redux-json-schema](https://blog.prismatik.com.au/redux-and-json-schema-c63368931143) is validating the whole state tree of a redux app against a json schema.

Guria:
@christianalfoni 
`field` computed is a must have for me.
Also I'd like to get rid of "global" default validation rules. Let's make it explicit by providing them in provider config:
```js
import FormsProvider, {rules} from 'cerebral-provider-forms'

  providers: [
    FormsProvider({
      // Add additional rules
      rules: {
        ...rules,
        myAddedRule (value, form, arg) {
          value // value of the field
          form // the form field is attached to
          arg // arg passed to the rule

          return true
        }
      },

      // errorMessage property added to field when invalid with the following rules
      errorMessages: {
        minLength (value, minLength) {
          return `The length is ${value.length}, should be equal or more than ${minLength}`
        }
      }
    })
  ]
})
```

packages/cerebral-provider-forms/src/helpers/getInvalidFormFields.test.js
```diff
@@ -0,0 +1,60 @@
+/* eslint-env mocha */
+import getInvalidFormFields from './getInvalidFormFields'
+import assert from 'assert'
+
+describe('getFormFields', () => {
```
fmal: forgot to change group name to `getInvalidFormFields` 😃

packages/cerebral-provider-forms/src/utils/runValidation.js
```diff
@@ -0,0 +1,24 @@
...
+  const hasValue = checkHasValue(form, field.value, isValueRules)
+  const result = validate(form, field.value, field.validationRules || [])
+  const isValid = result.isValid && (
+    (field.isRequired && hasValue) ||
+    !field.isRequired
+  )
+
+  const validationResult = {
+    isValid,
+    hasValue: checkHasValue(form, field.value, isValueRules),
```
fmal: look like `hasValue` is already available on line 6, why re-running?

fmal:
@christianalfoni how would one specify error message for required fields ? Looks like 'isValue' is handled separately (not a part of `utils/validate.js`). In the old cerebral-forms we had `requiredMessage` prop on form fields.

packages/cerebral-provider-forms/src/form.test.js
```diff
@@ -0,0 +1,288 @@
...
+      signals: {
+        test: [
+          ({forms}) => {
+            const form = forms.get('form')
+            assert.ok(form instanceof Form)
+            assert.ok(form.name instanceof Field)
+            assert.equal(form.name.value, 'Ben')
+            assert.equal(form.name.isValid, false)
+            assert.deepEqual(form.name.validationRules, ['minLength:4'])
+            assert.equal(form.name.requiredMessage, null)
```
fmal: looks like some leftover. `requiredMessage` is not set anywhere and thus `undefined`

christianalfoni:
Okay, these are the updates:

1. Adding rules

```js
import FormsProvider from 'cerebral-provider-forms'
import * from 'cerebral-provider-forms/rules'

Controller({
  providers: [
    FormsProvider({rules})
  ]
})
```

With this approach you can add specific rules using `import {minLength, maxLength}`.

2. Can now use `field`

```js
connect({
  field: field(state`form.someField`)
}, ...)
```

Note that this only validates the specific field. If you use `equalsField` rule you will have to use the `form` computed for it to find the other field.

3. @fmal Thanks for reviewing, fixed the issues :) About the `requiredMessage` I added to docs that you check this with `field.isRequired && !field.hasValue`. There really should not be any text inside the state-tree, cause it is like a UI only thing. And adding `requiredMessages` to the options does not really make sense as required is usually rather a color indication or just a * on some text. But yeah, just my initial thought :)

4. @Guria Updated your notes as well :)

fmal:
@christianalfoni still I think it would be nice to be able to specify global message/translation key for empty values in required fields the same way that you can specify error messages for rules:

```javascript
FormsProvider({
  errorMessages: {
    minLength () {
      return 'error_message'
    },
    isValue () {
      return `required field can't be empty`
    }
  }
})
```

Then the computed would return `errorMessage` so we don't store any text in state tree. I think this would make things more consistent: it wouldn't force you to handle empty values errors in some special way.

christianalfoni:
@fmal Hm, I guess we could just return the rule that failed, it being a validationRule or isValue rule. So if you have `isRequired` on the field and the `isValue` fails, it uses the defined error for that type of error :) That is your reasoning as well?

christianalfoni:
Btw, not sure what happens on the tests here... everything runs green locally. Probably new setup stuff? Let me know if I can do anything :)

packages/cerebral-provider-forms/src/field.js
```diff
@@ -0,0 +1,26 @@
+import {compute} from 'cerebral'
```
edgesoft: @christianalfoni You name it field.js but makes from './Field' import

christianalfoni: ah, fudge, will fix :)

fmal:
@christianalfoni sorry for deleting my previous question. Yes, that is exactly the type of error i would like to define once when passing global error messages to the provider config.

Currently to achieve this i would need to do the following:

```js
FormsProvider({
  errorMessages: {
    minLength () {
      return 'error_message'
    },
    isValue () {
      return `required field can't be empty`
    }
  }
})
```
and then explicitly add the `isValue` rule to the form field:
```js
form: {
  name: {
    isRequired: true,
    validationRules: ['isValue']
  }
}
```

christianalfoni:
@fmal I see, yeah, that makes sense :) Looking into something else now, but we should fix that yeah

packages/cerebral-provider-forms/src/helpers/formToJSON.js
```diff
@@ -0,0 +1,25 @@
...
+        newObject[key] = extractObject(object[key])
+      }
+    }
+
+    return newObject
+  }, {})
+}
+
+function extractArray (array) {
+  return array.map((object) => {
```
fmal: I think you could do `array.map(extractObject)` here without creating the unnecessary wrapping 😃

PR #720, opened 2017-03-07
Latch on debugger

christianalfoni:
This pull request does the following:

- Update docs on standalone debugger
- Updates Firebase to pass execution details on tasks
- Added "apps" to standalone debugger and possibility to use function-tree in cerebral debugger, and combine them
- Changed title and added icon ;-)

We have tested this on our project now and it will probably require more iterations, but this is working and does not break existing.


Guria:
It's better to avoid such mixed PRs: API docs not related to changes and firebase API package change looks within debugger update. At least we should better split such changes by commits.

christianalfoni:
Its just that it is all connected. It looks very separate after the PR is done, but it is not during the development of the PR. Firebase is related to the new `executedBy` logic in debugger.

I agree that the docs are not relevant here, but instead of creating an extra commit (because I have to commit to create a new branch), I just shove it in on my current PR so I do not forget it. I could stash as well, but often I still have untracked files that moves over to the next branch and then a `git add .` messes up everything.

It is just safer for me to work on one PR at a time. That will give least number of commit and I do not forget stuff :)

If it was possible to work on multiple PRs isolated at the same time, that would be great, but I can not change branches in the middle of things and stashing does not stash untracked files aaaand I can get interrupted and I have no idea what I have in my stash when I get back.

Git is like putting a square in a circle for me ;-)

henri-hulski:
So maybe it's easier to separate them by commits. 
Just `git add doc/`and then commit the doc changes. After `git add packages/firebase/` to commit the firebase changes.

christianalfoni:
@henri-hulski Yeah, indeed... I should be able to discipline myself to make that work :)

Like, its my personality, Git is getting in my way and I get easily annoyed. I just want to write the freakin` code ;-) I know how important it is to structure and keep a good history, but when I am working on something I get so easily frustrated :D

christianalfoni:
Btw, no idea why one travis fails here?

henri-hulski:
```sh
 ● Transform proxies to template tags › should allow imported variable to be renamed
    expect(value).toMatchSnapshot()
    
    Received value does not match stored snapshot 1.
    
    - Snapshot
    + Received
    
    @@ -1,3 +1,3 @@
     "
     import { state as anotherName } from 'cerebral/tags';
    -state => anotherName`hello.${ anotherName`world` }`;"
    +state => anotherName`hello.${anotherName`world`}`;"
```
There are more similar. 
Just missing space inside braces.

christianalfoni:
@henri-hulski Yeah, this is so weird. Cause these spaces are different in different setups. So locally these tests always fails. But usually they work on CI env... if we run it again, I am sure they will suddenly work.

christianalfoni:
restarted now...

henri-hulski:
But stange that it fails only on the push and not on the PR.

christianalfoni:
I suggest we remove these tests until next iteration of them... use a different testing tool, same as the rest maybe? Really annoying to always have red tests locally :)

henri-hulski:
It's related to the babel-plugin-cerebral. 
Maybe some AST manipulation.

henri-hulski:
@FWeinb Have you an idea what's happening here?

Guria:
Just ignore babel-* stuff. I'll see it later today

Guria:
rewrited a bit of history in this PR :)

PR #721, opened 2017-03-09
feat(function-tree): abort tree execution from async action

maxfrigge:
@christianalfoni please have a look and see if this is ok and doesn't break anything I may have missed.

### DIFF
--- a/packages/function-tree/src/index.test.js
+++ b/packages/function-tree/src/index.test.js
@@ -1,0 +1,37 @@
+/* eslint-env mocha */
+import FunctionTreeExecution from './'
+
+describe('FunctionTreeExecution', () => {
+  describe('execution.abort()', () => {
+    it('should stop if aborted from sync action', (done) => {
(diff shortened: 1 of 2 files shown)


packages/function-tree/src/index.test.js
```diff
@@ -0,0 +1,37 @@
...
+      ])
+    })
+    it('should stop if aborted from async action', (done) => {
+      const runTree = FunctionTreeExecution([])
+      runTree.on('abort', () => done())
+
+      runTree([
+        ({execution}) => {
+          return new Promise((resolve) => {
+            setTimeout(() => {
```
christianalfoni: Actually does not need timeout, promises are async in nature, so you can just return:

```js
  return Promise.resolve(execution.abort())
```

maxfrigge:
@christianalfoni refactored the test code but messed up the history after rebasing :-/

Guria:
@maxfrigge rebased again

PR #722, opened 2017-03-09
chore(Debugger): cleanup extra setup

Guria:


christianalfoni:
Okay, so I have given this a lot of thought and I think it is wise to make a decision about the debugger. Currently the standalone and chrome extension has been the exact same experience, but they have started to diverge. The standalone debugger has limitless possibilities, but the Chrome Extension will always have its restrictions in addition to these problems:

1. Chrome extension has some memory leak I can not figure out. It does not happen in Electron. Basically browser just becomes super slow after a while
2. There is this superweird bug where scrolling causes a bunch of signals to appear
3. The scrolling of signals/state-tree container stops working. This happens  a lot. You have to resize the devtools window to make it work again
4. Publishing, verifying and making sure new features works in Chrome Extension and Electron is just a maintenance hell
5. It does not work with the Webpackbin tutorial, due to it running in an iframe
6. We can not control the publishing of the Extension, risking breaking apps as suddenly there is a new version of debugger that does not work with your current version of Cerebral/Function Tree

So the choice here is to put a lot of efforts into keeping the chrome extension around, loosing efforts on improving the debugger experience in general... or just put all the little Cerebral debugging eggs in one basket and focus on the standalone version that has no limitations.

The thing is, I have always dreamed about the debugger becoming more than a debugger. Imagine that instead of having a CLI tool, you could actually create your app from the standalone Cerebral "IDE". It would create a folder, insert folder structure, webpack config with connection details to debugger etc.

You would be able to compose your signals in the standalone app with drag and drop of operators. Any custom actions can be named and files are created for them by the app. Basically you do all your coding in Atom or whatever, but we can give a visual experience of planning out logic. 

And there is so much more we can do even beyond this :)

As I told @Guria , I am just exhausted fighting with Chrome over the last year and I really think this strategy of focusing on one experience that is the same for any developer, on any platform and any browser makes a lot of sense :)

Chrome Extension will still live on until there are is breaking change.

edgesoft:
I think this is a great idé! Haven't used the standalone yet but whatever makes Cerebral move foreward and creates more productivity

NervosaX:
I just want to point out that the upsides of the chrome debugger is it is easy to work on multiple applications easily... switching to another app is simply just changing your tab. How can ease of switching project be addressed with the cerebral standalone debugger? Do you have to open multiple apps and manually connect them?

> The thing is, I have always dreamed about the debugger becoming more than a debugger. Imagine that instead of having a CLI tool, you could actually create your app from the standalone Cerebral "IDE". It would create a folder, insert folder structure, webpack config with connection details to debugger etc.

I like the idea about folder structure, but we should also be careful about forcing build tools on people. It could just as easily be browserify or jspm. 

Maybe a better route is to modularize parts of the debugger... have the parts that chrome extensions can use included, and allow the app to diverge with those parts plus more?

bfitch:
@NervosaX the latest standalone debugger has support for running multiple apps: https://www.youtube.com/watch?v=PX0_H4XSnVM

NervosaX:
Ah yes, I see. I think the memory leak issue alone is a reason to drop chrome extensions entirely for this application. I guess my one last issue could be connecting it to other applications (like nwjs?) Currently I can drop on the chrome extension into our nwjs app and have it load up. Good way to debug it.

Guria:
@NervosaX anyway it wasn't a case for chrome extension. it always was the same instance of debugger for each tab and caused some weird bugs. I still would like to see extension but in current state it is not very real.

gaspard:
Hmmm, as much as I understand your frustration and pain, having a new window for the debugger is a pain in my workflow. I have a split window with editor on one side and app+debugger on the other side. With a standalone debugger, I'm back in window switching hell again.

So if running the debugger "inside" the app does not work, why not run the app in the debugger ?

This means that instead of using "chrome" to debug Cerebral apps, we could run them in the "Cerebral browser".

christianalfoni:
Yeah, sorry about that @gaspard. Any reason you can not put the debugger underneath your browser? on the right of your split screen?

I think it is impossible to support every workflow, cause there are so many of them and every one of them has pain points. Like the extension from a workflow perspective:

1. Can not look at console at the same time as debugging Cerebral
2. The window is very small, lots of scrolling on large signals and have to click tab to look at state tree (if using it on right side)
3. The scolling bug is major pain. Have to resize the devtools all the time... does not happen to you?

Like, I do not disagree with any ideas on it being beneficial to make it work as an extension, splitting it up into reusable components for different environments, or research putting browser window into standalone debugger etc. It is just insane amount of work. Work that can rather be put into improving the debugger itself.

Maybe we should create a new channel "workflow" where we can rather discuss our setups? How do we work with the debugger, IDEs, browser etc. I am sure it will take a lot less time finding a good workflow with a standalone debugger, than making the standalone debugger work in any workflow ;-)

gaspard:
Hey, I agree.

And I'll do the work in the stand-alone debugger to support web app + debugger in the same application window. It should not be a huge amount of work, we'll see.

christianalfoni:
@gaspard There is a browser plugin for Atom you can take a look at: https://atom.io/packages/atom-webbrowser. Maybe that is helpful? :)

gaspard:
Hmmm, I think it really makes sense to have the debugger "as close as possible" to the app.

The water and the baby...

I really feel like we are dropping something valuable for the wrong reasons. If the current bugs and issues in the extension is a pain, I am all for using alternatives but not for dropping the "in-app" debugger altogether by adding breaking changes in the standalone debugger app...

I am sure there is a way to go forward with the standalone debugger without introducing breaking changes in the extension so that we can someday figure out how to fix the bugs you listed and continue with this thing a lot of developers have become used to.

christianalfoni:
@gaspard Since extension and standalone has the same codebase every new feature has to be evaluated and discussed in the context of both environments. Also workarounds has to be created to make only standalone features work there and be ignored in extension. It is not a good place to be to move on with the debugger.

The only thing the extension has going for it is that it is inside the browser. All other features works better in standalone, including several features not even possible in extension. 

Slowing down development in discussions, in implementation, in maintenance, in publishing and share motivation for working on the project... it is just not worth it :)

That said! If you want, maybe you want to copy over the previous setup for Cerebral debugger? Happy to see that project run separately. No reason why that could not work.

```js
import Devtools from 'cerebral-devtools-chrome'
``` 

? :)

PR #724, opened 2017-03-10
Replace RunSignal test helper with CerebralTest helper

garth:
RunSignal is still there but will show deprecation warning.

Test helper supports `getState` and `setState` as well as `runSignal`.

```js
import {CerebralTest} from 'cerebral/test'

it('should accumulate a count', () => {
  const cerebral = CerebralTest({
    modules: {
      math: math()
    }
  })
  cerebral.setState('math.count', 0)
  return cerebral.runSignal('math.plusOne').then(({state}) => {
    assert.equal(state.math.count, 1)
    return cerebral.runSignal('math.plusTwo').then(() => {
      assert.equal(cerebral.getState('math.count'), 3)
    })
  })
})
```

### DIFF
--- a/docs/docs/docs/developer_guide/test.md
+++ b/docs/docs/docs/developer_guide/test.md
@@ -190,30 +190,33 @@ it('should handle button clicks', () => {
 })
 ```
 
-### RunSignal factory
+### CerebralTest factory
 
-The `RunSignal` factory is similar to run signal except that it will return a runSignal function that can be called many times without resetting the controller in between.
+The `CerebralTest` factory returns runSignal, setState and getState functions that can be called many times without resetting the controller in between.
 
 ```js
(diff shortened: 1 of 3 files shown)


garth:
You can always ignore one. The state param is from the underlying runSignal which is used elsewhere, so probably won't remove it.

christianalfoni:
I think I need to distribute testing responsibilities to you guys :) Think it is important to have hands on experience and feel the pains to make a good API. I do not really do much testing on client side, so not a good candidate ;-) BUT! I am contributing with a test tool the firebase tasks on a new `function-tree-firebase-admin` coming soon.

So yeah, only thing is that this page needs to be in sync with testing: https://cerebral.github.io/docs/developer_guide/test.html

PR #736, opened 2017-03-13
Fix lib exports and move forms docs from tools to packages

garth:


packages/cerebral-provider-forms/index.js
```diff
@@ -0,0 +1 @@
+module.exports = require('./lib')
```
Guria: it's already covered with main property in package.json. No need to have extra file

garth: This is just the same as main cerebral package, but I will remove.

PR #737, opened 2017-03-14
Tutorial review

fopsdev:


fopsdev:
Also had to change 2 bins:
Action Bin:
https://www.webpackbin.com/bins/-Kf6G5AryR--9JX9xZ8M
-> removed wait operator because it will be introduced in next chapter

Compute Bin:
https://www.webpackbin.com/bins/-Kf6zrpYkwThZ6DpI8FS
-> made the initial bin working withe the "naive" approach, (users are tempted to run stuff when loading a bin...)

Issue #612, opened 2017-01-20
Debugger doesn't show state key:value if value is undefined


idream3:
We get an error in the console:

![image](https://cloud.githubusercontent.com/assets/1082632/22130987/c6ed4598-df14-11e6-9ec6-9d4e69790fe6.png)

But is a bit cryptic. Is there a way to print out the value type? undefined, function etc... in this case it is undefined. the key:value 'user:undefined' does not show in the debugger:

![image](https://cloud.githubusercontent.com/assets/1082632/22131002/e0d5aa7c-df14-11e6-86e8-c2daa403bbab.png)

I assume this is because it cannot be toJSON'd for the debugger, but wondering if we can indicate this some how.

`user: '[undefined]'`
`user: '[function]'`

christianalfoni:
I suggest we wrap the `StateProvider` method calls in a `try/catch`. That way this error can be caught inside the signal and we can populate it with more information. Like the signal name, action name...

```js
  function StateProvider (context, functionDetails) {
    context.state = provider = provider || createProvider(context)

    if (context.debugger) {
      context.state = methods.reduce((currentState, methodKey) => {
        if (methodKey === 'get' || methodKey === 'compute') {
          currentState[methodKey] = provider[methodKey]
        } else {
          const originFunc = provider[methodKey]

          currentState[methodKey] = (...args) => {
            const argsCopy = args.slice()
            const path = ensurePath(argsCopy.shift())

            context.debugger.send({
              datetime: Date.now(),
              type: 'mutation',
              color: '#333',
              method: methodKey,
              args: [path, ...argsCopy]
            })

            try {
              originFunc.apply(context.controller.model, args)
            } catch (e) {
              const path = args[0]
              const type = typeof args[1]
              const actionName = functionDetails.name
              const signalName = context.execution.name

              throwError(`Invalid value of type ${type} passed to state tree... bla bla bla`)
            }
          }
        }

        return currentState
      }, {})
    }

    return context
  }
```

christianalfoni:
We could serialize and show it in state tree, though it really is the wrong type and might not be caught unless we error it... causing problems "down the line" of sending to server, local storage etc.

idream3:
I suggest we error undefined values in development evn.

On Tue, Mar 14, 2017 at 10:11 PM, Christian Alfoni <notifications@github.com

PR #740, opened 2017-03-14
refactor(Providers): add descriptive error when invalid value passed …

bannerintheUK:
…to state tree

ISSUES CLOSED: #612


packages/cerebral/src/viewFactories/Hoc.js
```diff
@@ -161,7 +161,14 @@ export default (View) => {
             } else {
               currentProps[key] = value
             }
-          } else if (tag.type === 'signal' || tag.type === 'props') {
+          } if (tag.type === 'signal') {
```
henri-hulski: `else` is missing

PR #745, opened 2017-03-15
fix(forms): Allow isValidForm operator to work with any tag or compute

### LINKED ISSUE #744: Forms isValidForm operator fails when compute is provided
So long as the value passed to isValidForm can be resoled to a form then it shouldn't matter where the form comes from.

The following currently throws `Cerebral Forms - isValidForm factory requires a STATE TAG`, this check should probably be removed.

```js
import { isValidForm } from 'cerebral-provider-forms/operators';
import ItemForm from '../../computes/ItemForm';
export default [
  isValidForm(ItemForm(props`id`)),
  {
    true: [],
    false: []
];
```

garth:
Also add check in form to ensure that the tag/compute resolves to an object

fixes #744

### DIFF
--- a/packages/cerebral-provider-forms/src/form.js
+++ b/packages/cerebral-provider-forms/src/form.js
@@ -61,8 +61,9 @@ export default function computedForm (formValueTag) {
   return compute(
     formValueTag,
     (formValue) => {
-      if (typeof formValue !== 'object') {
-        throw new Error('Cerebral Forms - Form requires a tag or compute that resolves to an object')
+      if (!formValue || typeof formValue !== 'object') {
+        console.warn('Cerebral Forms - Form value did not resolve to an object')
+        return {}
       }
       return new Form(formValue)
     }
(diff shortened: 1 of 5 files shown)


packages/cerebral-provider-forms/src/form.js
```diff
@@ -61,6 +61,9 @@ export default function computedForm (formValueTag) {
   return compute(
     formValueTag,
     (formValue) => {
+      if (typeof formValue !== 'object') {
```
christianalfoni: This would still fail on `null` actually. Have to check for that as well :) Possibly also `!Array.isArray(formValue)` ... not sure how an initial array value be handled, maybe it works? No test for it though

garth: I think that there are valid cases where a formTag/compute may sometimes resolve to undefined or null (such as on startup). Maybe we need to handle these cases without a throw?

christianalfoni: Hm, yeah, good point... if it is not an object, just return the value directly. That way it is easy to debug as well.. cause problems will occur in the component, where you try to access something that is not there :)

garth: So how about adding a warning
```js
console.warn('Cerebral Forms - Form value did not resolve to an object');
```

edgesoft:
LGTM

PR #746, opened 2017-03-15
Firebase admin

christianalfoni:
- Fixed bug with firebase provider, where error path was not gone down when task rejects
- Added `function-tree-firebase-admin`, binding tree execution to Firebase Queue Tasks
- Allows for merged execution
- Cleaned up function-tree stuff not needed


packages/function-tree-firebase-admin/package.json
```diff
@@ -0,0 +1,51 @@
...
+  "main": "lib/index.js",
+  "scripts": {
+    "build": "BABEL_ENV=production babel src/ --out-dir=lib/ -s",
+    "coverage": "nyc --reporter=lcov --reporter=json npm run test",
+    "prepublish": "npm run build"
+  },
+  "devDependencies": {
+    "@cerebral/monorepo": "0.0.1"
+  },
+  "dependencies": {
```
edgesoft: dev dependencies missing faye-websocket and proxyquire

packages/function-tree-firebase-admin/src/TestTasks.js
```diff
@@ -0,0 +1,104 @@
+import FunctionTree from 'function-tree'
+import originalWebsocket from 'faye-websocket'
+import Provider from './Provider'
+import proxyquire from 'proxyquire'
+import FirebaseServer from 'firebase-server'
+import FirebaseProvider from './Provider'
+
+const firebase = proxyquire('firebase', {
+	'faye-websocket': {
```
Guria: tabs here?

christianalfoni: It was fixed? There is something really weird happening there... I could not remove the tabs. Tried everything, even a tool to remove tabs. So I manually used spaces... but now they are tabs again? But all checks has passed?

christianalfoni: Cause I did get linting errors on that initially

christianalfoni: Oh, it is fixed?

Issue #734, opened 2017-03-13
feat(function-tree): pass payload when aborting


maxfrigge:
It's currently not possible to pass a payload when aborting the tree execution. 

The suggestion is `execution.abort(payload)`.

The payload will then me merged with the previous payload and passed along with the abort event.

I am currently swamped, but will create a PR a.s.a.p.

christianalfoni:
Think it is a great idea :)

maxfrigge:
@christianalfoni is it ok to shallow merge with `Object.assign({}, payload, result)` should outputs be deep merged?

christianalfoni:
https://github.com/cerebral/cerebral/pull/756

christianalfoni:
Do not need deep merge :)

PR #756, opened 2017-03-16
feat(function-tree): pass payload when aborting

maxfrigge:
ISSUES CLOSED: #734

### DIFF
--- a/packages/function-tree/src/index.js
+++ b/packages/function-tree/src/index.js
@@ -70,7 +70,7 @@ class FunctionTreeExecution extends EventEmitter {
     }
 
     if (result instanceof Abort) {
-      const finalPayload = Object.assign({}, payload, result ? result.payload : {})
+      const finalPayload = Object.assign({}, payload, result.payload)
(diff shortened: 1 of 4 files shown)


packages/function-tree/src/index.js
```diff
@@ -70,7 +70,8 @@ class FunctionTreeExecution extends EventEmitter {
     }
 
     if (result instanceof Abort) {
-      return functionTree.emit('abort', execution, funcDetails, payload)
+      const finalPayload = Object.assign({}, payload, result ? result.payload : {})
```
christianalfoni: You have to check `result.payload ? result.payload : {}` :)

maxfrigge: sorry.. that was a copy and paste without much thinking :)

maxfrigge:
@christianalfoni I have update the PR. Rather then checking when passing, I gave the payload a default of `{}` in the Abort constructor. I think it's cleaner that way.

PR #760, opened 2017-03-19
toJSON added to provider

ozanturksever:
docs suggests toJSON in provider


Guria:
Agreed. @ozanturksever could you add a tests pleas?

PR #762, opened 2017-03-19
docs(website): add favicon, review BETA and add collaboration

christianalfoni:

PR #763, opened 2017-03-19
feat(website): add public folder

christianalfoni:


Guria:
merged as part of #762

PR #776, opened 2017-03-23
Fix errors with codemods and add a basic usage readme

NervosaX:
I tested the codemods on my cerebral 1 codebase and came up with some minor issues. I've also added a basic readme on usage.


Guria:
Lets add a link to codemods readme on website/migration

PR #788, opened 2017-03-23
Expose field on cerebral-provider-forms

edgesoft:
This will allow 

```js
import {field} from 'cerebral-provider-forms'

export default connect({
 myField: field(state`some.field`)
})
```

Validate will run and you will have `myField.isValid` etc


edgesoft:
@Guria Ok, now i'm done. Added some docs and did some rebase :)

PR #775, opened 2017-03-22
feat(function-tree): add abort sequence to sequence and parallel

christianalfoni:
- Added **abort** logic to function tree, with new tests
- Added **AbortProvider** to function tree, making `abort` available directly on context
- Updated docs
- Updated debugger to show abort sequence inline, just like the demo of merged client/server execution

I have already released new debugger as nothing breaking there.

Issue #670, opened 2017-02-09
Exceptions in compute are lost


garth:
Exceptions thrown in compute functions are not logged and the compute returns undefined without indicating that something failed.

Ideally cerebral should rethrow when computes throw exceptions.

christianalfoni:
Does this still happen and if you have time, could you create a PR with a test where it fails? I can jump on and look at it :)

edgesoft:
Can we close this? @Guria @christianalfoni @garth

PR #793, opened 2017-03-23
feat(Controller): adjust error throwing and catching

christianalfoni:
OKAY! So after the PR of https://github.com/cerebral/cerebral/pull/775 a huge discussion started. It was a long, somewhat confusing and a little bit frustrating discussion, but it was an important one and really glad we had it! Thanks to everyone involved for patience and understanding :)

The previous PR created a new concept of stopping execution, this is not ideal. We should always try to adjust to existing solutions first and the existing solution to stop execution in JavaScript is to `throw`. So by forcing us to find a concept that fits our needs within the limitation of normal `throwing` I think we have come to a better, more familiar and even more powerful concept.

The spark behind this RFC was to be able to write more vertical code:

```js
// Instead of this
[
  getThis, {
    success: saveThis,
    error: showError
  }
]

// You would write this
[
  getThis,
  saveThis
]
```

The big question is... how do we stop execution when `getThis` has an "error"? With this PR you will just throw:

```js
function getThis ({http}) {
  return http.get('/this')
    .then((response) => ({data: response.result})
    .catch((response) => throw new Error(response.error))
}
```

This error does not have to be a plain error though, you might rather want to extend your own error class like this:

```js
function getThis ({http}) {
  return http.get('/this')
    .then((response) => ({data: response.result})
    .catch((response) => throw new HttpError(response))
}
```

The `http-provider` and `firebase-provider` does export its own error types you can specifically handle.

In low level function-tree, errors goes to a normal `catch`. So when you execute a tree:

```js
runTree('some_name', [someFunc], {foo: 'bar'})
  .catch((error) => {})
```

This makes function-tree become a promise on functional steroids :)

In Cerebral we do not handle the tree at this low level, so we are free to make this more explicit by:

*someModule.js*
```js
{
  state: {},
  signals: {
    somethingHappened: [],
    somethingThatCanErrorHappened: {
      signal: [],
      catch: new Map([
        [Error, showError('wuuuut?')]
      ])
    }
  }
}
```

Now you are able to define two parts to a signal. Its normal execution and what should happen on an error. With no catch, errors are thrown to the console.

It was decided to use a `Map` to define what errors you want to handle as that allows us to check the `instanceof` under the hood. Making you able to act on specific errors, but catch them all generally in some "lower priority" error handler.

The catch sequences will have an `error` property added to its payload. It contains the error serialized as `{name: 'Error', message: 'whatevah', stack: '...'}`, or whatever is is defined on the `toJSON` method of the error.

So by default all errors are thrown to console, it is only when you insert a catch for a signal that you will start handling them. No matter if you filter out error, the debugger will show you exactly what action caused the error, what the error was and optionally it will show the catch execution if defined. 

(it is called "catch" in this PR)
screen shot 2017-03-23 at 14 16 50

You can still listen to errors globally like normal:

```js
controller.on('error', (error) => {})
```

These events are triggered no matter if you catch or not, good to use for error tracking in production.

Even though this has taken some time and seems like a massive change, it really is not. You can write your Cerebral application exactly as you do now. No breaking change here. It is only when you get into very complex signals and you want to "verticalize" them you now have the possibility to do so, and it is based on a core javascript concept of try/catch :-)

## Summary

- Function-tree execute functions now returns a promise if no regular callback is provided
- Catched errors execution is inlined with action that caused it in the debugger. You can now even follow a client execution, that triggers a server execution, that errors and runs a catch execution... pretty awesome!
- Debugger now fades out actions that are not run (general improvement), also other general improvements made to debugger
- The `function-tree-firebase-admin` package has changed its API
- `abort` and `retry` are now completely gone. They were not very well thought out ideas. If you have been using `abort`, throw an error instead
- When I played around with this I found issues with `addModule`, now modules are added correctly
- Signals can now be defined as `{signal: [], catch: []}` or `{signal: [], catch: {}}`
- Moved position of "architecture" docs, one step up
- Fixed parallel execution to not continue when error has been thrown
- Tutorial is updated using an example of error throwing
- `http-provider` and `firebase-provider` now exports errors

@Guria Sorry for mixing stuff again, but it was no "first I want to do this and then this", I discovered all this stuff along the way to implement it all.. was afraid to forget something. Though maybe split PR on commit per folder or something?

----

**This description was updated according final discussions and changes**


fopsdev:
what a walktrough @christianalfoni :+1:  ... fighting other bugs along the way in the forest of fresh and prosperous code ... now reaching the top of the hill with the sun in the face... wohooo :)
having the catch on signal level for now is cool.
and if there is a real need i somehow feel with this architecture it should also be possible to do another iteration and put it as well on sequence resp. on the module level.
but hey step by step :)

packages/cerebral/src/Controller.js
```diff
@@ -166,7 +166,30 @@ class Controller extends FunctionTree {
...
+          return currentModule ? currentModule.modules[key] : undefined
+        }, this.module)
+
+        if (!signalCatch) {
+          throw error
+        }
+
+        if (Array.isArray(signalCatch)) {
+          this.runSignal('catch', signalCatch, error.payload)
+        } else if (error.name in signalCatch) {
```
henri-hulski: Not better check for isInstance instead of error.name?
This would allow to use groups of errors.
You could have e.g. a HTTPError class and some subclasses depending on the HTTP status code (ForbiddenError, NotFoundError, ConflictError,  UnprocessableEntityError etc.).
If you want to treat all HTTP errors the same you can use HTTPError which includes all subclasses. Or you can handle some subclasses individually. 
Not sure if this works in Javascript ;-).

tiagocpontesp: suggest allowing primitives for consistency, not just sequences as arrays (if not that way already). commented here since this is the snippet that raised my suspicion.

henri-hulski: Ideally you would first handle the subclasses you are interested in and after the baseclass which would handle all errors which were not caught by the subclasses before.

gaspard:
Nice work @christianalfoni. I think this is a wise decision. "catch" on sequence could have become a complicated thing and handling errors at the signal level makes sense.

Having the "catch" near the sequence makes more sense if you tend to define signals (by referencing a chain or grouping actions) inside the module's index file.

Now that "catch" is defined at the signal level, I feel that it makes more sense to define signals in files instead of inside the module because of the proximity of actions and error catching code.

Anyway, thanks a lot for being such a great listener and for putting all this work into Cerebral. 👍

christianalfoni:
@fopsdev Yeah, exactly, we can build on it. What came apparent when experimenting with this approach is that you always know the signal name looking in the debugger, meaning you also know the debugger... meaning you know what file to check to find the catch handler. If we add "catch intercepteros" inside the composition it becomes a bit less apparent where it is defined in code, but again... that is why we experiment :) Lets move in that direction if this implementation becomes lacking

@henri-hulski @tiagocpontesp It would indeed be good with `instanceof`, but that requires the code to have access to the constructor. Can not check `someError instanceof`, you have to check it against something: `someError instanceof SomeError`. So not sure how to adjust the API to define by constructor instead of name. Would have to create a WeakMap for example:

```js
{
  somethingHappened: {
    signal: [],
    catch: new WeakMap({
      [MyCustomError]: []
    })
  }
}
```

@gaspard @henri-hulski Thank you for kind words, that goes straight into my community-love bank ;-)

christianalfoni:
What I do see here though is that packages like http/firebase should probably be refactored.

The reason is that now they convert errors to payload object. Meaning:

```js
http.get('/item')
  .catch((payload) => {
    payload // {error: 'some error text'}
  })
``` 

This was to make it possible to do:

```js
http.get('/item')
  .catch(path.error)
``` 

What I realise now is that this is not expected behaviour. A catch should always get an instance of an error. That is what you would expect. It would also reduce a lot of boilerplate code in packages and preventing inconsistencies between packages.

I will make an update to this PR with that refactor. It is breaking, but it is the cherry on top to make Cerebral completely predictable in terms of error handling, and I think it is good timing as people very likely will start refactoring their code to be vertical in these instances. So to explain in code:

```js
function someAction ({http, path}) {
  return http.get('/something')
    .then(path.success)
    .catch(path.error);
}
```

These would need to be refactored in one of three ways:

1. Update catch
```js
function someAction ({http, path}) {
  return http.get('/something')
    .then(path.success)
    .catch((error) => path.error({error: error.message}));
}
```

2. Just remove the catch and error path

```js
function someAction ({http, path}) {
  return http.get('/something')
    .then(path.success)
}
```

If an error occurs it is thrown to console.

3. Refactor to vertical and remove both paths

```js
function someAction ({http, path}) {
  return http.get('/something')
}
```

What we achieve here is that our package wrappers around http/firebase etc. is closer to the original API, which is always a good thing. Also .catch always receives an error, it being JS error or package error. It is just what you would expect .catch to contain.

henri-hulski:
Using class inheritance for errors could be really a powerful pattern. And I think it would be quite expected. 
At least that is the way how it works in Python. ;) 
Though not sure how to implement. 

BTW @tiagocpontesp was talking about something different.

christianalfoni:
@henri-hulski I suggest I allow the `catch` to be a function. That way it is possible to experiment with it at least :)

maxfrigge:
@christianalfoni looks great. I would prefer to leave `abort()` as is. I am currently using it to intentionally end the tree execution.. throwing a StopExecutionError would be a hacky workaround. It would also display as an error in the debugger, when it's not an error.. just a prematurely finished execution - no pun intended

gaspard:
For firebase, I think it is really a good idea to use your last proposition without any reference to path.success. We want as few as possible boilerplate also in providers. It shows the quality of Cerebral API. :-)

edgesoft:
@christianalfoni Is this related as well? #670

packages/cerebral-provider-firebase/src/onDisconnect.js
```diff
@@ -15,7 +16,7 @@ export function setOnDisconnect (path, value) {
 
 export function cancelOnDisconnect () {
   if (!ref) {
-    throw new Error('cerebral-provider-firebase: You have no set to disconnect')
+    throw new FirebaseProviderError('You have no set setOnDisconnect')
```
henri-hulski: "set setOnDisconnect" maybe one set to much?

packages/cerebral-provider-http/src/utils.js
```diff
@@ -86,20 +86,32 @@ export function parseHeaders (rawHeaders) {
...
+      if (path[error.message.status]) {
+        return path[error.message.status]({error: error.payload.error})
+      }
+
+      if (path.error) {
+        return path.error({error: error.payload.error})
       }
 
-      return callNextPath(response, path, 'error')
+      throw error
```
henri-hulski: No custom `HTTPError`?

christianalfoni: Ah, at this point it actually is the `HttpProviderError` :)

packages/cerebral-provider-firebase/src/onDisconnect.js
christianalfoni: 👍

christianalfoni:
@henri-hulski New update with docs now :)

docs/developer_guide/errors.md
```diff
@@ -0,0 +1,97 @@
...
+          return []
+        }
+      }
+    }
+  }
+}
+```
+
+With a function you get the error instance and matches expect to return an action sequence to trigger.
+
```
henri-hulski: Cool man! 
So error inheritance is now possible! 😊

packages/cerebral-provider-http/src/utils.js
henri-hulski: 👍

docs/packages/index.md
```diff
@@ -55,6 +55,17 @@ function updateDefaultHttpOptions({http}) {
...
 
+## error
+Throws **HttpProviderError**. Its message property contains:
+
+```js
+{
+  result: 'Message or response body',
+  status: 200,
+  isAborted: false
+}
```
henri-hulski: Hmm I think it can also contain a header object. 
See https://github.com/cerebral/cerebral/blob/ftErrors/packages/cerebral-provider-http/src/DEFAULT_OPTIONS.js#L46

docs/developer_guide/errors.md
christianalfoni: Yes :)

docs/packages/index.md
christianalfoni: Yeah, good point, let me fix that!

christianalfoni:
Okay, so we will be test driving this PR tomorrow. It has been working great on server side and will do client side refactors tomorrow :)

christianalfoni:
Righty tighty, refactoring in client has been going well. We figured out that also a `FirebaseProviderAuthenticationError` was nice to have. Allowing you to explicitly act on problems with authentication.

We are finishing the refactoring now and will have a status on this tomorrow (end of day here soon). Technically everything is looking sharp!

henri-hulski:
@christianalfoni Really difficult for me to get into the flow. Where you get the error in controller and is it not an instance of the errorclass which was catched?
Ideally the catch could look something like the above:
```js
catch: [
  except(ExpiredSignatureError, []),
  except(InvalidTokenError, [])
]
```
and the except factory something like
```js
function exceptFactory (errorClass, actions) {
  function except (error) {
    if (error instanceof errorClass) {
      return actions
    } else {
      throw error
    }
  }

  return except
}

export default exceptFactory
```
But that would need an array of functions not a single function.
So if the error is not an instance of the Error class you want it will rethrow and the next function would check again until it is caught.
When no Error class matches the error goes uncatched to the console.

But I don't know if we have access to the catched error and if we really can rethrow it.

What you think?

Your proposition
```js
{
  somethingHappened: {
    signal: [],
    catch: catchErrors(MyError, SomeOtherError, [])
}
```
seems not make sense as handle all E
Error classes the same.

tiagocpontesp:
I don't see what's wrong with enforcing a Map (if filtering), going into the first match _pronto_. 
Seems better looking, easier, fitting all scenarios and standard:
http://bluebirdjs.com/docs/api/catch.html#filtered-catch
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#Conditional_catch_clauses

christianalfoni:
@henri-hulski @tiagocpontesp 

I suggest we just check the cache value for `instanceof Map`, that way you can write this:

```js
{
  cache: new Map()
    .set(AuthenticationError, [])
    .set(SomeOtherError, [])
}
// Or alternative map definition
{
  cache: new Map([
    [AuthenticationError, []],
    [SomeOtherError, []]
  ])
}
```

Just plain JavaScript. Under the hood it will do a `cache.forEach()` when its a Map that is used to define them.... plain old... well... plain new JavaScript :)

henri-hulski:
@christianalfoni cache?

christianalfoni:
@henri-hulski Mind crash... catch ;-)

henri-hulski:
@christianalfoni exept I have stolen from Python:
```python
try:
    x = int('not_a_number')
except ValueError:
    print('Not a number')
```
You could also allow to except several Error Classes:
```js
catch: [
  except([ExpiredSignatureError, MissingRequiredClaimError], []),
  except(InvalidTokenError, [])
]
```

henri-hulski:
Just an example how I use it in [more.jwtauth](https://github.com/morepath/more.jwtauth/blob/master/more/jwtauth/main.py#L364-L378):
```python
        if not self.allow_refresh:
            raise InvalidTokenError('Token refresh is disabled')

        token = self.get_jwt(request)
        if token is None:
            raise InvalidTokenError('Token not found')

        try:
            claims_set = self.decode_jwt(
                token, self.verify_expiration_on_refresh
            )
        except DecodeError:
            raise DecodeError('Token could not be decoded')
        except ExpiredSignatureError:
            raise ExpiredSignatureError('Token has expired')
```

christianalfoni:
Okay, I can report that our finished refactor today has been super successful. This is a huge improvement and I got a lot of stuff fixed for the debugger as well. Things are running very smoothly now!

So, I am signing off this PR concept as "ready to go" :-)

NOW! One final thing is to express what to catch. In current PR it is using the "name" of an error, but that can be limiting as for example `FirebaseProviderError` could be a base error, where `FirebaseProviderAuthenticationError` would be a more specific error. This allows you to:

1. Not care about errors at all
2. Just care about general Firebase Provider errors
3. Care about specific errors

It would mean you had to import the classes like this:

```js
import {
  FirebaseProviderError,
  FirebaseProviderAuthenticationError
} from 'cerebral-provider-firebase'
```

Which is perfectly okay I think.

The question is how you "filter out" what errors you want to handle. First I think it is wise that we allow error instances to be passed into signals and state tree. That means when the `catch` signal/tree runs it will have the actual error instance on `{errr: Error}`. That allows us to check `instanceof` inside the execution. At the same time each errors defines how they are going to be serialized, which is good!

SO! The thing is that we can not introduce operators to handle this. The reason is that when the operator does not match the error it can not throw it, because that would stop further error matching. It would also show an error on the action of the catch execution in the debugger, which is not what we want.

That means we must have a "matching constructor" between the error and what signal/tree is should execute. There are three ways to do this:

1. Current approach, only use `Map` instead:

```js
{
  catch: new Map([
    [SomeError, []],
    [[OneError, TwoError], []]
  ]),
  // Or by setting
  catch: new Map()
    .set(SomeError, [])
    .set([OneError, TwoError], [])
}
```
It is just JavaScript, but Map is pretty unknown. 

2. Make `catch` a list of catch handlers
```js
{
  catch: [{
    match: SomeError,
    signal: []
  }]
}
```

3. Use a function

```js
{
  catch(error) {
    if (error instanceof SomeError) return []
    if (error instanceof OneError || error instanceof TwoError) return []
  }
}
```

This is surprisingly difficult. I am not really sure what is best approach here...

henri-hulski:
There is also the proposition of @tiagocpontesp:
```js
catch: [
 [MoreSpecificError,[]], 
 [LessSpecificError, []]
]
```
And turn it in a map during resolve with something like:
```js
    const errorMap = new Map(errorKv)

    for (const [errorClass, signal] of errorMap) {
      if (error instanceof errorClass) {
        return signal
      }
      break  // not sure if necessary
    }
```

tiagocpontesp:
I favour only supporting 1 (as an explicit Map).

In my original suggestion I was trying to avoid having the user instantiate the Map himself, but I've since changed my mind. The chained set() syntax looks good.

I think the next question is:
If not `instance of Map`, assume a catch-all tree? Or enforce a one-element Map with the base `Error` key for such scenario? I'm inclined towards the latter, as the former isn't production grade.

christianalfoni:
Okay, added as agreed... will testrun tomorrow morning and then this PR should be done :)

packages/cerebral/src/Controller.js
```diff
@@ -183,21 +183,17 @@ class Controller extends FunctionTree {
...
+        if (signalCatch instanceof Map) {
+          for (let [errorType, signalChain] of signalCatch) {
+            if (error instanceof errorType) {
+              this.runSignal('catch', signalChain, error.payload)
 
-          if (signalChain) {
-            this.runSignal('catch', signalChain, error.payload)
-          } else {
-            throw error
+              return
```
henri-hulski: Why not just use `break` instead of `return`?

christianalfoni: @henri-hulski It has to be return or else it will throw on the line after the IF, which is not what we want when the error has been handled :)

christianalfoni:
Okay, this should now be ready to ship! :-) @Guria

packages/cerebral/src/Controller.js
henri-hulski: @christianalfoni Yeah sorry `break` would only break out of the for...of loop, not the whole function.

christianalfoni:
Wait with this... it turns out babel does not allow extending Error class

christianalfoni:
Okay, so we figured out how this should work. I am just running a final test in our project tomorrow and THEN... 🎉  👍 

For anyone interested JS only lately allows `Error` to be extended, but a tiny wrapper using this Babel plugin fixes it: https://github.com/loganfsmyth/babel-plugin-transform-builtin-extend

christianalfoni:
Okay, can confirm that this update is ready to go :)

docs/packages/firebase.md
```diff
@@ -190,13 +237,18 @@ function transactionFunction(currentData){
...
 
 export default [
-  transaction('foo.bar', transactionFunction), {
-    success: [],
-    error: []
-  }
+  transaction('foo.bar', transactionFunction),
+
+  // Alternatively with explicit paths
+  transaction('foo.bar', transactionFunction)
```
Guria: explicit paths missing here

packages/cerebral-provider-firebase/src/helpers.js
```diff
@@ -83,3 +83,30 @@ export function createUser (user) {
     photoURL: user.photoURL
   }
 }
+
+export function createReturnPromise (returnPromise, path) {
```
Guria: createReturnPromise and convertObjectWithTemplates seems generic and useful enough to be exported from 'cerebral/helpers'

packages/cerebral-provider-firebase/src/remove.js
```diff
@@ -1,13 +1,14 @@
...
   const ref = createRef(path)
-  return new Promise((resolve, reject) => {
-    ref.remove().then(
-      () => resolve({}),
-      (error) => reject({error: error.message})
-    )
-  })
+
+  return ref.remove()
+    .then(() => undefined)
```
Guria: why not `() => {}` ?

Guria: here and in other files

packages/cerebral-provider-http/src/utils.js
Guria: maybe comment here would help make it more explicit

packages/cerebral/src/Module.js
```diff
@@ -14,8 +15,12 @@ class Module {
     module.signals = Object.keys(module.signals || {}).reduce((currentSignals, signalKey) => {
       const signal = module.signals[signalKey]
 
-      currentSignals[signalKey] = (payload) => {
-        controller.runSignal(path.concat(signalKey).join('.'), signal, payload)
+      currentSignals[signalKey] = {
+        signal: signal.signal || signal,
+        catch: signal.catch || controller.catch,
```
Guria: hmm. isn't it expected to be join of signal.catch and controller.catch here?

packages/function-tree-firebase-admin/src/Provider.js
```diff
@@ -24,43 +24,23 @@ function FirebaseAdminProvider (options = {}, customFirebaseInstance) {
...
       },
       remove (path) {
         const ref = firebase.database().ref(path)
 
-        return new Promise((resolve, reject) => {
-          ref.remove()
-            .then(() => resolve())
-            .catch((error) => reject({error: error.message}))
-        })
+        return ref.remove().then(() => undefined)
```
Guria: () => {} ?

packages/cerebral-provider-firebase/src/errors.js
```diff
@@ -0,0 +1,16 @@
+export class FirebaseProviderError extends Error {
```
Guria: So does it require babel plugin to build it?

debugger/src/components/Debugger/Signals/Signal/Action/index.js
```diff
@@ -42,60 +43,90 @@ function renderCode (error) {
...
+            <pre data-line={getLineNumber(error)}>
+              <code
+                ref={(node) => { this.errorElement = node }}
+                className='language-javascript'
+                dangerouslySetInnerHTML={{__html: renderCode(error)}} />
+            </pre>
             {executed}
-            {execution.output ? (
+          </div>
+        ) : null}
```
Guria: condition && (<jsx />)

debugger/src/components/Debugger/Signals/Signal/Action/index.js
```diff
@@ -42,60 +43,90 @@ function renderCode (error) {
...
+                {execution.data.filter(data => data.type !== 'mutation').map((service, index) => <Service service={service} key={index} />)}
               </div>
-            ) : null}
+              {executed}
+              {execution.output ? (
+                <div className='action-actionInput'>
+                  <div className='action-inputLabel'>output:</div>
+                  <div className='action-inputValue'><Inspector value={execution.output} /></div>
+                </div>
+              ) : null}
```
Guria: `condition && (<jsx />)`

debugger/src/components/Debugger/Signals/Signal/index.js
```diff
@@ -110,7 +98,7 @@ export default connect({
     renderAction (action, index) {
       if (action._functionTreePrimitive) {
         return (
-          <div key={index}>
+          <div key={index} onClick={(event) => event.stopPropagation()}>
             <span className='signal-groupName'><strong>{action.type}</strong>{action.name ? ': ' + action.name : null}</span>
```
Guria: `condition && 'string'`

debugger/src/components/Debugger/Signals/Signal/index.js
```diff
@@ -126,20 +114,22 @@ export default connect({
...
+
+      const executedBySignals = (
+        this.props.signal.functionsRun[action.functionIndex] && this.props.signal.functionsRun[action.functionIndex].executedIds.length
+      ) ? this.props.signal.functionsRun[action.functionIndex].executedIds.map((executedId) => this.props.executedBySignals[executedId]) : null
 
       return (
         <Action
           action={action}
-          faded={hasSearchContent === false}
+          faded={hasSearchContent === false || !isExecuted}
```
Guria: please add explicit brackets here

packages/cerebral-provider-firebase/src/remove.js
christianalfoni: Just avoid computation... but can do `{}`, agree it reads better

packages/cerebral/src/Module.js
christianalfoni: Hm, good question... Yeah, I agree, that would be expected

packages/cerebral-provider-firebase/src/errors.js
christianalfoni: Jup, in `.babelrc` there is a plugin that allows extension of Error type

debugger/src/components/Debugger/Signals/Signal/index.js
christianalfoni: Hm, I believe Inferno errors on `undefined`, must be `null`? can double check

debugger/src/components/Debugger/Signals/Signal/index.js
christianalfoni: hm, explicit brackets?

debugger/src/components/Debugger/Signals/Signal/index.js
```diff
@@ -126,20 +114,22 @@ export default connect({
...
       )
 
-      const executedBySignal = (
-        this.props.signal.functionsRun[action.functionIndex] && this.props.signal.functionsRun[action.functionIndex].executedId
-      ) ? this.props.executedBySignals[this.props.signal.functionsRun[action.functionIndex].executedId] : null
+      const isExecuted = Boolean(this.props.signal.functionsRun[action.functionIndex])
+
+      const executedBySignals = (
+        this.props.signal.functionsRun[action.functionIndex] && this.props.signal.functionsRun[action.functionIndex].executedIds.length
+      ) ? this.props.signal.functionsRun[action.functionIndex].executedIds.map((executedId) => this.props.executedBySignals[executedId]) : null
```
Guria: `: []`

debugger/src/components/Debugger/Signals/Signal/index.js
```diff
@@ -126,20 +114,22 @@ export default connect({
...
         <Action
           action={action}
-          faded={hasSearchContent === false}
+          faded={hasSearchContent === false || !isExecuted}
           execution={this.props.signal.functionsRun[action.functionIndex]}
           key={index}
           onMutationClick={this.onMutationClick}
-          onActionClick={this.onActionClick}
-          executed={executedBySignal ? (
+          executed={executedBySignals ? executedBySignals.map((executedBySignal, index) => (
```
Guria: we can remove condition if we ensure array above.

debugger/src/components/Debugger/Signals/Signal/index.js
```diff
@@ -152,7 +142,7 @@ export default connect({
               searchValue={this.props.searchValue}
               mutationClicked={() => {}}
             />
-          ) : null}
+          )) : null}
         >
           {action.outputs ? this.renderOutputs(action) : null}
```
Guria: &&

debugger/src/components/Debugger/Inspector/index.js
```diff
@@ -304,7 +300,11 @@ class Value extends Inferno.Component {
...
       return (
-        <div className={isExactHighlightPath ? 'inspector-highlight' : null}>
+        <div
+          className={isExactHighlightPath ? 'inspector-highlight' : null}
+          onMouseOver={this.onMouseOver}
+          onMouseLeave={this.onMouseLeave}
+        >
           {this.props.propertyKey ? this.props.propertyKey + ': ' : <span />}
           <span onClick={this.onClick}>{isString(value) ? '"' + this.shortenString(value) + '"' : String(value)}</span>
           {hasNext ? ',' : null}
```
Guria: &&

debugger/src/components/Debugger/Signals/Signal/Action/index.js
```diff
@@ -42,60 +43,90 @@ function renderCode (error) {
...
-          </div>
-          <strong>{error.name}</strong> : {error.message}
-          <pre data-line={getLineNumber(error)}>
-            <code className='language-javascript' dangerouslySetInnerHTML={{__html: renderCode(error)}} />
-          </pre>
+        className={error ? 'action action-actionError' : 'action'}
+        onClick={(event) => event.stopPropagation()}
+      >
+        <div className={titleClassname}>
+          {error ? <i className='icon icon-warning' /> : null}
```
Guria: &&

debugger/src/components/Debugger/Inspector/index.js
```diff
@@ -66,15 +66,11 @@ function renderType (value, hasNext, path, propertyKey, highlightPath, modelChan
 class ObjectValue extends Inferno.Component {
   constructor (props, context) {
     super(props)
-    const numberOfKeys = Object.keys(props.value).length
     const isHighlightPath = !!(this.props.highlightPath && isInPath(this.props.highlightPath, this.props.path))
     const preventCollapse = this.props.path.length === 0 && context.options.expanded
```
Guria: explicit brackets please

debugger/src/components/Debugger/Inspector/index.js
```diff
@@ -83,16 +79,12 @@ class ObjectValue extends Inferno.Component {
   componentWillReceiveProps (nextProps) {
     const context = this.context
     const props = nextProps
-    const numberOfKeys = Object.keys(props.value).length
     const isHighlightPath = !!(props.highlightPath && isInPath(props.highlightPath, props.path))
     const preventCollapse = props.path.length === 0 && context.options.expanded
```
Guria: explicit brackets please

debugger/src/components/Debugger/Inspector/index.js
```diff
@@ -280,7 +276,7 @@ class Value extends Inferno.Component {
     this.setState({isEditing: false})
   }
   shortenString (string) {
-    if (string.length > 50) {
+    if (!this.state.forceShowString && string.length > 50) {
```
Guria: explicit brackets please

debugger/src/components/Debugger/Signals/Signal/Action/index.js
```diff
@@ -17,7 +18,7 @@ function getActionName (action) {
 }
 
 function getLineNumber (error) {
-  const variable = error.name === 'TypeError' && error.message.match(/'(.*?)'/) ? error.message.match(/'(.*?)'/)[1] : error.message.split(' ')[0]
+  const variable = error.name === 'TypeError' && String(error.message).match(/'(.*?)'/) ? String(error.message).match(/'(.*?)'/)[1] : String(error.message).split(' ')[0]
```
Guria: explicit brackets please
condition
  ? truthy
  : falsy

debugger/src/modules/Debugger/actions/updateSignal.js
```diff
@@ -13,7 +13,8 @@ function updateSignal ({props, state}) {
   } else {
     state.merge(`${signalPath}.functionsRun.${execution.functionIndex}`, {
       payload: execution.payload,
-      data: execution.data ? [execution.data] : []
+      data: execution.data ? [execution.data] : [],
+      executedIds: []
     })
   }
   if (execution.data && execution.data.type === 'mutation') {
```
Guria: explicit brackets please

packages/cerebral/src/devtools/index.js
```diff
@@ -250,7 +250,8 @@ class Devtools {
             executionId: execution.id,
             name: execution.name,
             staticTree: execution.staticTree,
-            datetime: execution.datetime
+            datetime: execution.datetime,
+            executedBy: payload && payload._execution ? payload._execution : null
```
Guria: explicit brackets please

packages/function-tree/src/devtools/index.js
```diff
@@ -112,7 +112,7 @@ class Devtools {
             name: execution.name,
             staticTree: execution.staticTree,
             datetime: execution.datetime,
-            executedBy: payload._execution ? payload._execution : null
+            executedBy: payload && payload._execution ? payload._execution : null
```
Guria: explicit brackets please

docs/packages/function_tree.md
```diff
@@ -76,12 +85,36 @@ You can now use the debugger from your functions contexts and/or providers:
...
+
+  function createProvider (context) {
+    return {
+      doSomething() {},
+      doSomethingElse() {}
+    }
+  }
+
+  return (context) => {
+    context.myProvider = cachedProvider = cachedProvider || createProvider(context)
```
Guria: explicit brackets please

debugger/src/components/Debugger/Signals/Signal/index.js
Guria: like `(hasSearchContent === false) || !isExecuted`

debugger/src/components/Debugger/Signals/Signal/index.js
Guria: don't know about inferno but react at least doesn't render null, undefined, true, false.

packages/cerebral-provider-firebase/src/errors.js
Guria: can't see it in this pr and dependency defined as well

packages/cerebral-provider-firebase/src/remove.js
Guria: aren't it completely same?

packages/cerebral-provider-firebase/src/errors.js
henri-hulski: @christianalfoni Maybe you forgot to push last changes

packages/function-tree/package.json
```diff
@@ -14,6 +14,7 @@
   },
   "dependencies": {
     "chalk": "^1.1.3",
+    "es6-error": "^4.0.2",
```
henri-hulski: Shouldn't this go in dev-dependencies?

packages/cerebral-provider-firebase/package.json
```diff
@@ -30,6 +29,7 @@
   },
   "devDependencies": {
     "@cerebral/monorepo": "0.0.1",
-    "cerebral": "^2.0.0-b"
+    "cerebral": "^2.0.0-b",
+    "babel-plugin-transform-builtin-extend": "^1.1.2"
```
Guria: move it to cerebral-monorepo

debugger/src/components/Debugger/Signals/Signal/index.js
christianalfoni: done

docs/packages/firebase.md
christianalfoni: done

packages/cerebral-provider-firebase/src/helpers.js
christianalfoni: Hm... this is done several places so is a lot of refactor and documentation needs to be written. Do not consider it blocking, but will make note of it here, so we can do this in later PR... a good thing to improve even after release

packages/cerebral-provider-firebase/src/remove.js
christianalfoni: Not really, `{}` causes a merge to happen, while `undefined` is completely ignored... but `{}` makes you understand the code better so have changed that now :)

christianalfoni: Oh sorry, you mean just run it is a function.. yeah, can do that... just thought it would be implicit

christianalfoni: Damn ES6 arrow functions... so easy to read them wrong. I though you mean returning an object which would cause a merge. Just wanted to be explicit about returning `undefined`

christianalfoni: Somebody might think the `then` is not needed... but it really is

packages/cerebral/src/Module.js
christianalfoni: Fixed... merging `Map` is a little bit complicated, must explicitly spread it to merge it in, in case you wonder why I solved it that way. Anyways, fixed and test added.

packages/function-tree-firebase-admin/src/Provider.js
christianalfoni: Want to be explicit that we want to return undefined, so others do not think that the `then` can be removed

packages/cerebral-provider-firebase/src/errors.js
christianalfoni: Yeah, I forgot to push... weird though, was so sure I did it :)

debugger/src/components/Debugger/Signals/Signal/Action/index.js
christianalfoni: Fixed... though not sure if this is good idea. Cause very often you check a boolean, where you have to do `someBool ? (<jsx/>) : null`. Where `someBool && (<jsx />)` would return `false`, not `null` or `undefined`. Rather be consistent than giving the overhead of figuring out if it is actually valid to shorten it. hm hm... changed it now, but yeah. We should probably decide on this

packages/cerebral-provider-firebase/src/remove.js
Guria: You still don't get it right.
`() => {}` doesn't return anything {} is just a block.
`() => ({})` returns an object
`() => undefined` returns undefined  which can be accidentally become defined in outer/global scope

debugger/src/components/Debugger/Signals/Signal/Action/index.js
christianalfoni: done

debugger/src/components/Debugger/Signals/Signal/index.js
christianalfoni: Sure true and false do not become strings?

christianalfoni: Fixed though

debugger/src/components/Debugger/Signals/Signal/index.js
christianalfoni: fixed

debugger/src/components/Debugger/Signals/Signal/index.js
christianalfoni: fixed

debugger/src/components/Debugger/Signals/Signal/index.js
christianalfoni: fixed

debugger/src/components/Debugger/Inspector/index.js
christianalfoni: Fixed

debugger/src/components/Debugger/Signals/Signal/Action/index.js
christianalfoni: fixed

debugger/src/components/Debugger/Inspector/index.js
christianalfoni: fixed

debugger/src/components/Debugger/Signals/Signal/index.js
henri-hulski: @christianalfoni I think this is related:
https://facebook.github.io/react/docs/conditional-rendering.html#inline-if-with-logical--operator

debugger/src/components/Debugger/Inspector/index.js
christianalfoni: same?

debugger/src/components/Debugger/Inspector/index.js
christianalfoni: fixed

debugger/src/components/Debugger/Signals/Signal/Action/index.js
christianalfoni: Fixed, also fixed that line does not show when line is 0 (has to do with not always finding the line

debugger/src/modules/Debugger/actions/updateSignal.js
christianalfoni: fixed

packages/cerebral-provider-firebase/src/remove.js
Guria: Can use `export const noop = () => {}` from utils for this purpose?
Good point on being explicit, but returning undefined is dangerous in fact.

packages/cerebral/src/devtools/index.js
christianalfoni: Done

packages/function-tree/package.json
Guria: +1

packages/function-tree/src/devtools/index.js
christianalfoni: done

docs/packages/function_tree.md
christianalfoni: done

christianalfoni: Hm.... maybe we should just get prettier in there... so this stuff happens automatically

packages/function-tree/package.json
christianalfoni: Jup, it should :)

packages/cerebral-provider-firebase/package.json
christianalfoni: done :)r

christianalfoni:
Okay, I have made changes, thanks a bunch for going through!

What I realize though is that we should just get PRETTIER in there... spending time on stylistic stuff is waste of time really. They will get "no semicolon" support later... but honestly I do not care. Just want things to be consistent.

Sooo... hope this is ready now :)

edgesoft:
Nice work everyone

packages/cerebral-provider-http/package.json
```diff
@@ -27,6 +27,7 @@
   },
   "devDependencies": {
     "@cerebral/monorepo": "0.0.1",
+    "babel-plugin-transform-builtin-extend": "^1.1.2",
```
henri-hulski: This can be removed now.

christianalfoni:
@henri-hulski No need as it is always an array now :)

packages/cerebral-provider-http/package.json
christianalfoni: Fixed :)

packages/cerebral/src/Controller.js
```diff
@@ -166,7 +168,34 @@ class Controller extends FunctionTree {
...
-    this.runTree(name, signal, payload)
+    this.runTree(name, signal, payload, (error) => {
+      if (error) {
+        const signalPath = error.execution.name.split('.')
+        const signalCatch = signalPath.reduce((currentModule, key, index) => {
+          if (index === signalPath.length - 1) {
+            return currentModule.signals[key].catch
+          }
+
+          return currentModule ? currentModule.modules[key] : undefined
```
Guria: `: null`

christianalfoni: done

packages/cerebral-provider-firebase/src/helpers.js
```diff
@@ -110,3 +110,7 @@ export function convertObjectWithTemplates (obj, resolve) {
     return convertedObject
   }, {})
 }
+
+export function noop () {
+  return function () {}
+}
```
henri-hulski: Cool! Looks the same as `pass` in Python.   ;-)

Guria: Ouch. Smth wrong here. It is definitely not a noop. Why it returns empty function? It should be empty function on its own.

christianalfoni: oh, fudge... I was a bit quick here, should be just plain function yes... working on something else now, will fix it tomorrow as we need a final test anyways (running all the code against large codebase)

packages/cerebral-provider-firebase/src/deleteUser.js
```diff
@@ -1,9 +1,13 @@
...
-  return new Promise((resolve, reject) => {
-    const user = firebase.auth().currentUser
+  const user = firebase.auth().currentUser
 
-    return user.delete().then(() => resolve(), (error) => reject({error}))
-  })
+  return user.delete()
+    .then(noReturnValue)
+    .catch((error) => {
+      throw new FirebaseProviderError(error.message)
```
fmal: Looking at the source of `FirebaseProviderError` class it expects error objet but you're passing `message` directly in several actions. Will that result in message being undefined?

christianalfoni: @fmal Great catch! It should be error... kinda weird, I remember explicitly changing this to use error, going through the actions... hm... must have missed something. Anyways, great! :)

christianalfoni:
Okay... we have run the code on our project, found some error extending issues, but that has been fixed now :) Also fixed error indication in debugger. Now I hope we are ready to go :)

packages/cerebral-provider-firebase/src/onDisconnect.js
```diff
@@ -1,12 +1,13 @@
...
   createRef
 } from './helpers'
+import {FirebaseProviderError} from './errors'
 
 let ref = null
 
 export function setOnDisconnect (path, value) {
   if (ref) {
-    throw new Error('cerebral-provider-firebase: You have already set a disconnect')
+    throw new FirebaseProviderError('You have already a setOnDisconnect')
```
fmal: shouldn't this be `throw new FirebaseProviderError({ message: 'You have already...')` ? 😄

packages/cerebral-provider-http/src/utils.js
christianalfoni: fixed

packages/function-tree/package.json
christianalfoni: Okay, removed this, as not important for the test. Will create tests for errors in firebase and http

packages/cerebral-provider-http/src/utils.js
henri-hulski: Hmm in which way (or where)?

christianalfoni:
Riiiight...  I believe all is taken care of now :)

PR #797, opened 2017-03-24
Move demo to cerebral-provider-forms

edgesoft:


demos/demo/src/modules/user/index.js
```diff
@@ -1,39 +1,45 @@
...
       },
       password: {
         value: '',
         validationRules: ['minLength:5'],
-        validationMessages: ['loginValidationPasswordTooShort'],
+        validationMessages: {
+          'minLength': 'loginValidationPasswordTooShort'
+        },
         isRequired: true,
         requiredMessage: 'loginValidationPasswordRequired'
```
henri-hulski: requiredMessage doesn't exist anymore. You have to check in the view for `isRequired && !hasValue` or similar. 
You must also prevent now showing the error message on empty field when field is required (maybe a bug?). 
I don't use a required message anymore but just change the color of the field to indicate an error.

demos/demo/src/modules/user/index.js
```diff
@@ -1,39 +1,45 @@
...
+    $signIn: {
       email: {
         value: '',
         validationRules: ['isEmail'],
-        validationMessages: ['loginValidationEmailNotValid'],
+        validationMessages: {
+          'isEmail': 'loginValidationEmailNotValid'
+        },
         isRequired: true,
         requiredMessage: 'loginValidationEmailRequired'
```
henri-hulski: See below.

demos/demo/src/modules/user/index.js
edgesoft: @henri-hulski But yeah! Should remove requiedMessage 👍

henri-hulski: Yeah but for an empty field it should not show the error message (eg. "This is not a correct email address"). And it does only if the field is required. This was not the case with cerebral-forms.

henri-hulski: But ok since requiredMessage is gone maybe that makes sense.
You can always override the behavior.

edgesoft: Yes, I will remove the requiredMessage. This is the solution that the writers of demo took 
```js
const showError = field => signIn.showErrors && !field.isValid
```
I will let this be 👍

henri-hulski: I was actually adding cerebral-forms to the demo. ;-)

henri-hulski: So in cerebral-forms the behavior was that if a required field was empty the requiredMessage was shown. This is not the case anymore. 

BTW I was talking about a bug in cerebral-provider-forms not the demo.

edgesoft:
The french translation didn't have the error messages so the english is there.

PR #799, opened 2017-03-24
forms-provider, adjust failedRule behaviour

ArnoBuschmann:
Currently failedRule is still set, even if it is valid. With this code change it gets set to null. As I am a JS beginner, I would appreciate, if someone else could write a test, thanks :)

### DIFF
--- a/packages/cerebral-provider-forms/src/utils/validate.js
+++ b/packages/cerebral-provider-forms/src/utils/validate.js
@@ -33,7 +33,6 @@ export default function validate (form, value, validationRules) {
     } else {
       arg = undefined
     }
-
     
     const isValid = rule(value, form, arg)
 
--- a/packages/cerebral-provider-forms/src/utils/validate.js
+++ b/packages/cerebral-provider-forms/src/utils/validate.js
@@ -33,7 +33,7 @@ export default function validate (form, value, validationRules) {
     } else {
       arg = undefined
     }
-    
+
     const isValid = rule(value, form, arg)
 
     return {
--- a/packages/cerebral-provider-forms/src/utils/validate.js
+++ b/packages/cerebral-provider-forms/src/utils/validate.js
@@ -41,7 +41,7 @@ export default function validate (form, value, validationRules) {
       isValid,
       failedRule: isValid ? null : {
         name: ruleKey,
-        arg: arg
+        arg
       },
       errorMessage: rules._errorMessages[ruleKey] ? rules._errorMessages[ruleKey](value, arg) : null
     }
--- a/packages/cerebral-provider-forms/src/utils/validate.js
+++ b/packages/cerebral-provider-forms/src/utils/validate.js
@@ -34,13 +34,16 @@ export default function validate (form, value, validationRules) {
       arg = undefined
     }
 
+    
+    const isValid = rule(value, form, arg)
+
     return {
-      isValid: rule(value, form, arg),
-      failedRule: {
+      isValid,
+      failedRule: isValid ? null : {
         name: ruleKey,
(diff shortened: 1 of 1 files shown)


ArnoBuschmann:
@henri-hulski arrgh yeah, that line should stay unchanged. It dropped in because I played around in the version that has been built for the node package alread, if that makes sense?

henri-hulski:
I see. :-)

ArnoBuschmann:
@henri-hulski just changed that. Resulted in a second commit, being slightly unsure if this was the correct way how to do it as this is my first PR in general ;)

edgesoft:
@ArnoBuschmann Great to see your first PR. As you can see the build is failing. It's linting. you can check it by running npm run lint and make adjustments. Can you add a test there as well? :)

packages/cerebral-provider-forms/src/utils/validate.js
```diff
@@ -34,11 +34,14 @@ export default function validate (form, value, validationRules) {
...
+    const isValid = rule(value, form, arg)
+
     return {
-      isValid: rule(value, form, arg),
-      failedRule: {
+      isValid,
+      failedRule: isValid ? null : {
         name: ruleKey,
-        arg
+        arg: arg
```
henri-hulski: `arg` is the same as `arg: arg`. 
So you can leave it.

ArnoBuschmann:
@edgesoft thanks for the hints! Tehe, never have been writing tests. But sure, let's get into this as well, I will see, what I can do :)

packages/cerebral-provider-forms/src/utils/validate.js
ArnoBuschmann: Indeed, will change that next, together with the needed lint changes.

packages/cerebral-provider-forms/src/utils/validate.js
```diff
@@ -34,9 +34,12 @@ export default function validate (form, value, validationRules) {
       arg = undefined
     }
 
+
```
Guria: extra empty line

ArnoBuschmann: @Guria thanks, whoa bear with me :)

edgesoft:
@ArnoBuschmann Still linting validate.js:36

edgesoft:
@ArnoBuschmann Sorry! Saw in your first post that you wanted someone else write a test. Will fix!

ArnoBuschmann:
@edgesoft Always great to lean new things, so I will have a look at how to write tests. I'm just not sure yet, if I could provide this one quickly ;)

Guria:
@ArnoBuschmann just add some test cases into https://github.com/ArnoBuschmann/cerebral/blob/e5970de2f6400f0cbc68084cea3f80c1b9642a78/packages/cerebral-provider-forms/src/util.test.js

You just describe what are you expecting to get from function with predefined input and check it with assertions.

ArnoBuschmann:
@Guria thanks a lot for the explanation! Will have a look at it and give it a try, later today or early tomorrow.

packages/cerebral-provider-forms/src/utils/validate.js
```diff
@@ -34,13 +34,16 @@ export default function validate (form, value, validationRules) {
       arg = undefined
     }
 
+    
+    const isValid = rule(value, form, arg)
+
     return {
```
yusufsafak: Hi @ArnoBuschmann, 
If field is valid, I think it is better to return initialValidation.
 `return isValid ? initialValidation : {
  .....
}`

PR #800, opened 2017-03-25
feat(Views): add react fiber support

christianalfoni:
It was actually super easy to create fiber support. It is just a function that might or might not be there, with some intelligent logic to ensure that it does not fail, runs when it is not supposed to etc... 

The way it optimises is that any state mutation that happens after an async function is triggered uses the "requestIdleCallback" thingy. That means you do not have to think about this really. It also means that you might have some sync signal that is not high priority, but this is a balance. I do not see any reason to do low level optimisations by introducing new API. Lets us try fiber first and get this stuff for free :) Will create a video explaining a bit more.

1. Follow this guide to build fiber: https://gist.github.com/duivvv/2ba00d413b8ff7bc1fa5a2e51c61ba43
2. copy react.js and react-dom-fiber.js to your project. 
3. Configure webpack alias to point to these files as:

```js
{
  resolve: {
    alias: {
      'react': path.resolve('someFolder', 'react.js'),
      'react-dom': path.resolve('someFolder', 'react-dom-fiber.js')
    }
  }
}
```

Then it works :)

### DIFF
--- a/packages/cerebral/src/Controller.js
+++ b/packages/cerebral/src/Controller.js
@@ -62,16 +62,16 @@ class Controller extends FunctionTree {
 
     this.on('asyncFunction', (execution, funcDetails) => {
       if (!funcDetails.isParallel) {
-        this.flush()
+        this.flush(false, execution.isAsync)
       }
     })
     this.on('parallelStart', () => this.flush())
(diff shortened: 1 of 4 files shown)


christianalfoni:
Dropping this after feedback that API of fiber will change a lot

PR #802, opened 2017-03-26
fix(function-tree): make websocket work cross browser and node

christianalfoni:


packages/function-tree/package.json
```diff
@@ -15,6 +15,7 @@
   "dependencies": {
     "chalk": "^1.1.3",
     "eventemitter3": "^2.0.2",
+    "universal-websocket-client": "^1.0.1",
     "ws": "^1.1.1"
```
henri-hulski: ws is already dependency of universal-websocket-client (version 2.2.0)

Guria:
Hmm. Why not to pass ws instance to provider config?

christianalfoni:
@Guria Too much configuration in my opinion. This should just work... and it does now :)

packages/function-tree/package.json
christianalfoni: Good catch, fixed!

PR #805, opened 2017-03-27
feat(Devtools): add reconnect

christianalfoni:
Okay, so this has a couple of fixes:

- Now the devtools in Cerebral and Function Tree automatically tries to reconnect, meaning that you do not have to refresh if you open debugger after app is loaded
- I have fixed the FunctionTree constructor. We want it to normally return the `runTree` function, but that brakes inheritance. So now we only export the FunctionTree class, meaning you have to use `new`
- Renamed `runTree` to `run`, meaning you do:
```js
const ft = new FunctionTree()

ft.run([])
```
- Devtools can now be attached as simply as: `Devtools(ft, {})`, or with array of trees
- Updated docs
- Removed an error throwing in TodoMVC... no idea how that got there
- No more Chrome Extension code... cleaned up a lot
- The root sequence will now be shown as the name of execution in debugger:
```js
const ft = new FunctionTree()

ft.run(sequence('doAwesome', [

])
```
**doAwesome** is shown as name of execution in debugger.
- `StopExecution` provider which allows you to conditionally stop any execution by checking some condition on the context or whatever


packages/function-tree/src/index.js
```diff
@@ -191,6 +191,11 @@ export class FunctionTree extends EventEmitter {
     this.runTree.on = this.on.bind(this)
     this.runTree.once = this.once.bind(this)
     this.runTree.off = this.removeListener.bind(this)
+
+    // When not extended
+    if (!returnInstance) {
```
Guria: @christianalfoni have you tested it? looks dangerously for me. Aren't it loose context in this case?

Guria: I see this working, but I am still very unsure on this change. Why factory extracting runTree didn't satisfied you?

christianalfoni: It is a good question indeed. The background for it is to allow an API that does this:

```js
import FunctionTree from 'function-tree'

const execute = new FunctionTree()

execute([])
```

It feels awkward to do:

```js
import FunctionTree from 'function-tree'

const tree = new FunctionTree()

tree.runTree([])
```

Cause you should not have access to the other stuff in there anyways. That said, I totally agree with you, it is not ideal... maybe we should rather change the name of the method:

```js
import FunctionTree from 'function-tree'

const tree = new FunctionTree()

tree.run([]) // Change to "run"
```

??

Guria: But isn't previous factory does exactly same? It just expose only runTree without internal stuff, isn't it?
```js
export default (contextProviders) => {
  return new FunctionTree(contextProviders).runTree
}
```

Guria: behaves as you expect with current published function-tree https://www.webpackbin.com/bins/-KgVhuLwpBVMJB98y5UH

christianalfoni: Ah, I see what you mean... problem is Node, in Node we want this API:

```js
const FunctionTree = require('function-tree').FunctionTree
```

Doing:
```js
const FunctionTree = require('function-tree').default
```

Is just super weird to me... hm... what do others do with this? I have never seen a library where you have to require `.default`

Guria: We can have commonJs entry point instead. Like we are doing already for other entry points

christianalfoni: How does that work? Is it a package.json thing? Or is it how we build?

Guria: just put `index.js` to the root and point to it from "main" in package,json.
I'll see what can be done here on my own.

christianalfoni:
Okay, so I refactored the API to be:

```js
import FunctionTree from 'function-tree'

const ft = new FunctionTree()

ft.run([])
```

This makes it consistent:

```js
const FunctionTree = require('function-tree').FunctionTree
```

The API feels nice and there are no hacks. No extending eventemitter methods to the `runTree` function and what was previously added to this PR.

Guria:
@christianalfoni where this changes could be found?

christianalfoni:
Haha... what is happening to me... I forget to push... pushed now

packages/function-tree/providers.js
```diff
@@ -2,3 +2,4 @@ exports.ContextProvider = require('./lib/providers/Context').default
 exports.ExecutionProvider = require('./lib/providers/Execution').default
 exports.PropsProvider = require('./lib/providers/Props').default
 exports.PathProvider = require('./lib/providers/Path').default
+exports.PathProvider = require('./lib/providers/StopExecution').default
```
Guria: PathProvider ?

packages/function-tree/src/executeTree.js
```diff
@@ -1,12 +1,14 @@
+import {FunctionTreeExecutionError} from './errors'
+
 function isPrimitive (primitive, type) {
   return primitive._functionTreePrimitive && primitive.type === type
 }
 
 /*
   Runs through the tree providing a "next" callback to process next step
-  of execution
+  of execution,
```
Guria: hm :)

packages/function-tree/src/index.js
```diff
@@ -7,6 +7,8 @@ import PropsProvider from './providers/Props'
 import PathProvider from './providers/Path'
 import Path from './Path'
 import {Sequence, Parallel, Primitive} from './primitives'
+import {FunctionTreeExecutionError, FunctionTreeError} from './errors'
+export {FunctionTreeExecutionError, FunctionTreeError} from './errors'
```
Guria: Probably we should leave only reexports in index and move actual implementation to separate file

PR #808, opened 2017-03-29
WIP: Typescript support for function-tree

Ownelek:
Hi guys!
I have decided to start writing types for cerebral, and since function-tree is one of the dependecies for it, I started from this package.
This PR still needs work, because I couldnt figure out all the interfaces by myself, so I am opening PR to ask question about some shapes.
All questions are mentioned in code as comments.
Any feedback is appreciated of course !


packages/@types/function-tree/index.d.ts
```diff
@@ -0,0 +1,32 @@
+import { FunctionTreePrimitive, Parallel, Payload, Sequence } from "./primitives";
+// Dependency on "events" from @types/node, maybe is there a better way ? Or there should be another dependency ?
```
henri-hulski: At the moment we use
```js
import EventEmitter from 'eventemitter3'
```

Guria:
Thanks for PR. We will need a few iterations here.

* First of all, we don't need separate package here. Place d.ts files right into `function-tree` package near source files.
* Add "typings": "lib/index.d.ts" into package.json
* add copying d.ts files from src to lib in prepublish step
* add typescript tests

Feel free to ask any related questions.

Ownelek:
What do you mean by writing typescript tests ? How do I exactly writes tests for declarations files ?

Guria:
usually it just attempt to compile some ts file that imports a lib with definition a tries some valid examples from docs as if you write in your consumer code. It even doesn't supposed to be run or check any asserts. Just check that it compiled without any errors.

Guria:
see redux lib for example:
https://github.com/reactjs/redux/tree/master/test/typescript
https://github.com/reactjs/redux/blob/master/test/typescript.spec.js

Guria:
Cc @christianalfoni would you help with answers on questions listed in comments?

packages/@types/function-tree/Path.d.ts
```diff
@@ -0,0 +1,14 @@
+import { Payload } from "./primitives";
+
+// Is 'path' correct ?
+declare interface IPath {
```
christianalfoni: Not exactly sure what you mean by "correct" :-) But yeah, path is the prop that holds the name of the path

christianalfoni:
This looks great! I tried going through comments, but they require me to evaluate and also look into code that was written a while back... basically, it will take some time and I just have a lot on my plate currently. I will get to it as soon as I can, or maybe someone else takes the opportunity to dive into some source code and suggest solutions :)

gaspard:
This is a great project. I am already using Cerebral with Typescript so I could give a hand here.

@Ownelek Can you handle the comments by @Guria ? If not, I could do this but I must make sure you are not doing it at the same time :-)

Ownelek:
Sure, today I will try to provide tests for that, I already moved files from separate package, just gotta push it.

packages/@types/function-tree/Path.d.ts
Ownelek: By the 'correct' I mean if the type of the prop is correct. In this case 'is string the correct type for path ?'. I will update questions to be more precise.

Guria:
@Ownelek I'll send you invite to join repo so you can push your branch right into repo. It will simplify collaboration

Guria:
Superseeded by #892 
Please chech wip there and leave your feedback. Thanks for you contribution. Hope it was helpful for @reflog

PR #815, opened 2017-04-04
fix(function-tree): remove redux provider

christianalfoni:


saulshanabrook:
Is it still possible to use function-tree with redux? I am looking into the whole redux effects world again for a redux project and would love to use function-tree!

PR #819, opened 2017-04-05
fix(http): fix issue with utils pointing to wrong property

christianalfoni:


henri-hulski:
There is another error in line 105:
```sh
Unhandled Rejection: client/modules/app/app.test.js
  TypeError: Cannot read property 'error' of undefined
    node_modules/cerebral-provider-http/src/utils.js:105:35
```

PR #821, opened 2017-04-05
fix(router): update to new function-tree API

henri-hulski:


henri-hulski:
Hmm... tests are failing because the branch is not in sync with master.
@Guria What do you think? Maybe better close the PR?

Guria:
closing this since the work done on original branch

PR #826, opened 2017-04-07
feat(Devtools): make devtools instance based and flexible

christianalfoni:
Makes function-tree devtools flexible:

```js
const devtools = new Devtools({
  remoteDebugger: 'localhost:8585',
  reconnect: true // Can turn off automatic reconnect
})

devtools.add(functionTreeInstance)
devtools.remove(functionTreeInstance)
devtools.destroy() // removes all instances
```

### DIFF
--- a/packages/function-tree/src/devtools/index.js
+++ b/packages/function-tree/src/devtools/index.js
@@ -1,18 +1,13 @@
-import {FunctionTree} from '../'
 import WebSocket from 'universal-websocket-client'
 import Path from '../Path'
 const VERSION = 'v1'
 
-class DevtoolsClass {
-  constructor (tree, options = {
-    remoteDebugger: null
+export class Devtools {
+  constructor (options = {
+    remoteDebugger: null,
+    reconnect: true
   }) {
-    if (!Array.isArray(tree) && !(tree instanceof FunctionTree)) {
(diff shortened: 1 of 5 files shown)


Guria:
@christianalfoni just merged website updates. need to rebase this one since docs are moved

packages/function-tree/src/devtools/index.js
```diff
@@ -1,33 +1,28 @@
-import {FunctionTree} from '../'
 import WebSocket from 'universal-websocket-client'
 import Path from '../Path'
 const VERSION = 'v1'
 
-class DevtoolsClass {
-  constructor (tree, options = {
-    remoteDebugger: null
+export class Devtools {
```
henri-hulski: Why are you exporting `Devtools` 2 times?

christianalfoni: Due to Node

```js
import Devtools from 'function-tree/devtools'
const Devtools = require('function-tree/devtools').Devtools
```

Or you have to do: 

```js
const Devtools = require('function-tree/devtools').default
```

Which is super weird... so it is just a convention

Guria: Should it managed in entry-point file instead?

Guria: leave only default exports in src/* files, add extra exports in entries

PR #828, opened 2017-04-09
fix(Views): updated PropTypes to come from prop-types to avoid deprec…

fmal:
…ation warnings for React 15.5+

### DIFF
--- a/packages/cerebral/package.json
+++ b/packages/cerebral/package.json
@@ -18,8 +18,7 @@
   "homepage": "http://cerebral.github.io/cerebral-website",
   "dependencies": {
     "eventemitter3": "^2.0.2",
-    "function-tree": "^0.5.0-b",
-    "prop-types": "^15.5.4"
+    "function-tree": "^0.5.0-b"
   },
   "devDependencies": {
     "@cerebral/monorepo": "0.0.1"
--- a/packages/cerebral/package.json
+++ b/packages/cerebral/package.json
@@ -18,7 +18,8 @@
   "homepage": "http://cerebral.github.io/cerebral-website",
   "dependencies": {
     "eventemitter3": "^2.0.2",
-    "function-tree": "^0.5.0-b"
+    "function-tree": "^0.5.0-b",
+    "prop-types": "^15.5.4"
   },
   "devDependencies": {
     "@cerebral/monorepo": "0.0.1"
--- a/packages/cerebral/src/viewFactories/react.js
+++ b/packages/cerebral/src/viewFactories/react.js
@@ -1,9 +1,24 @@
 import React from 'react'
-import PropTypes from 'prop-types'
 import ContainerFactory from './Container'
 import StateContainerFactory from './StateContainer'
 import HocFactory from './Hoc'
 import connectFactory, { decoratorFactory } from './connect'
+
+let PropTypes
+
+if (React.version.slice(0, 4) === '15.5') {
+  try {
+    PropTypes = require('prop-types')
+  } catch (e) {
+    console.error(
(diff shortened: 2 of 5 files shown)


packages/cerebral/package.json
```diff
@@ -18,7 +18,8 @@
   "homepage": "http://cerebral.github.io/cerebral-website",
   "dependencies": {
     "eventemitter3": "^2.0.2",
-    "function-tree": "^0.5.0-b"
+    "function-tree": "^0.5.0-b",
+    "prop-types": "^15.5.4"
```
christianalfoni: This should not be there. React is not a hard dependency to Cerebral, you can also use inferno... and we are keeping options open to use other views as well :)

henri-hulski: Maybe instead it should be mentioned in the the docs for react view.

fmal: @christianalfoni  good catch. Maybe we could handle it in a similar way as here: https://github.com/paypal/glamorous/commit/14970c1bb59d1408d3066d209dafa356b5ce3173?  Try to require it only when React 15.5 detected and adding a note in the docs about `prop-types` installation?

fmal:
@christianalfoni i removed `prop-types` from dependencies and added React version checking. Users running react > 15.5 will continue to use `React.PropTypes` and there is no extra "overhead" for them. For those who installed react 15.5+, it will require `prop-types` package which is now dependency of React. Since npm3 modules installation is flat it will require no extra package installation from the user 😃

packages/cerebral/src/viewFactories/react.js
```diff
@@ -4,7 +4,23 @@ import StateContainerFactory from './StateContainer'
...
 
-export const Container = ContainerFactory(React)
-export const StateContainer = StateContainerFactory(React)
-export const connect = connectFactory(HocFactory(React))
-export const decorator = decoratorFactory(HocFactory(React))
+let PropTypes
+
+if (React.version.slice(0, 4) === '15.5') {
+  try {
+    PropTypes = require('prop-types')
```
Guria: it seems that it wouldn't work for react 16.0+ users, right?

fmal: @guria right, we could adjust the check to be:

```js
const [major, minor] = React.version.split('.');

if ((major === '15' && minor >= 5) || major === '16') {
  ...
}
```
let me try that with `react@16.0.0-alpha.9`

henri-hulski: Maybe 
```js
if ((major === '15' && minor >= 5) || major >= '16')
```
So it doesn't need to be adjusted on every major react release. Only if it breaks.

fmal:
@Guria @henri-hulski ok it pulls in `prop-types` for all React versions above 15.4, i have tried it with  latest react@16 alpha and everything works correctly

packages/cerebral/src/viewFactories/react.js
```diff
@@ -5,8 +5,9 @@ import HocFactory from './Hoc'
 import connectFactory, { decoratorFactory } from './connect'
 
 let PropTypes
+const [major, minor] = React.version.split('.')
 
-if (React.version.slice(0, 4) === '15.5') {
+if ((major === '15' && minor >= 5) || major >= 16) {
```
henri-hulski: Maybe remove the quotes also from `'15'` to be consistent. :)

Guria:
Do we really need prop types at all? Maybe just remove it completely?

fmal:
@guria i guess they're needed at least to define `childContexTypes` and `contextTypes`: https://facebook.github.io/react/docs/context.html

PR #845, opened 2017-04-13
docs(website): improve context, provider, API and router docs

christianalfoni:


docs/api/signal.md
```diff
@@ -0,0 +1,217 @@
...
+
+export default [
+  someAction,
+  someOtherSequence
+]
+```
+
+
+## Parallel
+A parallel runs all actions right after each other, even if they return a promise.
```
henri-hulski: Hmm I thought they run in parallel.

Guria: javascript can't run 2 functions at the same time :)
Just compare this sentence with sequence description.
Though I agree with you, current description is very confusing.

henri-hulski: Maybe there could be a deeper technical description of the differences between running actions in sequence or in parallel.

christianalfoni: Any suggestions on what to write instead? :) Cause this is exactly what happens, but yeah... people might have presumptions on parallel

henri-hulski: So what happens in sequence?

christianalfoni: If an action returns a promise it will wait before moving on :)

christianalfoni: Updated text now :-)

PR #848, opened 2017-04-15
feat(Providers): add provide factory to simplify concept of providers

christianalfoni:
Okay, so I woke up this night with an idea to a different approach making providers concept easier to understand.

So basically I have made some docs updates and added a `provide` helper. It looks like this:

```js
import {Controller, provide} from 'cerebral'
import axios from 'axios'

Controller({
  providers: [
    provide('http', axios)
  ]
})
```

The API makes it very explicit what you do, you "provide axios to Cerebral named http" . This just makes a lot more sense than some magical object in the providers array:

```js
Controller({
  providers: [{
    http: axios
  }]
})
```

This API is flawed cause it is not explicit, what about multiple objects etc.?

There is nothing breaking about this :)

### DIFF
--- a/docs/api/providers.md
+++ b/docs/api/providers.md
@@ -2,7 +2,7 @@
 
 Providers are functions that runs before any action in any signal. Their purpose is to define and sometimes manipulate the context passed into every action. The providers run before every action, meaning that each action has a unique context object.
 
-To add something to the context you create a provider:
+Providers are created with a function:
 
 ```js
 function MyProvider (context, functionDetails, payload, prevPayload) {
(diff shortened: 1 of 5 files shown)


Guria:
I'll repeat a question from previous suggestion. What about operators? Or it just a replacement for context provider?

christianalfoni:
@Guria Just replacement for ContextProvider :) The previous "magical object" did not make much sense. Now "injecting" a tool and using a custom provider is same approach.

```js
Controller({
  providers: [
    provide('axios', axios),
    provide('id', uuid.v4),
    FirebaseProvider(),
    FormsProvider()
  ]
})
```

Just more consistent API and the `provide` factory makes it very explicit that you provide a tool to Cerebral :) Basically a Cerebral version of the ContextProvider.

Issue #771, opened 2017-03-20
tutorial: no styles loaded in final solution on windows


fopsdev:
Just something we shouldn't forget :)
Steps to reproduce:
-Use Windows and recent NodeJS (Didn't test on other OS)
-Download the following BIN as instructed in the tutorial (http://cerebraljs.com/docs/developer_guide/organize.html)
https://www.webpackbin.com/bins/-KfDKApROapTmuhEF5_j
-> When running the solution you will see that styles.css can not be found

fopsdev:
can't we just include styles the old school way (in index.html)?
i'm asking that also because i see quite a few words about style handling in the tutorial which imo could confuse beginners

Guria:
Agreed. Lets simplify this.

fopsdev:
tutorial is not using styles anymore. so its fine :)

PR #854, opened 2017-04-17
docs(tutorial): review

fopsdev:


fopsdev:
only issue now with the tutorial i've found:
https://github.com/cerebral/cerebral/issues/771

docs/developer_guide/compute.md
```diff
@@ -250,7 +250,7 @@ import starsCount from './starsCount'
       showToast(string`The repos have ${starsCount} stars`, 5000, 'success')
     ],
     catch: new Map([
-      [HttpProviderError, showToast(string`Error: ${props`error.message`}`, 5000)]
+      [HttpProviderError, showToast(string`Error: ${props`error.body.message`}`, 5000)]
```
henri-hulski: `error.message` should be correct. 
See https://github.com/cerebral/cerebral/blob/master/packages/cerebral-provider-http/src/HttpProviderError.js

fopsdev: @henri didn't work for me when running it (from the BIN)
what i did:
i changed the repo to a name that doen't exist... then i got: Error: undefined
after changing it to .body.error it was fine...
maybe a different version of http-provider on webpackbin?

fopsdev: you can test it via the bin which is used in the "Router" -part:
https://www.webpackbin.com/bins/-KdBeIDJoRv0PQlF7uWU

but it is wrong anyway...so there is a bin change :)

henri-hulski: @fopsdev Yeah you're right.
No idea why.

PR #864, opened 2017-04-21
fix(Controller): throw correct errors related to state updates

christianalfoni:
There was only one error thrown here, making it super confusing. Now it throws these two errors:

### Wrong state path
Cerebral - The Signal "newTodoTitleChanged" with action "operator.set(state`newTodoTitle.mip`, props`title`)" has an error: The path "newTodoTitle.mip" is invalid, can not update state. Does the path "newTodoTitle" exist?

### Wrong state value
Cerebral - The Signal "newTodoTitleChanged" with action "operator.set(state`newTodoTitle`, props`title.foo`)" has an error: You are passing a non serializable value into the state tree on path "newTodoTitle"

### DIFF
--- a/packages/cerebral/src/Model.js
+++ b/packages/cerebral/src/Model.js
@@ -139,7 +139,7 @@ class Model {
       this.devtools &&
       !isSerializable(value, this.devtools.allowedTypes)
     ) {
-      throwError(`You are passing a non serializable value into the state tree on path ${path.join('.')}`)
+      throwError(`You are passing a non serializable value into the state tree on path "${path.join('.')}"`)
     }
     if (this.devtools) {
       forceSerializable(value)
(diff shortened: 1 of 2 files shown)


packages/cerebral/src/Model.js
```diff
@@ -121,7 +121,7 @@ class Model {
           })
         }
       } else if (!currentState[key]) {
-        throwError(`The path "${path.join('.')}" is invalid, can not update state. Does the path "${path.splice(0, path.length - 1).join('.')}" exist?`)
+        throw new Error(`The path "${path.join('.')}" is invalid, can not update state. Does the path "${path.splice(0, path.length - 1).join('.')}" exist?`)
```
Guria: Why not using `throwError` here?

PR #869, opened 2017-04-22
function-tree-firebase-admin specify authentication, task

edgesoft:
*) Updated some docs
*) Allow to specify authentication true or false
*) add task to provider

### DIFF
--- a/packages/function-tree-firebase-admin/src/QueueHandler.js
+++ b/packages/function-tree-firebase-admin/src/QueueHandler.js
@@ -5,14 +5,15 @@ function authenticate ({firebase, props, path}) {
     .then((decodedToken) => ({uid: decodedToken.uid}))
 }
 
-function createRunTask (task, cb) {
+function noop () {}
+
+function createRunTask (task, shouldAuthenticate, cb) {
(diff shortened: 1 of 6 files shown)


edgesoft:
Will do @henri-hulski

packages/function-tree-firebase-admin/src/QueueHandler.js
```diff
@@ -5,14 +5,15 @@ function authenticate ({firebase, props, path}) {
...
+
+function createRunTask (task, shouldAuthenticate, cb) {
   return (data, progress, resolve, reject) => {
     const _execution = data._execution
 
     delete data._execution
-
     cb(task.specId, [
-      authenticate,
+      shouldAuthenticate ? authenticate : noop,
```
christianalfoni: Instead of having an empty function popping up in the debugger, this should rather be something like:

```js
[].concat(
  shouldAuthenticate ? authenticate : []
).concat(task.tree)
```

edgesoft: @christianalfoni Cool... done

PR #365, opened 2016-10-26
	fix: make npm run setup work on windows (using cross-env)

fopsdev:

PR #870, opened 2017-04-24
Support building cerebral on non-posix environments

reflog:


reflog:
The travis job is failing, I'm not sure why :( I've tested that it's working both on Linux and Windows

henri-hulski:
Try to add cross-env to the dev-dependencies of the monorepo.

reflog:
same result

henri-hulski:
Hmm in the Travis log it says 
```
Error: Command failed: npm run prepublish 
sh: 1: cross-env: not found
```

henri-hulski:
@reflog have a look at #365.
There @fopsdev uses
```js
   "build": "node ../../node_modules/cross-env/bin/cross-env.js BABEL_ENV=production ../../node_modules/.bin/babel src/ --out-dir=lib/ -s --presets es2015",
```

henri-hulski:
So it seems that you have to use the path to the bin directory of the monorepo node_modules to run cross-env.

Guria:
see a bin a section of monorepo package.json

package.json
```diff
@@ -74,5 +62,9 @@
       "**/node_modules/**",
       "**/build/"
     ]
+  },
+  "devDependencies": {
```
Guria: move to general deps for consistency

Guria:
hmm. seems link-parent-bin does the same as my trick with bin section

henri-hulski:
Now there is a different error 
```
> function-tree@0.5.0-b build /home/travis/build/cerebral/cerebral/packages/function-tree
> cross-env BABEL_ENV=production babel src/ --out-dir=lib/ -s
Errored while running BootstrapCommand.execute
Error: Command failed: npm run prepublish 
events.js:160
      throw er; // Unhandled 'error' event
      ^
Error: spawn babel ENOENT
    at exports._errnoException (util.js:1018:11)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:193:32)
    at onErrorNT (internal/child_process.js:367:16)
    at _combinedTickCallback (internal/process/next_tick.js:80:11)
    at process._tickCallback (internal/process/next_tick.js:104:9)
    at Module.runMain (module.js:606:11)
    at run (bootstrap_node.js:393:7)
    at startup (bootstrap_node.js:150:9)
    at bootstrap_node.js:508:3
```

reflog:
@guria I've used link-parent-bin instead of your trick, because on Windows you need to link to cmd files and not the ones without extension

henri-hulski:
So it seems he finds now cross-env.

reflog:
ok, fixed the issues with CI, should be all good now. less hacks than before!

Guria:
My bin trick uses standard nodejs mechanism and must be working in windows as well. You just need add an appropriate entry into it.

I suppose we should keep only one of:
* my bin trick whith monorepo being controlled by lerna
* use link-parent-bin package
* use new lerna [hoisting mechanism](https://github.com/lerna/lerna/blob/master/doc/hoist.md) 

All of them has own pros and cons which I'd like to discuss. I will write up a bit more details on it this weekend I hope.

package.json
```diff
@@ -4,9 +4,10 @@
   "version": "0.0.1",
   "license": "MIT",
   "scripts": {
-    "setup": "lerna bootstrap && link-parent-bin -s true",
-    "setup:packages": "lerna bootstrap --scope \"@(function-tree|cerebral*)\" && link-parent-bin -s true",
-    "setup:babel": "lerna bootstrap --scope \"@(function-tree|cerebral*|babel*)\" && link-parent-bin -s true",
+    "link": "link-parent-bin -s true && ink-parent-bin -s true -c babel",
```
henri-hulski: `ink-parent-bin`?

reflog:
@Guria - unfortunately bin trick does not work on windows. It links the wrong script.
@henri-hulski - gosh, no idea how that slipped by. will fix

package.json
```diff
@@ -4,9 +4,10 @@
   "version": "0.0.1",
   "license": "MIT",
   "scripts": {
-    "setup": "lerna bootstrap",
-    "setup:packages": "lerna bootstrap --scope '@(function-tree|cerebral*)'",
-    "setup:babel": "lerna bootstrap --scope '@(function-tree|cerebral*|babel*)'",
+    "link": "link-parent-bin -s true && ink-parent-bin -s true -c babel",
```
henri-hulski: I think I should comment here. 
`ink-parent-bin` seems to be a typo.

henri-hulski:
@Guria the lerna hoisted approach looks cool. 
Nice that local package.json could stay unchanged. 
Don't know if modules resolution is a problem for us.

Guria:
@reflog can you check one step more then? It seems there is no more need to monorepo being controlled by lerna. Can you try to remove it from lerna.json and from packages dev dependencies?

.travis.yml
```diff
@@ -8,6 +8,8 @@ notifications:
   email: false
 before_script:
 - npm prune
+- link-parent-bin -s true
+- link-parent-bin -s true -c babel
```
Guria: demos and website also must be linked for publishing purposes. and btw why not using npm run link here?

package.json
```diff
@@ -15,19 +16,8 @@
...
-    "nyc": "./node_modules/.bin/nyc",
-    "react-scripts": "./node_modules/.bin/react-scripts",
-    "npm-run-all": "./node_modules/.bin/npm-run-all",
-    "run-p": "./node_modules/.bin/run-p",
-    "run-s": "./node_modules/.bin/run-s",
-    "shx": "./node_modules/.bin/shx",
-    "gh-pages": "./node_modules/.bin/gh-pages",
-    "jest": "./node_modules/.bin/jest"
+    "deploy": "cd website/publisher && npm run ci",
+    "link-parent-bin": "link-parent-bin"
```
Guria: is it extra now?

package.json
```diff
@@ -4,9 +4,10 @@
   "version": "0.0.1",
   "license": "MIT",
   "scripts": {
-    "setup": "lerna bootstrap",
-    "setup:packages": "lerna bootstrap --scope '@(function-tree|cerebral*)'",
-    "setup:babel": "lerna bootstrap --scope '@(function-tree|cerebral*|babel*)'",
+    "link": "link-parent-bin -s true && link-parent-bin -s true -c babel",
+    "setup": "npm run link && lerna bootstrap",
```
Guria: maybe just call `npm run link` as postinstall script?

package.json
reflog: yep. will remove

package.json
reflog: that would make sense!

.travis.yml
reflog: will do

reflog:
@Guria - can you elaborate what exactly do you want me to try here:
"can you check one step more then? It seems there is no more need to monorepo being controlled by lerna. Can you try to remove it from lerna.json and from packages dev dependencies?"

Guria:
@reflog lerna.json has `"."` entry that instructs lerna to treat monorepo as dependency. It was necessary to make bin trick working. And seems there is no need in it anymore. And after that we should remove @cerebral/monorepo from dev deps across packages.

package.json
```diff
@@ -4,9 +4,10 @@
   "version": "0.0.1",
   "license": "MIT",
   "scripts": {
-    "setup": "lerna bootstrap",
-    "setup:packages": "lerna bootstrap --scope '@(function-tree|cerebral*)'",
-    "setup:babel": "lerna bootstrap --scope '@(function-tree|cerebral*|babel*)'",
+    "link": "link-parent-bin -s true && link-parent-bin -s true -c babel && link-parent-bin -s true -c demos && link-parent-bin -s true -c website",
+    "setup": "npm run link && lerna bootstrap",
```
Guria: I would suggest to remove link from here, since now we have it in postinstall

reflog:
removing the '.' from package didn't work. broke the build :(

Guria:
Just removing dot is not enough. I have mentioned that extra dependency on monorepo across packages should be removed.

reflog:
all done. works like a charm

PR #872, opened 2017-04-27
New monorepo module - shortcuts

reflog:


Guria:
Does it worth to be a module package that must be mantained later?
I thinks it fits better to some sort of article or how-to.

reflog:
@Guria - I've initially created it as my personal npm module, but @christianalfoni suggested pulling it into the monorepo...

gaspard:
Hey @reflog this is really nice and thanks for the shortway discovery :-)

I agree with @Guria that we need to discuss what gets added to monorepo and what will become maintained, etc.

I think the best solution for such a small project is to create a tiny tutorial or documentation explaining how things work: there are some interesting things for documentation here:

- use case for module declared as function
- module.controller.on ( 'initialized', ...
- how to handle shortcuts

The added value is greater as documentation I think....

christianalfoni:
@Guria @gaspard @reflog Thanks for feedback :) Yeah, this is something we just need to decide on.  On one side we do want to make it easy for people to power their app with new batteries. This was the intention of modules, make it easy to share common functionality. On the other hand we also want people to use the same concepts internally in their own apps, and this is a good example of how to achieve it, making it suitable for an article.

Maintainability is also of course a question here. If we bring in too many packages to monorepo it will be more to handle.

Community is also something to consider. I want to attract contributors to the Cerebral monorepo and having developed packages in the same repo brings them closer to the other packages and codebase... making the threshold of further contributions lower.

I think the concept of this module is "common functionality", so it should be a published module for people to use. Writing an article about is also valuable, but does not have to be one or the other, can do both :-) And for now I do think it should be in the monorepo for community reasons. If a lot of modules are created we should of course not just blindly bring in everything... but we are not in that situation yet, and this is such a simple power-up to your app and has low degree of needed maintainability :)

gaspard:
@christianalfoni I agree with you and understand your point.

In this case, I would just require that all modules added to monorepo have tests, at the very least something that validates that the package runs without throwing errors. This way we can spot breaking problems early and it helps on the long run.

Having tests for such a small module is also of good teaching value for other contributions.

:-)

christianalfoni:
Yeah, good point @gaspard!  @reflog , would you mind creating a couple of tests? Just copy test setup from `forms` for example :) I believe you can dispatch a keypress event manually.

reflog:
Sure thing!

Guria:
Don't forget to add package into bootstrap command inside travis.yml

packages/cerebral-module-shortcuts/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2016 Cerebral
```
henri-hulski: Could be 2017 :)

reflog:
all done

Guria:
Should be rebeased after #885

PR #878, opened 2017-05-03
test(forms): add more tests for code coverage and some changes

yusufsafak:
Hi @christianalfoni ,
- use runCompute instead of controller in form.test.js
- catch errors using new cerebral error handling 
- fix field error message, if the field is valid. It should return initialValidation without any error message.

By the way why initialValidation has failedRule key?

### DIFF
--- a/packages/cerebral-provider-forms/src/operators/resetForm.js
+++ b/packages/cerebral-provider-forms/src/operators/resetForm.js
@@ -3,7 +3,7 @@ import resetFormHelper from '../helpers/resetForm'
 function resetFormFactory (formPath) {
   function resetForm ({state, resolve}) {
     if (!resolve.isTag(formPath, 'state')) {
-      throw new Error('Cerebral Forms - resetForm factory requires a STATE TAG')
+      throw new Error('Cerebral-Provider-Forms operator.resetForm: You have to use the STATE TAG as an argument')
     }
 
     const path = resolve.path(formPath)
--- a/packages/cerebral-provider-forms/src/operators/resetForm.js
+++ b/packages/cerebral-provider-forms/src/operators/resetForm.js
@@ -3,7 +3,7 @@ import resetFormHelper from '../helpers/resetForm'
 function resetFormFactory (formPath) {
   function resetForm ({state, resolve}) {
     if (!resolve.isTag(formPath, 'state')) {
-      throw new Error('Cerebral-Provider-Forms operator.resetForm: You have to use the STATE TAG as an argument')
+      throw new Error('Cerebral Forms - operator.resetForm: You have to use the STATE TAG as an argument')
     }
 
     const path = resolve.path(formPath)
--- a/packages/cerebral-provider-forms/src/operators/resetForm.test.js
+++ b/packages/cerebral-provider-forms/src/operators/resetForm.test.js
@@ -107,7 +107,7 @@ describe('resetForm', () => {
               ({props}) => {
                 errorCount++
                 assert.equal(props.error.name, 'Error')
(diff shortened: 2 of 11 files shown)


packages/cerebral-provider-forms/src/operators/resetForm.js
```diff
@@ -3,7 +3,7 @@ import resetFormHelper from '../helpers/resetForm'
 function resetFormFactory (formPath) {
   function resetForm ({state, resolve}) {
     if (!resolve.isTag(formPath, 'state')) {
-      throw new Error('Cerebral Forms - resetForm factory requires a STATE TAG')
+      throw new Error('Cerebral-Provider-Forms operator.resetForm: You have to use the STATE TAG as an argument')
```
edgesoft: #882

packages/cerebral-provider-forms/src/operators/resetForm.test.js
```diff
@@ -93,4 +93,35 @@ describe('resetForm', () => {
...
+        test: {
+          signal: [
+            resetForm('form')
+          ],
+          catch: new Map([
+            [Error, [
+              ({props}) => {
+                errorCount++
+                assert.equal(props.error.name, 'Error')
+                assert.equal(props.error.message, 'Cerebral-Provider-Forms operator.resetForm: You have to use the STATE TAG as an argument')
```
edgesoft: #882

packages/cerebral-provider-forms/src/operators/setField.js
```diff
@@ -1,7 +1,7 @@
 function setFieldFactory (fieldPath, fieldValue) {
   function setField ({state, resolve}) {
     if (!resolve.isTag(fieldPath, 'state')) {
-      throw new Error('cerebral-provider-forms operator.setField: You have to use the STATE TAG as first argument')
+      throw new Error('Cerebral-Provider-Forms operator.setField: You have to use the STATE TAG as first argument')
```
edgesoft: #882

packages/cerebral-provider-forms/src/operators/setField.test.js
```diff
@@ -29,4 +29,35 @@ describe('setField', () => {
...
+        test: {
+          signal: [
+            setField('form.name', 'foo')
+          ],
+          catch: new Map([
+            [Error, [
+              ({props}) => {
+                errorCount++
+                assert.equal(props.error.name, 'Error')
+                assert.equal(props.error.message, 'Cerebral-Provider-Forms operator.setField: You have to use the STATE TAG as first argument')
```
edgesoft: #882

christianalfoni:
This looks great @yusufsafak ! Just need to rename the messages to fit the new package naming for release :-)

yusufsafak:
thanks @christianalfoni 
why initialValidation has failedRule key? Do we need it?

christianalfoni:
@yusufsafak You have a pointer in code? I believe it was added to avoid checking for existence of key as well, when using it in component. So you would not get like "can not access prop of undefined". Have to look at code though

yusufsafak:
https://github.com/cerebral/cerebral/blob/master/packages/cerebral-provider-forms/src/utils/validate.js#L6

@christianalfoni I understand what you meant. But if someone wants to use field.failedRule.name without controlling isValid value or  field.failedRule, it will also throw an error.

I think, we should always check first isValid value of the field when using validation in component. When the field is valid, failedRule and errorMessage are useless.

If you want to keep initial values of failedRule and errorMessage as null, please let me know. I have to change this pull request a little bit.

christianalfoni:
Good point! Yeah, I think it is okay to have it removed then. As you say... to verify if a failedRule should be used you check if it is invalid first :-)

PR #879, opened 2017-05-03
feat(controller): add getModule function

yusufsafak:


Guria:
Can you add this method to api docs as well?

yusufsafak:
Hi @Guria , 

I checked it again and I think, it is not a good idea to expose getModule method :( you can set unserializable value to module state directly! 
I can create a new PR to change it to private method like _getModule or private func. 

The following test is working.

```
it('should expose method to getModule using path', () => {
    const controller = new Controller({
      modules: {
        foo: {
          modules: {
            bar: {
              state: {
                bar: 'baz'
              }
            }
          },
          signals: {}
        }
      }
    })

    const barModule = controller.getModule('foo.bar')
    class Test {}
    barModule.state.bar = new Test()
    assert.ok(controller.getState('foo.bar.bar') instanceof Test) 
  })
```

Guria:
Seems a bug for me. I suppose module.state must be used only as initial state and must not be kept as refference. cc @christianalfoni

christianalfoni:
Hm, yeah... I am not sure why we have this method. I do not think there was any reason... it was just added because we added `addModule`. I think we should rather remove it.... do no see any reason to have it?

Guria:
It was introduced by @yusufsafak in this PR to get rid of code duplication. See Changes tab.
Just removing this method doesn't solves the problem. Because module can access this.state if it written as class.

christianalfoni:
Hm, we can not prevent user from doing things completely wrong. Even `controller.module.state` can be manipulated.

So I think this is rather a docs and API surface thing.

Related to this PR I suggest we change `getModule` to be just a private function that takes needed arguments.

I have never heard of anyone misunderstanding the API to such a degree that they do wrongful mutations and we can not prevent it anyways. In any lib you can fudge up things :)

So in my opinion I would be happy with moving `getModule` to private function. If we get an actual use case where someone misunderstands the API to such a degree that they mutate state outside of the state API, then we can do something about it :-)

Issue #859, opened 2017-04-18
operator when: throw errors for wrong input


Guria:
```
// should throw:
when(false)
when(true)
```

reflog:
when(true) already seems to throw an error:
```javascript
 assert.throws(() => {
      const controller = Controller({
        signals: {
          test: [
            when(true, {
              true: [],
              false: []
            })
          ]
        }
      })
      controller.getSignal('test')()
    });
```

or was this not the case you are describing?

PR #880, opened 2017-05-04
fix(Operators): operator when: throw errors for wrong input

reflog:
ISSUES CLOSED: #859

### DIFF
--- a/packages/cerebral/package.json
+++ b/packages/cerebral/package.json
@@ -21,7 +21,7 @@
     "function-tree": "^0.5.0-b"
   },
   "scripts": {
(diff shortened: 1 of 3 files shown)


Guria:
Great addition but doesn't covers my original case. It wasn't about path is not provided.
```js
[
  when(someTag === someTag). // wrong usage. must throw even if path true/false is defined
  {true: [], false: []} 
]
```

reflog:
@Guria - so what exactly do you want to check? Right now stuff like `when(true)` already throws an error...

Guria:
But what about `when(false), {true:[], false:[]}`? I think it eill fail due to my observations.

reflog:
ok, i think this time I covered both cases

Issue #841, opened 2017-04-12
attempt to get catch error chain fails for signal ran with Controller.runSignal


Guria:
```
Uncaught TypeError: Cannot read property 'catch' of undefined
    at http://localhost:3000/static/js/bundle.js:35235:49
    at Array.reduce (native)
    at http://localhost:3000/static/js/bundle.js:35233:41
    at http://localhost:3000/static/js/bundle.js:36033:12
    at FunctionTreeExecution.runFunction (http://localhost:3000/static/js/bundle.js:35873:17)
    at runBranch (http://localhost:3000/static/js/bundle.js:36474:18)
    at runNextItem (http://localhost:3000/static/js/bundle.js:36412:8)
    at http://localhost:3000/static/js/bundle.js:36429:12
    at FunctionTreeExecution.runFunction (http://localhost:3000/static/js/bundle.js:35932:10)
    at runBranch (http://localhost:3000/static/js/bundle.js:36474:18)
```

Guria:
steps to reproduce:
```js
controller.runSignal('foo', [() => throw new Error()])
```

PR #881, opened 2017-05-04
fix(Controller): catch error chain for signal ran with Controller.runSignal

reflog:
ISSUES CLOSED: #841


gaspard:
This commit was integrated into the router branch as we had a similar fix there.

Issue #732, opened 2017-03-13
Move to prettier


edgesoft:
What are the pros and cons of moving cerebral and all other packages to prettier.

https://github.com/prettier/prettier

If no obvious cons I suggest that we move to prettier.

christianalfoni:
I think @Guria has the strongest voice here, being the maintainer of repos :) I think prettier looks really awesome and as I understand it would be an addition to linting? Cause linters picks up stuff that prettier does not?

henri-hulski:
I really enjoy the semicolon free approach of standard. Would really missing it. 
What's nice about prettier is that he just takes your code and normalize it. I also like that he introduces a max. line length and automatically reformat to it.

So maybe a combination of prettier and standard would be fine which keeps the semicolon free approach.

gaspard:
@henri-hulski Prettier has a setting to not add semicolon. I think this tool is a good idea (but it needs to be tested first).

I really like consistency regarding source code and would love something like "prettier" for my own needs (http://new.lucidogen.io/jessy/)...

henri-hulski:
@gaspard It does???

henri-hulski:
https://github.com/prettier/prettier/issues/12

gaspard:
@henri-hulski Oh, sorry. I got confused with the "trailing comma" option. You are right: semicolons everywhere... Back to 1960.

sheerun:
@christianalfoni @Guria I've created https://github.com/sheerun/prettier-standard if that helps you

edgesoft:
#923

PR #883, opened 2017-05-05
chore: move to prettier, use lint-stage to auto fix files before commit

reflog:
ISSUES CLOSED: #732

### DIFF
--- a/babel/babel-plugin-cerebral-optimize-tags/index.js
+++ b/babel/babel-plugin-cerebral-optimize-tags/index.js
@@ -12,7 +12,10 @@ function isAllowedImport (importName) {
 export default function ({types: t}) {
   const optionsMap = {
     state: t.objectExpression([
-      t.objectProperty(t.stringLiteral('isStateDependency'), t.booleanLiteral(true))
+      t.objectProperty(
+        t.stringLiteral('isStateDependency'),
+        t.booleanLiteral(true)
+      )
     ]),
     string: t.objectExpression([
       t.objectProperty(t.stringLiteral('hasValue'), t.booleanLiteral(false))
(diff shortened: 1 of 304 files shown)


reflog:
I know this commit is noisy, there are differences in style between prettier and what we had, but I think it'll be very good in the long run. What are your thoughts? Also, lint-staged approach is better than pre-test, since it avoids bad commits.

henri-hulski:
@reflog Could you split the PR in 2 commits?
One with your changes and one with the refactor by prettier?
Like this it's impossible to review.

reflog:
done.

henri-hulski:
Don't think we need prettier-standard as prettier supports now semi option.
When using prettier directly we will have a more recent version.
We should also discuss which options to use.
I'm using
```json
{
  "trailingComma": "es5",
  "singleQuote": true,
  "bracketSpacing": false,
  "semi": false
}
```
The last 3 makes it similar to standard but the first is not in standard.
But it makes diffs more readable (no lines which only changes trailing commas)
and also easier to copy/paste parts of arrays/objects (no need to add/remove trailing commas).

reflog:
@henri-hulski - didn't know prettier now has semi option... good! about 'recent' issue - it's not a problem, prettier-standard uses peerDependency, so we can add prettier of any version ourselves.
Regarding specific style - that's for the core team to decide I suppose, I'm good with any solution, as long as it's enforced by the tooling :)

Guria:
Good point. Agree with you.

reflog:
👌

Guria:
@reflog I will close it now since it would make no sense to rebase it over comming changes.

Btw, would be great to have separate commit for configuration changes and another commit for applied changes from prettier.

Also make sure to check https://github.com/feross/standard/issues/811 before next attempt please.

PR #887, opened 2017-05-07
feat(debugger): add update notification

christianalfoni:
Okay, so this is a simple feature that gives debugger a popup when a new version is released.

This is a fallback to self-update, to avoid it blocking the release of Cerebral.

### DIFF
--- a/debugger/connector/index.js
+++ b/debugger/connector/index.js
@@ -10,6 +10,12 @@ const connector = {
       data: payload
     })
   },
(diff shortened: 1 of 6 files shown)


Guria:
Where are with this now? Is it still valid?

christianalfoni:
@Guria Not valid for release, but might use it later to notify about updates to Cerebral package itself through the debugger. Let me close it, can use it as reference later :)

PR #892, opened 2017-05-09
Complete Typescript support

### LINKED ISSUE #796: Typescript support ?
Hi guys,
As far as i can see, there is no types nor DefinitelyTyped typings for cerebral.
Is it planned ?

reflog:
I'd like to create a complete typescript support for Cerebral.
I'll go through all the packages, in each create typings folder and reference it from package.json and will add a typings test.
Will close #796 and hopefully supersede https://github.com/cerebral/cerebral/pull/808


.gitignore
```diff
@@ -12,3 +12,4 @@ website/builder/docs
 **/yarn.lock
 **/yarn-error.log
 !/yarn.lock
+.vscode/launch.json
```
Guria: it's better add environment specific ignore into your local gitignore rather than in projects repos.

packages/cerebral/typings/__test__/examples/code1.ts
```diff
@@ -0,0 +1,13 @@
+/// <reference path="../../index.d.ts" />
```
Guria: do we really need referencing using this syntax?  It looks very odd to me

packages/cerebral/typings/index.d.ts
```diff
@@ -0,0 +1,109 @@
+// <reference types="node" />
+// <reference types="function-tree" />
+
+
+declare module 'cerebral/devtools' {
```
Guria: I don't think we should use ambient module definitions.

Guria: see eg redux definitions. it doesn't use declare.

reflog:
All great points, will take care of it

packages/cerebral/typings/index.d.ts
reflog: Are you sure about this? Redux is an easier case - they just have function and no sub-modules... Maybe I'm wrong, but I can't find an example project that has a structure more similar to Cerebral's

reflog: Nope. Can't find a way to handle `'cerebral/devtools'` properly. Suggestions welcome.

Guria: you just place an d.ts file for each entry point with the same name. typescript compiler itself will resolve it

reflog: That's not that simple. Unless I put index.d.ts in our root folder (/packages/cerebral/), TS compiler doesn't recognize it as 'cerebral' module and doesn't allow `import { Controller } from "cerebral";`. 

So if I move index.d.ts to src folder (in order to work with /devtools/ subfolder) the compilation breaks.

Also, another problem will come with inferno and react 'submodules' inside viewFactories folder, since they are not folders, so I'm not sure how to place d.ts files there.

Help! :)

.cz-config.js
```diff
@@ -12,6 +12,7 @@ module.exports = {
     {value: 'test', name: 'test:     Adding missing tests'},
     {value: 'chore', name: 'chore:    Changes to the build process or auxiliary tools\n            and libraries such as documentation generation'},
     {value: 'revert', name: 'revert:   Revert to a commit'},
+    {value: 'ts', name: 'ts:   Typescript/Typings retlated changes'},
```
Guria: retlated -> related

.gitignore
```diff
@@ -11,4 +11,4 @@ website/builder/docs
 *darwin-x64
 **/yarn.lock
 **/yarn-error.log
-!/yarn.lock
+!/yarn.lock
```
Guria: please keep newlines at the end of all files

packages/cerebral/devtools.d.ts
```diff
@@ -0,0 +1,13 @@
+export interface DevOptions {
```
Guria: Do we need to export this interface? Take a note that all exported interfaces becomes part of public API and it will be covered by semver. Eg,, renaming exported interface or any other breaking chage in it would require major bump.

packages/cerebral/index.d.ts
```diff
@@ -0,0 +1,35 @@
+// <reference types="node" />
```
Guria: what this stands for?

packages/cerebral/react.d.ts
```diff
@@ -0,0 +1,9 @@
+export class StateContainer {
+
+}
+
+export class Container {
```
Guria: shouldn't it be extended from React.Component?

packages/cerebral/tags.d.ts
```diff
@@ -0,0 +1,6 @@
+import { IPath } from "../function-tree/index";
```
Guria: from 'function-tree'

packages/cerebral/typings/__test__/examples/code1.ts
```diff
@@ -0,0 +1,12 @@
+// if i try importing from 'cerebral' and 'cerebral/devtools' , it fails to compile
```
Guria: let's remove this comment.
it's expected behavior. js tests also requires using relative pathes.
btw, why this file burried so deep? can we collocate tests near d.ts files?

packages/cerebral/typings/__test__/examples/code1.ts
```diff
@@ -0,0 +1,12 @@
...
+import { Controller } from "../../../index";
```
Guria: index could be ommited

packages/cerebral/typings/__test__/examples/code1.ts
```diff
@@ -0,0 +1,12 @@
...
+import DevTools from "../../../devtools";
+
+const controller = new Controller({
+    devtools: new DevTools({
+        remoteDebugger: "1.1.1.1"
+    })
+});
+
+
+controller.getSignal("ASD");
```
Guria: should we have a wider set of API used?

packages/forms/index.d.ts
```diff
@@ -0,0 +1,6 @@
+import { IPath } from "../function-tree/index";
```
Guria: from 'function-tree'

packages/forms/operators.d.ts
```diff
@@ -0,0 +1,5 @@
+import { IPath } from "../function-tree/index";
```
Guria: from 'function-tree'

packages/function-tree/index.d.ts
```diff
@@ -0,0 +1,141 @@
+/// <reference types="node" />
+
+export interface IPayload {
```
Guria: Lets make sure we are exporting here only needed interfaces.

Guria: Let also avoid using I prefix for interface names as well.

packages/function-tree/index.d.ts
```diff
@@ -0,0 +1,141 @@
...
+
+export type Provider = (context: any, funcDetails: FunctionTreePrimitive, payload: IPayload, next: IPayload) => any
+
+export function ReduxProvider(store: any): Provider
+export function PropsProvider(): Provider
+export function PathProvider(): Provider
+export function ExecutionProvider(execution: any, abort: Abort): Provider
+export function ContextProvider(extendedContext: any): Provider
+
+export class FunctionTree extends NodeJS.EventEmitter {
```
Guria: we extend eventemitter3. which seems to [have own typings](https://github.com/primus/eventemitter3/blob/master/index.d.ts) why we declare here NodeJS.EventEmitter?

packages/http/index.d.ts
```diff
@@ -0,0 +1,7 @@
+import { Provider } from "../function-tree/index"
```
Guria: from 'function-tree'

reflog:
ok, addressed all comments except the location of the typescript tests. please propose a structure, because I need both the test runner and a folder with .ts files that will execute the tests and I don't want to litter the codebase too much..

Guria:
sorry. just found pending review comments

packages/cerebral/tags.d.ts
Guria: why this left?

packages/cerebral/package.json
```diff
@@ -35,5 +38,10 @@
       "**/*.test.js",
       "**/testHelper.js"
     ]
+  },
+  "devDependencies": {
+    "@types/node": "^7.0.18",
+    "typescript": "^2.3.2",
```
Guria: all dev dependencies should be defined only in monorepo package.json

Guria:
about tests location. let put them in one folder deep, eg, `ts`:
```
ts/
  typescript.spec.js
  some-file-to-compile.ts
  another-file-to-compile.ts
lib/
src/
index.js
index.d.ts
operators.js
operators.d.ts
```

henri-hulski:
@reflog you should not merge master but rebase on master. 
Like this this the PR cannot be merged back. 
If you need help I can give you detailed instructions.

reflog:
crap! i thought that's what I did! I ran `git rebase origin/master` in my branch,...

henri-hulski:
Strange. That sounds correct. I'm using `git rebase master` on the branch after pulling to master, but this looks even faster. 
But here definitely happens a merge.

henri-hulski:
After you need to do `git push --force`.

Guria:
@reflog will you resolve rest of comments?

packages/cerebral/tags.d.ts
Guria: Ping

reflog:
@Guria - I've the commit up.

packages/function-tree/package.json
```diff
@@ -46,8 +46,5 @@
       "**/*.test.js",
       "**/testHelper.js"
     ]
-  },
```
gaspard: The package `mock-socket` is required to run tests in function-tree. Please leave the devDependencies where they are needed. Most of them are missing in other packages and this is wrong and we will need to fix it. I would like to setup a rule where dependencies not in any package's `devDependencies` or `dependencies` will  get removed from main package.json at some point.

packages/cerebral/package.json
```diff
@@ -35,8 +35,5 @@
       "**/*.test.js",
       "**/testHelper.js"
     ]
-  },
-  "devDependencies": {
-    "mock-socket": "^6.0.4"
```
gaspard: `mock-socket` does not seem to be used in cerebral testing so this is good cleanup.

packages/function-tree/package.json
reflog: done

gaspard:
What is blocking this PR ? Can we move forward ? It would really be good to clear pull requests so that we can finish transition to "no lerna" thing and work on release and build issues.

PR #898, opened 2017-05-14
feat(operator): Add increment operator.

sladiri:
I thought that a generic increment operator might be a handy addition. :)


reflog:
Looks like a nice addition!

sladiri:
I added two more tests. :)

Guria:
Please add an section into website api documentation

sladiri:
I added the section now.

PR #900, opened 2017-05-16
Refactor devtools

yusufsafak:
Hi all,

PR includes:  
- create new DevtoolsBase class (get rid of code duplication)
- Cerebral Devtools and Function-tree Devtools extending from DevtoolsBase  
- new devtools tests using mock-socket

function-tree devtools
- change destroy method name to removeAll because it removes all trees but not destroy devtools
- fix remove tree should remove provider
- fix removeAll (destroy) .first copy trees then remove them safely
- fix reconnect. Before it listened onclose method two times.

cerebral devtools
- remove doReconnect from options
- fix init memory leak :) only call watchExecution if it is the first time.
- use safeStringify method for cerebral devtools.

I pushed also some fixes for debugger.  https://github.com/cerebral/cerebral-debugger/pull/3

I tested with my cerebral project. It seems OK. @christianalfoni can you please test it? I don't have any function-tree project yet :) 

By the way why do we have two different debuggers at webpage and where is the function-tree debugger source code?

### DIFF
--- a/packages/cerebral/package.json
+++ b/packages/cerebral/package.json
@@ -35,5 +35,8 @@
       "**/*.test.js",
       "**/testHelper.js"
     ]
+  },
+  "devDependencies": {
+    "mock-socket": "^6.0.4"
   }
 }
--- a/packages/cerebral/src/devtools/index.js
+++ b/packages/cerebral/src/devtools/index.js
@@ -160,10 +160,11 @@ export class Devtools extends DevtoolsBase {
 
     this.isResettingDebugger = true
     this.sendMessage(message)
-    this.sendBulkMessage(this.backlog, 'c')
+    if (this.backlog.length) {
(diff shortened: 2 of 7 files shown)


Guria:
Man, you rock. Thanks for another great contribution.

christianalfoni:
Haha, this is really amazing stuff :)

Yes, let me checkout this branch and try it out on our project, also using function-tree. Waiting for debugger test deploy so I can try that at the same time :)

yusufsafak:
You are welcome !!
I have just tried function-tree on server. It is really awesome :)
I also used this PR and new debugger repo. It looks good at least at my simple configuration :)

> why do we have two different debuggers at webpage and where is the function-tree debugger source code?

http://cerebraljs.com/docs/addons/function_tree.html#devtools should be replaced with the new debugger. Or am I missing something?

PR #902, opened 2017-05-17
First steps to new repo structure

Guria:
Sorry guys for one more repo restructuring, but I am failed with trying make lerna fit our needs.
I have decided to organize repo according to great writeup https://github.com/boennemann/alle
You can see here only first steps, that allowed me to make first beta release finally.
I will add a checklist on what else is needed to complete transition.


Guria:
That is why I marked it as WIP and said that it is only a first step.
My concerns was not only about publishing process.

You may take a note that this approach doesn't involve long cross linking process that was happen with lerna bootstrap. Here we have native module resolution mechanism.

Script to run tests was removed only because it was used lerna to run and coveralls is not. Tests itself not gone. Of course tests must be restored.

I will prepare a list on what must be restored and improved. And I hope you guys could help with it. Really I see a big improvements we can get with this approach.

Guria:
closing this for now until we finish with others prs

PR #906, opened 2017-05-18
Update install instructions and some minor fixes

henri-hulski:


README.md
```diff
@@ -21,13 +21,15 @@ A state controller with its own debugger
 ## Try out Cerebral 2
 To try out Cerebral 2 you can install any package with:
 
-`npm install {cerebralPackage}@next --save --save-exact`
+`npm install {cerebralPackage}@beta --save`
+
+For packages which are scoped with `@cerebral/` `@beta` can be omitted.
```
Guria: It is just a side effect of first publish. Lets omit this note.

README.md
```diff
@@ -21,13 +21,15 @@ A state controller with its own debugger
...
 
 This will install the latest version of that package. New versions are automatically deployed whenever pull requests are merged into master.
 
 When you want to update your packages you **have to update all used packages in one go**, for example:
```
Guria: Hope to get rid of this and have separate versioning soon.

packages/firebase-admin/README.md
```diff
@@ -3,11 +3,11 @@
 ## install
 **NPM**
 
-`npm install @cerebral/firebase-admin@next --save --save-exact`
+`npm install @cerebral/firebase-admin --save`
```
Guria: Lets keep beta tag here and for other packages

README.md
henri-hulski: @Guria Why you want to keep beta tag? There are no Cerebral 1 packages with this name and after release the beta tag will be removed.

README.md
henri-hulski: So remove this now?

Guria: lets keep it until we figure out how we deal with it.

README.md
Guria: just to be on safe side before release. eg router will be updated and potentially could break smth if updated automatically.

PR #908, opened 2017-05-18
Run tests and coverage without lerna

gaspard:
We now have a `run.js` script which we can use to run any set of scripts inside the packages, including test and coverage.

The script runs tasks in parallel and is 4x faster then lerna on my machine (5s instead of 19s).

The last dependency to lerna is for setup and this will be resolved when we finish merging other PR in.

image


package.json
```diff
@@ -92,5 +92,8 @@
       "**/node_modules/**",
       "**/build/"
     ]
+  },
+  "devDependencies": {
```
Guria: Now we have all kind of deps in "dependencies". It makes sense to split them I suppose since now we have "normal" dependencies in root package.json . What do you think?

gaspard:
OK, I rebased on master branch, added 'devDependency' check and separated package.json dependencies accordingly.

PR #914, opened 2017-05-23
chore: corrected package.json to make sure tests can run on windows

gaspard:
This fixes bug when running tests in current master. The commit is by @reflog and was on Typescript branch. Moving it here to ease things.


gaspard:
@christianalfoni This needs to be pulled in before other PR as these rely on this fix.

PR #916, opened 2017-05-24
feat(state): add toggle mutation

sveint:
Adds toggle mutation to state for convenience.

### DIFF
--- a/docs/api/state.md
+++ b/docs/api/state.md
@@ -20,6 +20,8 @@ The only way to update the state of your application is in an action. Here is a 
 function someAction({state}) {
   // Set or replace a value
   state.set('some.path', 'someValue')
+  // Toggle a boolean value
+  state.toggle('some.path')
   // Unset a key and its value
   state.unset('some.path')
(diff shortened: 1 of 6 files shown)


gaspard:
Hi @sveint nice work. Could you also change `toggle` operator so that it uses this new method in state for consistency ?

https://github.com/cerebral/cerebral/blob/master/packages/cerebral/src/operators/toggle.js

sveint:
Good point! Changed it now.

Guria:
Also would be great to see it in operators documentation

sveint:
@Guria Since the `toggle` operator was added previously be someone else, it's already there. See http://cerebraljs.com/docs/api/operators.html#toggle . Or did you refer to something else?

edgesoft:
@sveint Think he means here http://cerebraljs.com/docs/api/state.html

sveint:
@edgesoft That should already be covered in this pull request.

Guria:
Ahh. I see now.

sveint:
Thank you :-)

PR #917, opened 2017-05-24
new repository structure

gaspard:
This push is mainly to test travis and deploy.


.gitignore
```diff
@@ -1,4 +1,5 @@
-node_modules
+/node_modules
+/packages/*/**/node_modules
```
henri-hulski: Why not?
```
node_modules
!/packages/node_modules
```

package.json
```diff
@@ -5,9 +5,7 @@
   "license": "MIT",
   "scripts": {
     "link": "link-parent-bin -s true && link-parent-bin -s true -c babel && link-parent-bin -s true -c demos && link-parent-bin -s true -c website",
```
henri-hulski: Do we still need "link"? At the moment you use it in `.travis.yml` only. But seems redundant.

package.json
```diff
@@ -87,7 +90,6 @@
     "istanbul-combine": "^0.3.0",
     "jest": "^18.1.0",
     "jsdom": "^9.9.1",
-    "lerna": "^2.0.0-rc.4",
     "link-parent-bin": "^0.1.1",
```
henri-hulski: When link is not needed we can remove "link-parent-bin" also.

.gitignore
gaspard: Cool. Much better. I thought it would not ignore "node_modules" behind. But works fine.

Guria:
Thanks guys. Great work. Few notes here:

* lerna.json could be removed
* it's not neccessary to put demos under node_modules. it can live near babel / codemods in right in `packages` folder
* I would suggest to remove license files from every subpackage
* packages package.json could omit properties that exact same as in root package.json. eg repository, bugs, license, homepage. see [my attempt as refference](https://github.com/cerebral/cerebral/pull/902/files#diff-ad23321f5ce54c09780f5fde223fc4e6)
* we should avoid prepublish step on demos in postinstall script

.travis.yml
```diff
@@ -26,11 +22,10 @@ after_success:
   fi
   if [[ $TRAVIS_BRANCH == 'next' && $TRAVIS_PULL_REQUEST == 'false' ]]; then
     # TODO: check changed path to decide if npm publish necessary
-    npm run link;
-    lerna bootstrap;
     echo //registry.npmjs.org/:_authToken=${NPM_TOKEN} > ~/.npmrc;
-    lerna publish -c --npm-tag next --yes --exact;
-    npm run deploy && echo 'Website updated.';
+    # TODO: implement publish script @gaspard
```
Guria: don't hurry with publish script. there are a lot of details we must consider. let's finish this step first

package.json
```diff
@@ -4,10 +4,9 @@
   "version": "0.0.1",
   "license": "MIT",
   "scripts": {
-    "link": "link-parent-bin -s true && link-parent-bin -s true -c babel && link-parent-bin -s true -c demos && link-parent-bin -s true -c website",
-    "setup": "lerna bootstrap",
-    "setup:packages": "lerna bootstrap --ignore \"@(cerebral-*|babel-*)\"",
-    "setup:babel": "lerna bootstrap --ignore cerebral-*",
+    "link": "node ./scripts/run.js link",
```
Guria: scripts needs some love in sorting, I suppose :)

henri-hulski:
I'm also not sure if `website` and `codemods` should be inside the packages directory.

gaspard:
OK. I cleaned up commits and moved demos. Everything works fine now, including build and run of demos.

To summarise, this PR does the following:

- [ ] Move packages to `packages/node_modules`
- [ ] Link binaries from root node_modules to individual packages (to allow running npm commands from inside packages) using custom script (link-parent-bin was too complicated and got confused)
- [ ] Fix build issues with node-sass, react-scripts, standard and eslint (all these did not support properly looking for `node_modules` and had to be upgraded or patched)
- [ ] Add cross-env for demos

Guria:
okey. fixed cerebraljs avatar

PR #923, opened 2017-05-25
chore: use prettier for code formatting

henri-hulski:


gaspard:
This is really cool @henri-hulski ! Everything will be a tad different but so much more consistent...

.eslintrc.json
```diff
@@ -0,0 +1,36 @@
...
+  ],
+  "rules": {
+    "prettier/prettier": [
+      "error",
+      {
+        "trailingComma": "es5",
+        "singleQuote": true,
+        "bracketSpacing": false,
+        "semi": false,
+        "bracketSpacing": true
```
gaspard: bracketSpacing twice in settings...

package.json
```diff
@@ -4,22 +4,22 @@
...
     "setup": "node ./scripts/run.js prepublish nodemos",
-    "test": "node ./scripts/run.js test"
+    "build:demos": "node ./scripts/run.js prepublish demos",
+    "link": "node ./scripts/run.js link",
+    "postsetup": "npm run link",
+    "checkdeps": "node ./scripts/dependencies.test.js",
+    "pretest": "npm run checkdeps",
+    "test": "node ./scripts/run.js test",
+    "deploy": "cd website/publisher && npm run ci",
+    "format": "prettier-eslint --write \"**/*.js\"&&prettier-eslint --write \"packages/node_modules/**/*.js\""
```
gaspard: Please keep scripts sorted...

package.json
```diff
@@ -56,23 +62,9 @@
...
-    "ignore": [
-      "/node_modules/**",
-      "/packages/*/**/node_modules/**",
-      "**/build/",
-      "**/*.min.js",
-      "**/bundle.js",
-      "coverage/**",
-      "vendor/**"
-    ]
-  },
```
gaspard: Yeheee ! Won't regret this one.

package.json
```diff
@@ -94,17 +95,19 @@
...
     "node-sass": "4.5.3",
     "npm-run-all": "^4.0.2",
     "nyc": "^10.3.2",
+    "prettier": "^1.3.1",
+    "prettier-eslint-cli": "^4.0.2",
     "prop-types": "15.5.10",
     "react-addons-test-utils": "^15.4.2",
     "react-scripts": "^1.0.6",
     "shx": "^0.2.2",
-    "standard": "github:gaspard/standard#v8.6.0fix2",
```
gaspard: Hahah, this fix didn't live long. Cool.

packages/codemods/1.x-2.x/__testfixtures__/computed-decorator.input.js
```diff
@@ -1,15 +1,15 @@
 /* eslint-disable */
-import {connect as Cerebral} from 'cerebral/react';
-import {limitedList} from "./limitedList";
-import {anotherComputed} from "./anotherComputed";
+import {connect as Cerebral} from "cerebral/react"
```
gaspard: Why isn't single quote rule respected here ?

package.json
henri-hulski: @gaspard I tried to sort them logically. ;-)

packages/codemods/1.x-2.x/__testfixtures__/computed-decorator.input.js
henri-hulski: `packages/codemods/1.x-2.x/__testfixtures__` is excluded from linting as it uses decorators whhich are not compatible with prettier.

.eslintrc.json
henri-hulski: Removed it entirely as it's the default.

packages/codemods/1.x-2.x/__testfixtures__/computed-decorator.input.js
henri-hulski: `single quote rule`? Or do you mean `bracketSpacing`?
I ran prettier once before excluding it but changed the `bracketSpacing` rule after.
I can fix that.

edgesoft: "singleQuote": true

henri-hulski: Does not work. Seems they are also excluded on file base.
Will leave them as they're not part of code base.

henri-hulski: Ah now I see. Yeah this directory is not linted at all.

henri-hulski: If the fixtures have to look exactly like the resulted code we should fix this by hand.
Maybe after this PR is merged.

package.json
Guria: @henri-hulski it's very hard to reason logic in this case. I have tried it and there always was ambiguous result. So seems alphabetic order is only option to have stable unopinionated order.

packages/codemods/1.x-2.x/__testfixtures__/computed-decorator.input.js
gaspard: OK, no problem.

edgesoft:
Awesome!

PR #925, opened 2017-05-26
docs(tutorial): fix router explanation

christianalfoni:


edgesoft:
@christianalfoni Could you maybe update migration as well

PR #926, opened 2017-05-26
feat(Views): warn when using objects and arrays on connected component

christianalfoni:
Connected components should not have objects and arrays passed as props:

```js
export default connect({
  foo: state`foo`
},
  function Comp ({foo}) {
    return <div>{foo}</div>
  }
)

// Some other component
<div>
  <Comp obj={{}} array={[]} />
</div>
```

The reason is that connected components has `return false` on `shouldComponentUpdate` and strict equality check on the props to optimize. Which means if one of these objects are mutated it will not update itself. BUT! Passing objects and arrays to connected components should not happen, cause:

1. They are very likely from the state tree, so should rather be connected directly
2. Generally you should not pass mutable objects and arrays as props, because it prevents render optimizations in general

### DIFF
--- a/packages/node_modules/cerebral/src/utils.js
+++ b/packages/node_modules/cerebral/src/utils.js
@@ -275,3 +275,17 @@ export function createDummyController(state = {}, signals = {}) {
     },
   }
 }
+
+export function verifyPlainProps(props, componentName) {
+  if (!props) {
+    return
+  }
+
+  Object.keys(props).forEach(prop => {
+    if (typeof props[prop] === 'object' && props[prop] !== null) {
+      console.warn(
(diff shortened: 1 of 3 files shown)


christianalfoni:
Need to think a bit more on this, cause there are valid cases for passing objects and arrays to connected components :)

christianalfoni:
Okay, I figured out a way to handle this. When objects/arrays are extracted by components they (and immediate objects) will be flagged as "Cerebral State", allowing connected components to identify these objects and give warning :)

Guria:
@christianalfoni be careful. PR was rebased

christianalfoni:
Closing this as I found a much better way of handling freezing of state objects, which will also fix this issue :-) So will be a bigger change in different PR

PR #929, opened 2017-05-26
Lint

Guria:


package.json
```diff
@@ -11,13 +11,14 @@
     "coverage": "node ./scripts/run.js coverage",
     "coverage:upload": "cat ./coverage/lcov.info | coveralls",
     "deploy": "cd website/publisher && npm run ci",
-    "format": "prettier-eslint --write \"**/*.js\"&&prettier-eslint --write \"packages/node_modules/**/*.js\"",
+    "format": "eslint --fix \"**/*.js\"&&eslint --fix \"packages/node_modules/**/*.js\"",
```
henri-hulski: Not sure if we need the part after &. 
It was necessary for prettier-eslint-cli as it ignores node_modules directories by default. Here I don't know.

Guria: Ok, i'll check it. Thanks

Guria: yeah it works, thanks

PR #939, opened 2017-05-27
fix: options can not be null or undefined

edgesoft:


Guria:
@edgesoft how about some tests?

edgesoft:
@Guria Going away on vacation early in the morning. Will write test for this. Maybe merge anyway because this is a bug :)

edgesoft:
#940

PR #942, opened 2017-05-28
feat(forms): use get from computed instead of form in rules

christianalfoni:
This change is breaking IF you have created custom rules using the second/third argument. Now the args to rules are: `value`, `arg` and `get`. The `get` is from the computed allowing a rule to depend on any state or props passed to component, also tracking changes to it for rerender. This fixes issues where one field depends on an other etc.

### DIFF
--- a/packages/demos/forms-demo/src/index.js
+++ b/packages/demos/forms-demo/src/index.js
@@ -10,17 +10,17 @@ import App from './components/App'
 import FormsProvider from '@cerebral/forms'
 
 const controller = Controller({
-  router: Router({
-    routes: {
-      '/': 'app.routed',
-      '/simple': 'simple.routed',
-    },
-    onlyHash: true,
-  }),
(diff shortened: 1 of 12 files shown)


Guria:
hmm. could we have form or field path in validation rule?

christianalfoni:
Yeah, I was thinking about that @Guria , could add path as last argument there. Do not think formPath makes any sense, cause we do not know the path to the form based on the field, cause the field might be underneath one or multiple namespaces. But yeah... passing in the fieldpath makes sense :) Cause then equalsField rule could be:

```js
// ['equalsField:otherField']
function equalsField (value, field, get, fieldPath) {
  const pathArray = fieldPath.split('.')
  const parentPath = patArray.slice(0, pathArray.length - 1)

  return value === get(state`${parentPath}.${field}`)
}
```

Guria:
ok. lets keep it as possible addition in future

PR #951, opened 2017-05-31
feat(tooling): automatically write new dependencies in package.json

gaspard:
If there are no conflicts, this automatically writes new dependencies in package.json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "license": "MIT",
   "scripts": {
     "build:demos": "node ./scripts/run.js prepublish demos",
-    "checkdeps": "node ./scripts/dependencies.test.js",
+    "checkdeps": "node ./scripts/check.dependencies.js --auto-fix",
     "commit": "git-cz",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
   "license": "MIT",
   "scripts": {
     "build:demos": "node ./scripts/run.js prepublish demos",
-    "checkdeps": "node ./scripts/dependencies.test.js",
+    "checkdeps": "node ./scripts/check.dependencies.js --auto-fix",
```
Guria: not sure it's a good idea. it could autofix on CI but it would stay on CI only then

gaspard: But as it does not do an install after checkdeps, that has no influence on the test which should fail anyway due to missing installed dependencies. The idea is that people run `npm run test` before the stage stuff and commit. With this workflow:

1. write code in new branch
2. test
3. all good =>git add ..., git cz

Guria: but it not enforced to run a test, so it will fail at some point

Guria: could we fail if `process.env.CI === true` instead of autofix happen?

gaspard:
OK. I rebased and disabled `autofix` when running on `CI`.

PR #953, opened 2017-06-01
View tests, fix typos 

yusufsafak:
Views
- add view tests and comment out unnecessary react tests (maybe they can be use later)
- fix typo mistakes
- move some React errors to View
- change View onMount and onUnmount method name to mount and unMount

Devtools
- update devtools tests using new Views
- add sendBulkMessage test

Vuejs tests are coming soon :)  just for testing I have learned basic stuffs. I have never used it before :)

By the way views refactoring is awesome !!

PR #955, opened 2017-06-02
fix(Views): fix vue props handling

christianalfoni:


yusufsafak:
fix is already in https://github.com/cerebral/cerebral/pull/953

PR #956, opened 2017-06-06
docs(website): added contributors following all-contributors spec

reflog:
I think it would be nice to recognize everyone who's been working on the project.
What do you guys think?


reflog:
test is failing currently due to TS problems that @gaspard is currently fixing in another pull request

.all-contributorsrc
```diff
@@ -0,0 +1,128 @@
...
+    },
+    {
+      "login": "fopsdev",
+      "name": "Philipp Knecht",
+      "avatar_url": "https://avatars0.githubusercontent.com/u/5448300?v=3",
+      "profile": "http://www.itflies.ch",
+      "contributions": [
+        "doc",
+        "test"
+      ]
```
henri-hulski: Here we should add "tutorial" as he started with the new tutorial for Cerebral 2.

Guria:
consider adding missing from here:
http://cerebral-website.herokuapp.com/contributors

Guria:
see also contributors group in discord

.all-contributorsrc
```diff
@@ -0,0 +1,128 @@
...
+        "code",
+        "infra",
+        "tool"
+      ]
+    },
+    {
+      "login": "gaspard",
+      "name": "Gaspard Bucher",
+      "avatar_url": "https://avatars3.githubusercontent.com/u/21707?v=3",
+      "profile": "http://feature-space.com",
```
gaspard: Just changed: http://gaspardbucher.com

gaspard:
Hmmm, I have some reluctance to such a list for maintainability reasons and because these lists always end up out-of-sync, only listing the contributors that were present at the time of list creation.

I would simply have the creators of the project and then a link to this: 

https://github.com/cerebral/cerebral/graphs/contributors

I feel the above link is a much better indicator of who did what, it evolves naturally with time, allows people to see who is currently active and who seems like a newbie but used to be super present, etc.

If we really want to have some kind of nice looking view of the above list, I think it would be better to use the github API to automatically generate this on website build. This ensures that pictures, websites, names, and contributors are in sync and there is not too much biais in the list.

gaspard:
And then what about super supportive people in forums and other non-code related parts of the project ? What about our spouses who support our time consuming activities ?

What is the use of such a list anyway ? I don't feel like I it to be grateful to all the contributors for their hard work. The only reason for such a list is to help newcomers get a grasp of who knows what.

In this regard, I would instead create a "resources" list: names of people and in what environment they use Cerebral + other remotely related skills. Maybe the Wiki is a better place for such things...

henri-hulski:
I'm agree with @gaspard.

christianalfoni:
Yeah, you make a good point @gaspard. Should rather have a general, non naming, thank you to the community as a whole. And then put names where people have taken extra responsibility :)

PR #959, opened 2017-06-08
refactor(operators): refactor increment operator

yusufsafak:


packages/node_modules/cerebral/src/operators/increment.js
```diff
@@ -5,42 +5,27 @@ export default function(target, value = 1) {
         'Cerebral operator.increment: You have to use the STATE or PROPS TAG as first argument'
       )
     }
+    let targetValue = resolve.value(target)
+    if (!Number.isInteger(targetValue)) {
```
Guria: Number.isInteger requires polyfill to work in IE and older browsers. :(

Guria: I would suggest putting this in utils file:
```js
const isInteger = Number.isInteger || (value) => 
  typeof value === 'number' && 
  isFinite(value) && 
  Math.floor(value) === value
```

yusufsafak: I thought the same :) but we don't need to put it into utils unless we need it in the other files. I would suggest keep it in this file. 
Btw. I am closing this PR because it does not support deep props setting. I will fix it in #960.

PR #960, opened 2017-06-09
WIP: refactor operators and add new features

yusufsafak:
- refactor array operators (works with props, no breaking changes, less code, add more functionality to array operators )
- add new operators filter and indexOf (we can easily add more operators from array.prototype :) )
- included https://github.com/cerebral/cerebral/pull/959 

Some of the use cases:

```
state: {
 some: {
   list: [12, 5, 8, 130, 44, 30]
 }
}
```

**concat**

concat(target, valueN)

```
concat(state`some.list`, [7,9])  //  some.list = [12, 5, 8, 130, 44, 30, 7, 9]

concat(state`some.list`, [7,9], [1]) // support more than one array some.list = [12, 5, 8, 130, 44, 30, 7, 9, 1]

props: {
 list: [7,9]
}
concat(state`some.list`, props`list`)  //some.list = [12, 5, 8, 130, 44, 30, 7, 9]

concat(props`list1`, state`some.list`,  props`list`, [1]) //props list1=[12, 5, 8, 130, 44, 30, 7, 9, 1] can be also override props list
```

**pop**

pop(target)
pop(target, arrayValue)

```
pop(state`some.list`) // some.list = [12, 5, 8, 130, 44]

props: {
 list: [7,9]
}

pop(props`list`) //props list=[7] 

pop(props`list1`, state`some.list`) //props list1= [12, 5, 8, 130, 44]   and  some.list = [12, 5, 8, 130, 44,30]
```

**indexOf**
indexOf(target, arrayValue ,searchElement)
indexOf(target, arrayValue, searchElement, fromIndex)

```
indexOf(props`index`, state`some.list`, 5) //props index = 1

indexOf(state`index`, state`some.list`, 5) //state index = 1
```

**filter** 
filter(target, arrayValue, callback)

```
filter(state`some.filteredList`, state`some.list`, function bigData (value) {
     return value >= 10
})  // some.filteredList = [12, 130, 44, 30] and some.list = [12, 5, 8, 130, 44,30]

filter(props`filteredList`, state`some.list`, function bigData (value) {
     return value >= 10
})  // filteredList = [12, 130, 44, 30] and some.list = [12, 5, 8, 130, 44,30]
```
In the debugger you can see it like 
```
operator.filter state`app.filteredList`, props`filteredList`, bigData
```

If these changes are OK, I will continue to add more operators, docs and tests.


christianalfoni:
I think this overall looks really great @yusufsafak! My only concern is the `createArrayOperator` function. I think it might be "too clever", meaning that it will be difficult for other developers to understand its generic nature and add new operators. That said, I do not have a clear picture of how it would look if each operator was implemented "as is".

Dry code often comes with a cost, and the cost here I think is further contributions will be more difficult.

Hope I am not being discouraging here, it is a difficult thing to balance. Do you think there is some compromise to be made here? Making it easier to understand how to build an array operator?

Guria:
@christianalfoni I see no problem here with `createArrayOperator`. Looks pretty reasonable for me.

yusufsafak:
@christianalfoni I had the same feeling when I was writing the code :) so I wanted to ask you before I go a step further. 
I understand what you meant but I hate also code duplication :)  If you look at the current array operators, you can easily see a lot of code duplication.
Yesterday I checked my code again and found some issues. So I created another two version of it at gist. 
One is just adding props support and refactoring. No new operators. 
https://gist.github.com/yusufsafak/d0fe7902284637b8c81b9a9b145e6225 

The other one is adding almost all of the array.prototype methods except forEach, length and keeping the model supported operator the same as above because they are modifying state array different than just setting a new array.
https://gist.github.com/yusufsafak/703e07bdc7a561e249126a63444b6b56 

I think, it is much more clear and understandable now.

Btw. I can create also another repo, if it doesn't seem good enough.

yusufsafak:
With a little changes now you can easily create a new array operator using lodash, underscore, ... or your custom function :)
more generic version of createArrayOperator and the demo 
https://www.webpackbin.com/bins/-KmWCfjA9Lk4d-8GHx8w

christianalfoni:
Yeah, since this is so low level and will probably never be touched again I think it is okay. As the "second developer" I am super scared of changing anything in `createArrayOperator` and have no idea what operators will be affected by a code change in there... BUT, that is why we have tests :-) Absolutely not saying it is bad code, it is just very clever ;-)

But okay, I will certainly not block this, it is a great contribution! Thanks @yusufsafak !

christianalfoni:
We should add docs for this as well :)

Guria:
@yusufsafak waiting for documentation and d.ts files update

gaspard:
I am a huge fan of higher order programming: I find that it makes the code actually more robust because we avoid all the silly mistakes that happen with repetitive code when refactoring. `createArrayOperator` is really a good idea. Nice work @yusufsafak !

christianalfoni:
Hm, looking through the API changes I do not recognise the signature of all of them. Like `indexOf` for example. Since we do not have any requests for these operators I will leave it until @yusufsafak pops up again :)

Or somebody else wants to review it

christianalfoni:
Okay, I have to close this is at is expiring. Please reopen! :)

PR #965, opened 2017-06-14
feat(Views): Expose cerebral controller in Angular.js service

sveint:
Having direct access to cerebral's controller is crucial for many use cases within Angular. This PR exposes it as a `controller` property on the `cerebral` service.

Also updated docs with an example.


sveint:
Rebased and fixed conflicts. Should be ready to merge.

PR #966, opened 2017-06-15
feat(Views): Component name for Angular

sveint:
Add possibility for adding component name when using angular

### DIFF
--- a/docs/api/components.md
+++ b/docs/api/components.md
@@ -235,11 +235,23 @@ angular.component('myComponent', {
   controller: connect({
     foo: state`foo`,
     click: signal`clicked`
-  }, function MyController () {
+  }, 'myComponent', function MyController () {
     // Optionally add custom behaviour to controller
   })
 })
 ```
+Since angular doesn't expose the component name,
+you will need to provide one to `connect` for the
+component to be given a name in cerebral.
+
+You can call connect in the following ways:
+```
+connect(dependencies)
+connect(dependencies, name)
+connect(dependencies, controller)
+connect(dependencies, name, controller)
+```
+
 
 ## Vue (BETA)
 [Website](https://vuejs.org/)
--- a/packages/node_modules/cerebral/src/views/angularjs/connect.js
+++ b/packages/node_modules/cerebral/src/views/angularjs/connect.js
@@ -1,10 +1,17 @@
-function connect(dependencies, extendedCtrl) {
+function connect(dependencies, arg1, arg2) {
+  // Supported signatures:
+  // connect(mapToProps, displayName: string, extendedCtrl?: () => {});
+  // connect(mapToProps, extendedCtrl: () => {});
+
+  let displayName = typeof arg1 === 'string' ? arg1 : 'Unknown'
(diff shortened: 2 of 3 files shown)


docs/api/components.md
```diff
@@ -235,11 +235,17 @@ angular.component('myComponent', {
...
     // Optionally add custom behaviour to controller
   })
 })
 ```
+Since angular doesn't expose the component name,
+you will need to provide one to `connect` for the
+component to be given a name in cerebral.
+
+Both component name and controller are optional arguments.
+You can provide either or both, and in any order.
```
Guria: Do we really need such flexibility?

packages/node_modules/cerebral/src/views/angularjs/connect.js
```diff
@@ -1,32 +1,47 @@
-function connect(dependencies, extendedCtrl) {
-  if (typeof extendedCtrl === 'function') {
+function connect(dependencies, ...args) {
+  let displayName = 'Unknown'
+  if (args.length > 2) {
+    throw new Error(
+      `connect() expects 2 or 3 arguments, got ${args.length + 1}.`
+    )
+  }
+  let displayNameIdx = args.findIndex(a => typeof a === 'string')
```
Guria: findIndex is not supported in IEs? Is it worth to use it just for flexible argument position?

docs/api/components.md
christianalfoni: You mean "in any order"? I think we do because there is mostly no need to add controller. And not sure if we should force naming... cause we do not do that in Vue either.

Only change here would be to force naming the component, which could force argument order... but since both controller and name is optional they have to be flexible. Easier to say "any order" than "any order of last two arguments" :)

packages/node_modules/cerebral/src/views/angularjs/connect.js
christianalfoni: Is it not transpiled?

Guria: nope. it is not something related to syntax. it requires polyfills

docs/api/components.md
Guria: We still can support:
```ts
connect(mapToProps, displayName, controller?);
connect(mapToProps, controller?);
```
but not 
```ts
connect(mapToProps, controller, displayName?);
```

packages/node_modules/cerebral/src/views/angularjs/connect.js
sveint: Thanks, I'll fix this.

docs/api/components.md
sveint: @christianalfoni and me landed on this after some discussion (before this PR was created), but for me it doesn't matter much, changing it is fine with me. Depends on whether we want to force users to create a name when creating a controller...

sveint:
Made the changes. Please let me know if there's more to fix.

Guria:
@sveint could you squash related commits, so only 2 left as was originally.
I can assist you if you don't know how/

sveint:
Done :-)

PR #967, opened 2017-06-17
feat(UniversalController): allow multiple runs of Signals.

sladiri:
Add an option to skip runOnce check in UniversalController.

### DIFF
--- a/packages/node_modules/cerebral/src/UniversalController.js
+++ b/packages/node_modules/cerebral/src/UniversalController.js
@@ -1,5 +1,4 @@
 import Controller from './Controller'
-import { throwError } from './utils'
 
 class UniversalController extends Controller {
   constructor(controllerOptions) {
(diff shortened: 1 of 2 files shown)


christianalfoni:
@sladiri I suggest we allow multiple runs by default. We rather set `hasRun` to `true` when `getScript()` is called.... meaning that there is no reason to update the controller anymore. What you think? :)

sladiri:
@christianalfoni OK, I was cautious in case you might want to keep the old default behavior. It works for me so far. :)

sladiri:
@christianalfoni Do you require that the coverage is at least the same, should I add tests to increase it? I did not want to add a test just for the sake of better coverage.

sladiri:
I botched a test before, it ran just one signal (`should allow multiple runs`). Fixed it now.

PR #969, opened 2017-06-19
fix(function-tree): correct async function names.

sladiri:
Fix substring index for async functions.


sladiri:
There is actually an issue with shorthand functions too:
```javascript
{
  fn() {...
}
```
Should I edit this PR to address this issue too?

sladiri:
Hm, the build server does not support async functions. I guess this async part cannot be done then?

sladiri:
I created a new PR with broader scope https://github.com/cerebral/cerebral/pull/970

Issue #968, opened 2017-06-19
Function tree truncates async function names


sladiri:
The staticTree.js does not support async functions

```javascript
function getFunctionName(fn) {
  let ret = fn.toString()
  ret = ret.substr('function '.length) // This index is wrong for async functions
  ret = ret.substr(0, ret.indexOf('('))

  return ret
}
```

PR #970, opened 2017-06-19
fix(function-tree): fix name extraction for functions.

sladiri:
…and syntax when extracting their names.


sladiri:
Sorry, this fixes https://github.com/cerebral/cerebral/issues/968

PR #972, opened 2017-06-21
Developer help fixes

christianalfoni:
We have had a pretty strict error/warning system now to help developers not doing "the wrong thing". We have been a bit too strict here, so I have changed a couple of strategies:

1. Now it is allowed to set a non existing path:

```js
/*
  Given state:
  {}
*/
function myAction ({state}) {
  state.set('foo.bar.baz', 'mip')
}
/*
  New state:
  {foo: {bar: {baz: 'mip'}}}
*/
```

2. When pointing to a non existing signal path in `connect` it will just return undefined:

```js
connect({
  someSignal: signal`a.non.existing.path`
},
  function MyComp ({someSignal}) {
    someSignal // "undefined"
  }
)
```

3. The model no longer does a bunch of freezing and copying of objects. This has cleaned up a lot of code. It is highly unlikely that someone would mutate state inside `components`, the concern is rather inside `actions`. So now the `state.get` API rather copies and freezes objects extracted when devtools has `preventExternalMutations`.

The reason we do not want this inside components is because `dev` and `prod` will behave differently. Since all objects are now copied and freezed any comparison of objects inside components will  be of immutable nature in dev, but mutable in prod... not good.

4. Instead of components now checking if objects are frozen to identify a prop being object from state tree, rather the model adds a "hidden" property to be checked. This is a lot more effective.


packages/node_modules/cerebral/src/utils.js
```diff
@@ -276,12 +276,61 @@ export function createDummyController(state = {}, signals = {}) {
...
+      return Object.freeze(makeCopy(object))
+    }
+
+    return object
+  }
+
+  return freezeObject(rootObject)
+}
+
+export function addCerebralStateKey(object) {
```
Guria: could we have a few words on what this for and why we need it?

christianalfoni: This is needed to identify when developers passes objects and arrays from the Cerebral state tree as props down to other Cerebral components. That is not the ideal solution. All objects/arrays should in these scenarios rather be directly connected.

This was already there in previous version, but based on frozen or not. This was rather fragile, as it turns out also children of React is frozen. No matter, freezing is not part of model anymore so now it rather adds a "hidden key" to allow the `View` to identify these passing of Cerebral objects/arrays as props from one connected component to an other.

docs/api/devtools.md
```diff
@@ -26,6 +26,10 @@ const controller = Controller({
     // state dependencies or signals above the set number  
     bigComponentsWarning: 5,
 
+    // Warnings when passing objects and arrays as props to child
+    // components. They should rather be connected directly
+    warnStateProps: true,
```
christianalfoni: Updated

packages/node_modules/cerebral/src/views/View.js
```diff
@@ -5,8 +5,11 @@ import {
   throwError,
   ensureStrictPath,
   createResolver,
+  getStateTreeProp,
 } from './../utils'
 
+let verifyPropsWarned = false
```
yusufsafak: it should be like https://github.com/cerebral/cerebral/pull/953/files#diff-90e0cc5d48b146c1d2708838e101fc6dR39

packages/node_modules/cerebral/src/utils.js
Guria: Would be great to have few words about it right in code.

christianalfoni: Done :)

PR #974, opened 2017-06-22
feat(router): add getSignalUrl back

christianalfoni:
Added back getSignalUrl. In combination with the new `UniversalController` and loading the router on server side, this should "just work" :-)

### DIFF
--- a/packages/node_modules/@cerebral/router/README.md
+++ b/packages/node_modules/@cerebral/router/README.md
@@ -174,26 +174,45 @@ export default [
 ```
 
 ### getSignalUrl
-Given the routing of:
+Allows you to convert a signal to its corresponding url. This is useful when you actually want to produce the url for a hyperlink etc. To do this you need to create the router in its own file:
 
+*router.js*
 ```js
-{
-  path: '/items/:itemKey',
-  signal: 'items.itemRouted'
-}
+import Router from '@cerebral/router'
+
+export default Router({
+  routes: [{
+    path: '/items/:itemKey',
+    signal: 'items.itemRouted'
+  }]
+})
 ```
 
-You will be able to use the signal to extract the url in your components:
+And attaching it to the controller:
+
+*controller.js*
+```js
+import {Controller} from 'cerebral'
+import router from './modules/router'
+import items from './modules/items'
+
+export default Controller({
+  modules: {items, router}
+})
+```
+
+You will be able to use the same router instance to produce url based on registered signals:
 
 ```js
-connect({
+import router from './router'
+
+export default connect({
   item: state`items.list.${props`itemKey`}`,
-  itemRouted: signal`items.itemRouted`
 },
   function ListItem ({itemKey, item, itemRouted}) {
     return (
       <li key={itemKey}>
-        <a href={itemRouted.getSignalUrl({itemKey})}>{item.name}</a>
+        <a href={router.getSignalUrl('items.itemRouted', {itemKey})}>{item.name}</a>
       </li>
     )
   }
--- a/packages/node_modules/@cerebral/router/README.md
+++ b/packages/node_modules/@cerebral/router/README.md
@@ -173,6 +173,33 @@ export default [
 ]
 ```
 
+### getSignalUrl
+Given the routing of:
+
+```js
+{
+  path: '/items/:itemKey',
+  signal: 'items.itemRouted'
+}
+```
+
+You will be able to use the signal to extract the url in your components:
+
+```js
+connect({
+  item: state`items.list.${props`itemKey`}`,
+  itemRouted: signal`items.itemRouted`
+},
+  function ListItem ({itemKey, item, itemRouted}) {
+    return (
+      <li key={itemKey}>
+        <a href={itemRouted.getSignalUrl({itemKey})}>{item.name}</a>
+      </li>
+    )
+  }
+)
+```
+
 ### EXPERIMENTAL
 We are currently working on functionality that allows you to bind urls to your state, also allowing you to create more complex relationships between your application state and the url. This API is very likely to change, but please feel free to try it out and give us feedback.
 
--- a/packages/node_modules/@cerebral/router/src/base.js
+++ b/packages/node_modules/@cerebral/router/src/base.js
@@ -18,7 +18,23 @@ export default function(options = {}) {
     throw new Error('Cerebral router - mapper option must be provided.')
   }
 
-  return ({ controller }) => {
-    return new Router(controller, addressbar, options.mapper, options)
+  let router
+
+  const routerModule = ({ controller }) => {
+    return (router = new Router(
+      controller,
+      addressbar,
+      options.mapper,
+      options
+    ))
   }
+
+  routerModule.getSignalUrl = function getSignalUrl(signalPath, payload) {
+    return (
+      options.baseUrl +
+      options.mapper.stringify(router.routesBySignal[signalPath], payload || {})
+    )
+  }
+
+  return routerModule
 }
--- a/packages/node_modules/@cerebral/router/src/index.js
+++ b/packages/node_modules/@cerebral/router/src/index.js
@@ -1,12 +1,7 @@
-import { compute } from 'cerebral'
 import urlMapper from 'url-mapper'
 import baseRouter from './base'
 
 const defaultMapper = urlMapper({ query: true })
-
-export function getSignalUrlFactory(config) {
-  return (signal, payload) => compute(signal, payload)
-}
 
 export default function(options = {}) {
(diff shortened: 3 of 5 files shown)


gaspard:
@christianalfoni Seems good but...

Just a question for the future: how will we handle state/computed mapping since we do not have a signal there ?

My gut feeling is that getting an url from a signal is counter intuitive (this actually shows in the code with the signalMapping warning about using the same signal for different paths).

I think this should come from the router with a "named" route:

```js
// router file
export const router = Router({
  routes: [
    {
      name: 'root',
      path: '/',
      signal: 'app.homeRouted'
    },
    {
      name: 'projects',
      path: '/projects/:projectId',
      signal: 'app.projectRouted',
    }
  ]
})
```

```js
// component/Foo.js
import { router } from '../../router'

// ...

  <a href={ router.url ( 'projects', { ... } ) }>bong</a>
```

gaspard:
We could even use this like:

```js
import { helper } from 'cerebral/tags'
connect({
  projectUrl: helper`router.url.projects`
}, ({ url }) => <a href={ projectUrl (xxx) } />
)

**Comment by aguilera51284** [comment]
i need this :)! :+1:

**Comment by Guria** [comment]
Will review it on Monday

**Comment by christianalfoni** [comment]
@gaspard Hm, yeah, the mapping causes a crash here.

I think your naming idea is the only way to go to support mapping really. I think the helper becomes too magical, bringing in the router itself makes more sense. But, there is a problem bringing in router instance on the server. Because on server you might fire up different versions of your controller, meaning different routers.

An important technical limitation is that we need to create a link between the actual instantiated router on the controller and the components.

I have been thinking about the ability to configure the "tag getters". Basically it would allow providers and modules to expose additional tag getters and expose their own tags. An example here with the router:

```js
function RouterModule({controller}) {
  const router = Router()

  return {
    tags: {
      // A tag is just a target and a function that receives the path used on
      // the tag. In this case the path is the name of the url
      // example: url`root`
      url: router.getUrlByName
    },
    state: {},
    signals: {}
  }
}
```

Now all tags will be passed `{state: () => {}, props: {}, url: () => {}}`, so the only missing piece is the router exposing its tag:

```js
import {url} from '@cerebral/router/tags'

connect({
  rootUrl: url`root`
}, ...)
```

It would open up for new stuff as well.

gaspard:
I think `helper` is more powerful and semantically similar to `signal`: it grabs a function from a module where `signal` grabs a signal from a module. There are many similarities and it works for your own modules without having to know anything about the internals of tags. The difference is that:

1. signal is async and it has access and can mutate **state**
2. helper is sync and it has access and can mutate the **module** instance

```js
// Router declares itself as
export const router = {
  state: { ... },
  signals: { ... },
  helpers: {
    url: function ( module, path, args ) {}
  }
}
```

What the `helper` tag does is simply resolve the first two levels as:

```
[module].[helper function name] // controller.getHelper ( path.slice ( 0, 2 ) )
```

And passes the rest as `path` argument (a string array).

The only issue with this is that it is... too easy to misuse by putting more an more hidden state inside modules.

### Use app.state <== Best solution IMO

If it was possible to have router internal state stored in app state, all this would be a no brainer: we would just use a computed (exported by router module):

```js
import { url } from '@cerebral/router'

connect ({
  href: url ( 'routeName', { params: state`some.param` } )
}, ( { href } ) => <a href={ href }>boom</a>
)
```

gaspard:
If the router stores it's internal state and behaves more like a normal module, we can really start to understand and work with the browser addressbar as if it was a component (what `addressbar` package already does).

What I would love to have is something like **triggers**:

```js
export const myModule = {
  state: { ... },
  signals: { ... },
  // Triggers simply avoid us from inserting a pseudo DOM element
  triggers: {
    addressbar: myAdressbar
  }
}

// myAdressbar
import { connect } from 'cerebral/trigger'

export default connect({
  // the state I want to use with the url
  barChanged: signal`
}, { params, barChanged } => {
  // I could fire signals here if I wanted to
  addressbar.value = mapper.serialize ( params )
  addressbar.addEventListener ( 'change', event => {
    event.preventDefault ()
    barChanged ( { value: event.target.value } )
  })
  // no return value
)
```

christianalfoni:
@gaspard Yeah, I agree it would be nice with a generic tag for accessing "helpers"... but at the same time I am a bit hesitant too as it does open up a wormhole for doing all sorts of crazy stuff, as you point out :-)

Storing router state in state tree is a bit problematic. Mostly because it needs to wait for `initialized` to make sure it has access to all defined signals. That means it would need to trigger a signal after `initialized` to update its state about routes. Also problem with computed is accessing the router instance, which has the mapper that converts route to url. You only have access to state and props. If the mapper method is a pure function we could of course just expose it to the computed, but that would mean creating a mapper just to expose the method for the computed.

I think it is wise of us not to introduce any new API here. The router is an edge case. Devs are building Cerebral apps just fine without exposing anything more than state and signals. If we expose more than that we push business logic into components and basically start to look like angular injection.

Also the mapping is experimental. I do not think that should affect our current approach. We might come to the conclusion that in practice a url does require a signal, because what url change does not require business logic? Like authentication and/or data fetching? Or we get to the conclusion that we need to change how the `url` from router is exposed to components. We do not have to decide on it now.

gaspard:
I agree for now and this PR. But as I said on discord, side effects other then the DOM deserve some attention and the router is just showing this. I have other such side effects for which I actually created a `stateWatcher` which is similar to this `trigger` thing but less clean.

Guria:
Hmm. I was expected a more robust solution for new router. 
Now we depend on whether was signal included in Router config or not. And will get very silly `undefined is not a function` errors :(.
Though I can't provide it on my own right now. So I am frustrated now as we should provide some working solution already. But current one could lead to some confusing developer experience.

My thought was to provide API like:
```js
// router.js
import Router, {getSignalUrlFactory} from '@cerebral/router'

const routerConfig = {
  /* ... */
}

export getSignalUrl = getSignalUrlFactory(routerConfig) // computed fn for components
export default Router(routerConfig) // module for controller
```

christianalfoni:
Hm, I do not understand what you mean by:

"Now we depend on whether was signal included in Router config or not. And will get very silly undefined is not a function errors :(."

If you want to create a url based on a signal the route needs to define a signal:

```js
{
  path: '/',
  signal: 'app.homeRouted'
}
```

And you have to ask for the url based on the signal path:

```js
connect({
  url: getSignalUrl('app.homeRouted')
}, ...)
```

Does not matter if we use a factory or an attached method to the signal... it is just way more convenient API to attach method to signal... no extra wiring

Guria:
@christianalfoni 
You don't know whether `signal.getSignalUrl()` is defined when you write component and attach a signal with connect. You have to see a router config to make sure that this signal is used in routes config. It may lead to lots of confusion unfortunately.

christianalfoni:
@Guria Ah, I see what you mean.

But what about a compromise to avoid wiring too much:

```js
// router.js
import Router, {getSignalUrlFactory} from '@cerebral/router'

const routerConfig = {
  /* ... */
}

export default Router(routerConfig) // module for controller
```

And in component:

```js
import {getSignalUrl} from './router'

connect({
  url: getSignalUrl('app.homeRouted')
}, ...)
```

It will also allow us to later do:

```js
import {getUrl} from './router'

connect({
  url: getUrl('nameOfRoute')
}, ...)
```

To support the new experimental mapping stuff?

Guria:
In your example there is no `getSignalUrl` named export from router.js file. So `import {getSignalUrl} from './router'` wouldn't work

Guria:
зштп

Guria:
ping

christianalfoni:
@Guria It is a method on the router instance :)

Guria:
It must be a named export to make `import {getSignalUrl} from './router'` syntax work

christianalfoni:
Wow, this is confusing. `module.exports` is same as `export` for every property of whatever you are exporting on `module.exports`.... but not on `export default`, https://www.webpackbin.com/bins/-Ko41bTmhaAMrfCpN7aW

Anyways:

```js
import router from './router'

connect({
  url: router.getUrl('nameOfRoute')
}, ...)
```

Less wiring

christianalfoni:
I just found a different use case... we really want to extract url wherever. So what I suggest is that we do not used computed, we just:

```js
import router from './router'

connect({
  /* Whatever */
}, 
  function () {
    <a href={router.getUrl('some.signal')}>hello</a>
  }
)
```

The reason is that you might not want to connect the signal at all. You just want to create links dynamically or whatever, no reason to force connecting

Guria:
@christianalfoni Ok, let's go forward with this solution for now.

Guria:
@christianalfoni will you make changes according to latest discussion?

Issue #979, opened 2017-06-23
UniversalController: Module with Router throws error


sladiri:
I created a module with a Router sub-module. When I pass it to the UniversalController on the server, it throws an error:
Inside `Router.onSignalStart`, the `urlMapper.stringify` throws an error, because `payload` is undefined.

I can fix it by deleting the router in the module before creating the UniversalController instance.

christianalfoni:
Thanks @sladiri , I though router was compatible with node due to tests, but it seems it is not. This needs to be fixed :)

sladiri:
It looks like the Router calls `UniversalController.run` with three parameters, but only two are expected. The following code fixes it, but I am not sure what the best way to deal with it is.
```js
// UniversalController.js
run(sequence, payload, foo) {
    if (typeof foo === 'object') {
      sequence = payload
      payload = foo
    }
    return super.run('UniversalController.run', sequence, payload)
  }
```

christianalfoni:
@sladiri Oh, I see... I believe router should rather call `runSignal`, not `run` :)

sladiri:
@christianalfoni As far as I understand, this is a name clash because `UniversalController.run` has a different signature from its superclass. I just guessed that Router called it before, but I cannot see any better way than to check inside `UniversalController.run`, since it deviates from `Controller.run`.

PR #981, opened 2017-06-24
fix(UniversalController): rename run() since super.run() has greater arity

sladiri:
…ferent arity than super.run().

When using the Router module, the payload becomes undefined, because FunctionTree.run() is expected
to take at least three parameters, while currently UniversalController.run() takes only two.

ISSUES CLOSED: #979

### DIFF
--- a/docs/api/universalcontroller.md
+++ b/docs/api/universalcontroller.md
@@ -67,7 +67,7 @@ controller
   })
 ```
 
-You can add any providers to the controller to do database fetching etc. inside this **run** execution. Think of it as a signal the updates the state of the app before rendering it on the server.
+You can add any providers to the controller to do database fetching etc. inside this **runSequence** execution. Think of it as a signal the updates the state of the app before rendering it on the server.
 
 **NOTE!** You should instantiate the controller for each run you want to do.
 
--- a/docs/api/universalcontroller.md
+++ b/docs/api/universalcontroller.md
@@ -18,7 +18,8 @@ const controller = UniversalController({
 ```
 
 ## Methods
-### run
+### run and runSequence
+#### run
 If you need to update the state of the controller you can run a signal execution for doing so:
 
 ```js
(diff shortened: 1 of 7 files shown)


christianalfoni:
Hm.... I am thinking here that maybe running "a signal" is the wrong approach. What I have noticed after actually using this is that you will grab all your state using other mechanisms to go an grab the state. So basically you just want to prepare some state and the shove it into the controller.

So maybe we should rather do:

```js
controller.setState({
  'app.user': user,
  'admin.list': adminList
})
```

Improves readability as well as there is no promise return. Does this fit your scenario as well?

sladiri:
@christianalfoni I get the value of updateState, that would be a good API, but in my case I want to update the model by calling the signals I defined. I thought that I could replace side effects via the context for example.

I have not yet tried much, but do you suggest that I run a FunctionTree instance, not the signals from the module/controller? But how would I pass in the model to such a FunctionTree, so that it would become the state for the actions?

christianalfoni:
@sladiri Hm, yeah, allowing running signals is a cool feature... just replacing the providers with server side providers. 

There is nothing wrong doing both here really. It depends on your approach. Either running custom backend logic to extract state or actually implement providers to run signals. But maybe then that should be the separation?

Lets write a spec here:

1. We implement `setState` as a way to just set some state and use the controller

2. We implement `runSequence` as a way to just run a sequence and return the promise (runOnServer) basically

You wanna play with it or I can have a go at it :-)

sladiri:
@christianalfoni I would like to have a go, yes. :)

Guria:
cc @christianalfoni

christianalfoni:
Great stuff @sladiri :-)

Two small last things: `getModelPathArray`, could you move it to the `utils` instead? Not a very useful public method on the controller :-)

Second just document the two new `UniversalController` methods in the docs? I can do this if you are unsure how to add it :-)

sladiri:
@christianalfoni Certainly, now I see again that `runSequence` is just a convenience delegator, which sets a default name for a signal. (I have not read the docs well enough.) Is this additional method really required? :) I guess everyone would just use `run`, and use their own name, wouldn't they?

packages/node_modules/cerebral/src/utils.js
```diff
@@ -334,3 +334,7 @@ export function getStateTreeProp(props = {}) {
     return hit
   }, null)
 }
+
+export function getModelPathArray(path) {
```
yusufsafak: @sladiri code duplication :) you can just use utils.ensurePath  :)

yusufsafak:
@sladiri you can use utils.ensurePath instead of creating getModelPathArray !

christianalfoni:
@yusufsafak Oh, for sure, thanks for that :)

sladiri:
Nice catch thank you. :)

sladiri:
I guess that manually getting the signal out of the controller is not ideal? :)
Maybe it should accept a name or an array? The name would be used to get the signal inside `runSequence` with `Controller.getSignal`.

sladiri:
In my last change I added the possibility to pass in a signal-path to runSequence. I hope it is OK, I find it cumbersome to extract a predefined module-signal manually from the controller object.

sladiri:
Nice, that was no problem :) Thank you.

docs/api/universalcontroller.md
```diff
@@ -43,10 +43,50 @@ controller.run([
...
+})
+
+controller
+  .runSequence('app.aSignal', {isAwesome: true})
+  .then(() => {
+    // I am done running
+  })
+```
+
 You can add any providers to the controller to do database fetching etc. inside this **run** execution. Think of it as a signal the updates the state of the app before rendering it on the server.
```
Guria: Should **run** be chaged to **runSequence** here?

sladiri: It seems so, yes. I missed that while copying the old text.

christianalfoni:
Moved this PR to https://github.com/cerebral/cerebral/pull/1020, due to current security issues and also rebasing

PR #983, opened 2017-06-25
Views tests

yusufsafak:
#953 rebased and fixed conflicts


packages/node_modules/cerebral/src/views/preact/StateContainer.js
```diff
@@ -8,7 +8,7 @@ class StateContainer extends Component {
     }
   }
   render() {
-    return this.props.children
+    return this.props.children[0]
```
yusufsafak: Btw. Preact Containers don't support more than one child :(
it renders only first children.  you can't use it like 
```
<StateContainer>
    <TestComponent />
    <AnotherComponent />
</StateContainer>
```
It does not render AnotherComponent.

christianalfoni: Does this mean it is fixed? :) `StateContainer` is legacy stuff btw, `UniversalController` is really the replacement

christianalfoni: Though nice for testing, but yeah... totally fine this change :)

Guria:
Sorry @yusufsafak Seems needs once more rebase now :(

PR #986, opened 2017-06-27
Responses and errors in HTTP and Firebase

christianalfoni:
- Makes operators consistently use `response` or `error` output
- Updated docs
- Updated marksy for similar nested links


Guria:
does it brings breaking changes into http?

christianalfoni:
Yeah, both firebase and http will have breaking changes when using `operators` and `paths`. I am not sure how we can avoid breaking here. This is example of change:

### CURRENT

**Success**
```js
[
  httpGet('/items'),
  /*
    PROPS: {
      status: 200,
      result: [],
    }
  */
]
```

**Error**
```js
[
  httpGet('/items'), {
    error: [
      /* PROPS: {error: "message"} */
    ]
  }
]
```

### NEW VERSION

**Success**
```js
[
  httpGet('/items'),
  /*
    PROPS: {
      response: {status: 200, result: [], headers: {}}
    }
  */
]
```

**Error**
```js
[
  httpGet('/items'), {
    error: [
      /* PROPS: {error: HttpProviderError} */
    ]
  }
]
```

Guria:
I am just thinking about how to inform current users about this breaking

christianalfoni:
Hm, yeah... I am not sure of the state of using operators like this. Cause using the provider inside action everything is the same. 

It will be a pretty huge release I suppose? Lots of changes? So maybe we just have to hype it up a bit, maybe even go to RC? Then at the very top list BREAKING?

edgesoft:
This is a important PR.

Guria:
Conflicts in documentation

edgesoft:
@christianalfoni Can you have a look at the conflicts?

packages/node_modules/@cerebral/firebase/README.md
```diff
@@ -260,37 +343,33 @@ Note: When using transactions with Security and Firebase Rules in place, be awar
 
 ## value
 
-*action*
+### action
 ```js
 function someAction({ firebase }) {
-  return firebase.value('someKey.foo', {
```
Guria: @christianalfoni I have a feeling that this part should be kept in some way. Could you confirm that?

Guria:
@christianalfoni made a cleaner rebase here. please verify it again. Also left some placeholders for breaking changes descriptions in commit messages.

packages/node_modules/@cerebral/firebase/README.md
christianalfoni: Okay, reviewed that at made it very explicit, good catch! :)

packages/node_modules/@cerebral/http/src/HttpProviderError.js
```diff
@@ -1,11 +1,11 @@
 export default class HttpProviderError extends Error {
-  constructor(status, headers, body, message = null, isAborted = false) {
+  constructor(status, headers, result, message = null, isAborted = false) {
     super()
     this.name = 'HttpProviderError'
     this.message = message
     this.status = status
     this.headers = headers
-    this.body = body
+    this.result = result
```
gaspard: Not sure I understand the reason to use 'result' instead of 'body'. "Result" is a very generic term that does not rely add any meaning. I think the only places this parameter is used is with "xhr.responseText" so why not name it "responseText" instead ? This is a well documented Wep API name.

christianalfoni: Yeah, I can see why you would want that @gaspard , but the `result` is not the `responseText`. If the error response was JSON, it will be parsed, meaning it is rather a result of that. But `result` is generic. Maybe move back to body?

gaspard: I think "body" would be better. I was scratching my head as to what a "result" could be when obviously it failed (since there is an error)... Or call it "response"...

christianalfoni: Yeah, we can make it the same as the output of the `response` on operators. That would make things consistent.

So:

```js
    this.name = 'HttpProviderError'
    this.message = message // responseText
    this.response = {
      status, headers, result
    }
```

Looks good?

gaspard: Great !

PR #992, opened 2017-06-30
task(test): Remove deprecated test function RunSignal 

garth:


Guria:
what means **task** in commit message?

garth:
A task is a item of work that has no user value.

Guria:
But it still has user impact. it removes part of api. And we don't have `task` type in convention we use:
https://docs.google.com/document/d/1QrDFcIiPjSLDn3EL15IJygNPiHORgU1_OOAqWjiDU5Y/edit#
https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit

christianalfoni:
Where are we on this @Guria and @garth ?

garth:
not sure, just wanted to remove the deprecated test fn before we hit 2.0, but I guess we are too late for that now.

christianalfoni:
@Guria What changes do you require here?

Guria:
I want that only supported commit types was used. For me here is a `feat(test):` with a breaking change section.

henri-hulski:
I think this one should also be fixed and go in now.

christianalfoni:
Moved to: https://github.com/cerebral/cerebral/pull/1100

PR #996, opened 2017-07-04
feat(cerebral): Immutable model

garth:
Add support for overriding the default Cerebral model and include an alternative based on Baobab.

```js
import { Controller } from 'cerebral';
import ImmutableModel from 'cerebral/lib/ImmutableModel';

export default Controller({
    Model: ImmutableModel, // replaces the default
    state: {},
    modules: {}
});
```

This makes is possible for complex projects to move from Cerebral 1.0, which was immutable by default, without having to make a complete rewrite. It's also a safer alternative to the default store for projects who appreciate the benefits of immutability and can accept the performance overhead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "@types/eventemitter3": "1.2.0",
     "addressbar": "1.0.3",
     "aphrodite": "1.2.0",
+    "baobab": "^2.4.3",
     "bulma": "0.2.3",
     "chalk": "1.1.3",
     "classnames": "2.2.5",
--- a/packages/node_modules/cerebral/src/Controller.js
+++ b/packages/node_modules/cerebral/src/Controller.js
@@ -1,7 +1,7 @@
 import DependencyStore from './DependencyStore'
 import FunctionTree from 'function-tree'
 import Module from './Module'
-import Model from './Model'
+import DefaultModel from './Model'
 import {
   ensurePath,
   isDeveloping,
@@ -50,6 +50,7 @@ class Controller extends FunctionTree {
       router,
       devtools = null,
       options = {},
+      Model = DefaultModel,
     } = config
     const getSignal = this.getSignal
 
--- a/packages/node_modules/cerebral/src/ImmutableModel.js
+++ b/packages/node_modules/cerebral/src/ImmutableModel.js
@@ -1,0 +1,92 @@
+import Baobab from 'baobab'
+
+class ImmutableModel {
+  constructor(initialState = {}, devtools = null) {
+    this.devtools = devtools
+
+    this.state = new Baobab(initialState)
+    this.changedPaths = []
+  }
+  /*
+    Returns array of changes
+  */
+  flush() {
+    const changes = this.changedPaths.slice()
+
+    this.changedPaths = []
+
+    return changes
+  }
+  get(path = []) {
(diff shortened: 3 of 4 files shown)


garth:
Preliminary testing shows that this model has fixed a lot of issues in one my older apps that was converted to 2.0 beta.

christianalfoni:
I think this is a great idea @garth :-)

package.json
```diff
@@ -26,6 +26,7 @@
     "@types/eventemitter3": "1.2.0",
     "addressbar": "1.0.3",
     "aphrodite": "1.2.0",
+    "baobab": "^2.4.3",
```
edgesoft: Should this be a dependency? If someone want´s immutable they need to install it?

edgesoft: Ok, sorry. You are providing one out of the box 👍

christianalfoni:
I also created a PR to support mobx-state-tree, which has split the controller into BaseController and Controller... I think maybe there will be a conflict here. https://github.com/cerebral/cerebral/pull/1000

Anyways, both will work, but let me know what you think about the other PR, how it fits in here?

garth:
not sure about the mobx state tree, I can see that these changes will affect each other, but I see this one as being more critical as it affects existing apps and is not an additional feature.

I am currently running a fork of cerebral until this PR or similar fix to address issues of migrating apps from 1.0 is applied.

christianalfoni:
Okay, cool, lets see how things go with merging PRs :) There might be a small change to API here to adjust to the general refactor, but lets look into that when PRs are being merged. It is super minor if any change :)

christianalfoni:
Bring in this PR first: https://github.com/cerebral/cerebral/pull/1000

garth:
Can we just allow the default model to be swapped out with a custom one, then immutable could be a separate module? Until we have this I cannot run the official cerebral release.

Guria:
@garth you can use webpack plugin to override model

garth:
@Guria I'm not sure that patching cerebral via a build process is a good idea. 1st it relies on using webpack (or other build tool that would support this) also projects using CRA without ejecting probably couldn't configure webpack to do this even though they have it.

I understand that supporting projects coming from a previous major point release is not required, but it would be a shame not to do it if we can.

christianalfoni:
With #1000 you can replace the Model in one of two ways:

### Use Cerebral HOC
```js

Controller({
  Model: MyModel
})
```

`MyModel` will be instantiated with the `new` keyword and receive the controller instance. That means with Baobab you would have to wrap it in your own Model like:

```js
import {extractModulesProp} from 'cerebral/lib/utils'

function MyModel (controller) {
  // This is a new utility to extract and merge properties from modules
  const initialState = extractModulesProp('root', controller.module, 'state')

  this.tree = new Baobab(initialState)
}
```

### Use Baobab HOC
```js
import {BaseController} from 'cerebral'

BaseController({
  Model: MyModel
})
```

Now there is no tracking of changes. You would rather use `baobab-react` to connect components and expose signals manually.

### Summary
I would suggest refactoring this PR to use the first method there. Basically it is just wrapping what you have now to retrieve that controller and extract initial state.

christianalfoni:
Closing this, moved to https://github.com/cerebral/cerebral/pull/1071

PR #997, opened 2017-07-05
refactor(router): adding the ability to use the custom addressbar

andrey-sh:
Adds an optional field addressbar to the router options object


christianalfoni:
Hi @andrey-sh, you have some background story for this change? Just curious why you need it :)

christianalfoni:
Closing this as it is getting old and not sure why it is necessary, please create new PR if necessary :)

PR #998, opened 2017-07-08
WIP: provide failing test case for immutable state

sladiri:
As suggested @christianalfoni, this is a PR with a failing test for the issue I have with current master.

This error happens on node 6, node 8, Chrome and Edge.

If I call state.get() without parameter, then a subsequent state.set fails.


naturalethic:
https://github.com/cerebral/cerebral/pull/1015

naturalethic:
Provided in https://github.com/cerebral/cerebral/pull/1015

PR #1000, opened 2017-07-10
feat(Controller): refactor to base controller and add mobx state tree

christianalfoni:
- `BaseController` handles modules, signals and providers
- `Controller` is now only related to default Cerebral model, dependency tracking and flushing
- Moved `StateProvider` as property to the Model itself, so each model can have its own state provider
- Added `cerebral-mobx` package, which just exports the Mobx (Mobx State Tree) controller and handles `model` property in modules

Check README.md on how it works in practice. 

This is of course rather experimental and should be tested more before any new release. But I really want to explore this as it simplifies Cerebral code and at the same time "empowers" the model layer with typing etc. It does of course not support a bunch of view layers, like default Cerebral model, so this could be just a choice for the developer.

### DIFF
--- a/packages/node_modules/cerebral-mobx/package.json
+++ b/packages/node_modules/cerebral-mobx/package.json
@@ -1,0 +1,41 @@
+{
+  "name": "cerebral-mobx",
+  "version": "0.0.0-next",
+  "description": "Combine Cerebral with Mobx state tree",
+  "main": "index.js",
+  "author": "Christian Alfoni <christianalfoni@gmail.com>",
+  "contributors": [
+    "Aleksey Guryanov <gurianov@gmail.com>"
+  ],
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/cerebral/cerebral.git"
+  },
+  "bugs": {
+    "url": "https://github.com/cerebral/cerebral/issues"
+  },
+  "homepage": "http://cerebral.github.io/cerebral-website",
+  "peerDependencies": {
+    "cerebral": "^2.0.0-beta.3",
+    "mobx": "^3.2.0",
+    "mobx-state-tree": "^0.9.1"
+  },
+  "scripts": {
+    "build": "cross-env BABEL_ENV=production babel src/ --out-dir=lib/ -s",
+    "coverage": "nyc --reporter=lcov --reporter=json npm run test",
+    "prepublish": "npm run build"
+  },
+  "nyc": {
+    "exclude": [
+      "node_modules",
+      "lib",
+      "tests",
+      "**/*.test.js",
+      "**/testHelper.js"
+    ]
+  },
+  "devDependencies": {
+    "prop-types": "^15.5.10"
+  }
+}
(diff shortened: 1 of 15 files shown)


Guria:
Must split this in a few commits

Guria:
Is there any breaking changes introduced in public (or not so public) APIs?

christianalfoni:
Hehe, no breaking :-)

packages/node_modules/cerebral/src/Controller.js
```diff
@@ -1,92 +1,32 @@
...
-        'You are grabbing a signal before controller has initialized, please wait for "initialized" event'
-      )
-    }
-
-    if (router) {
-      console.warn(`DEPRECATION: the 'router' module should be in modules.`)
-      modules.router = router
-    }
+    super(
+      Object.assign(config, {
```
Guria: maybe in reverse order? so config would be able to override default and not confuse the user.

packages/node_modules/cerebral-mobx/package-lock.json
```diff
@@ -0,0 +1,18 @@
+{
```
Guria: delete it please

packages/node_modules/cerebral-mobx/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "cerebral-mobx",
+  "version": "2.0.0-beta.3",
```
Guria: align with monorepo

packages/node_modules/cerebral-mobx/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/cerebral/cerebral.git"
+  },
+  "bugs": {
+    "url": "https://github.com/cerebral/cerebral/issues"
+  },
+  "homepage": "http://cerebral.github.io/cerebral-website",
+  "dependencies": {
+    "cerebral": "^1.0.0-beta.3",
```
Guria: peer?

Guria: for all deps here

packages/node_modules/cerebral/src/Controller.js
christianalfoni: You er perfectly right, fixed :-)

packages/node_modules/cerebral-mobx/package-lock.json
christianalfoni: Jup, done

packages/node_modules/cerebral-mobx/package.json
christianalfoni: Done

packages/node_modules/cerebral-mobx/package.json
christianalfoni: Aaand, done

christianalfoni:
Okay, gone through it all, hope I understood changes requested correctly :)

PR #1005, opened 2017-07-18
feat(website): improve navigation and responsive design

bdjnk:
I've restructured navigation for greater power and flexibility; it's also more straightforward. The mobile layout, which hides the navigation menu, is now activated at `850px`.


christianalfoni:
I am working on some improvements here, also on content :-)

christianalfoni:
# New Cerebral branding

# What is Cerebral?

> A declarative **state and side effects management** solution for popular JavaScript frameworks

Writing your code **declaratively** means doing:

```js
<h1 class="header">Hello</h1>
```

Instead of:

```js
const h1 = document.createElement('h1')
h1.className = 'header'
document.body.appendChild(h1)
```

It makes perfect sense for us to write our UIs in a declarative manner. The reason is that we need to reuse UI elements and compose them together in different configurations. One can also claim that declarative code reads better, not just because it is less syntax, but because you only describe **what** you want, not **how** you want it.

But what about our business logic, can we get the same benefits there? This is how business logic can be expressed **declaratively**:

```js
[
  set(state`isLoadingUser`, true),
  httpGet('/user'), {
    success: set(state`user`, props`user`),
    error: set(state`error`, props`error`)
  },
  set(state`isLoadingUser`, false),
]
```

Instead of:

```js
function getUser() {
  this.isLoading = true
  ajax.get('/user')
    .then((user) => {
      this.data = user
      this.isLoading = false
    })
    .catch((error) => {
      this.error = error
      this.isLoading = false
    })
}
```

You might think this example tries to highlight "less lines of code", but that is just a result of these core properties:

- Composability
- Readability
- Reusability
- Testability

All these properties makes **better code**, there is no question about that, and the way we achieve it is through declarative code. **Read on to learn more about how Cerebral enforces these properties in your codebase**.

henri-hulski:
@christianalfoni Should this not go in a seperate PR as this one seems to be ready?

You can base a PR not only on master but on any branch also on this one (`responsive-website`).
Just merging it after can become a little mess if it isn't done carefully. ;-)

bdjnk:
@henri-hulski This one is as ready as I can make it without testing by others. I've tested it on various systems and browsers, and gone through all the scenarios multiple times.

@christianalfoni I do think it makes sense to add content changes to a separate pull request; it seems to me orthogonal work.

christianalfoni:
@henri-hulski @bdjnk Jup, I will branch out this PR :)

Guria:
merged within #1007

PR #1010, opened 2017-07-24
WIP: 'next' branch and releases

gaspard:
Preparing for "@next" release setup on **master**.

- [x] set all versions to 'next'
- [ ] create script for '@next' release (not replacing version, not bumping version, just release and tag with 'next'). Not sure how we handle rolling version like this with npm.
- [ ] make the script run on travis for PR merging in 'master'
- [ ] setup semver publishing on travis for `release` branch
- [ ] protect `release` branch


henri-hulski:
It seems not to be possible to release a package on npm without version number. 
But when adding a custom tag as 'next' it will not be installed by default and doesn't get the latest tag. 

So it seems that even on next releases we have to calculate the new version number and add something like '.beta.1'. Don't know if '.next.1' is also possible.

gaspard:
Working on `publishScript` branch. Let's ignore this for now. And `next` is not a valid version for npm package.

PR #1015, opened 2017-07-28
fix(utils): fixed copyAndFreezeObject to not interfere with original state

naturalethic:


packages/node_modules/cerebral/src/utils.js
```diff
@@ -277,33 +277,24 @@ export function createDummyController(state = {}, signals = {}) {
...
-        } catch (e) {}
-      }
-
-      return Object.freeze(makeCopy(object))
-    }
-
-    return object
+  function deepFreezeObject(obj) {
+    Object.getOwnPropertyNames(obj).forEach(name => {
+      deepFreeze(obj[name])
```
sladiri: Isn't the frozen, returned object thrown away here?

packages/node_modules/cerebral/src/utils.js
```diff
@@ -277,33 +277,24 @@ export function createDummyController(state = {}, signals = {}) {
...
+    })
+    return Object.freeze(obj)
   }
-
-  return freezeObject(rootObject)
+  function deepFreezeArray(arr) {
+    arr.forEach(val => {
+      deepFreeze(val)
```
sladiri: Similar to above, I think you need to map the array with the freeze function?

naturalethic: Object.freeze() will freeze in place:
```
> a = {}
{}
> Object.freeze(a)
{}
> Object.isFrozen(a)
true
```

naturalethic: although map would be smarter probably

sladiri: I interpreted the MDN page about Object.freeze wrongly and I did not try it out myself in the Browser, sorry.
Since this is in place mutation, I think that the forEach is just right.

packages/node_modules/cerebral/src/utils.js
```diff
@@ -277,33 +277,16 @@ export function createDummyController(state = {}, signals = {}) {
...
+      })
+      return Object.freeze(val)
     }
-
-    return object
+    return val
   }
-
-  return freezeObject(rootObject)
+  return deepFreeze(JSON.parse(JSON.stringify(rootObject)))
```
kgdev: instead of JSON.parse and JSON.stringify, can we do like the before using recursion?
which one has better performance?

kgdev: Also return undefined if rootObject is undefined?

christianalfoni: I think JSON will work fine as this will only happen using `state.get` inside actions. So it is pointing to a specific value. It is also only in development :)

christianalfoni:
Okay, closing this in favour of https://github.com/cerebral/cerebral/pull/1018

We are having some security issues with forks currently. I also fixed an issue related to "non complex objects" :)

PR #1020, opened 2017-08-01
Universal controller fix run with router module

christianalfoni:
Moved from: https://github.com/cerebral/cerebral/pull/981

It changes `UniversalController` to have its own `runSignal` kinda of method, called `runSequence`. Since UC extends normal `Controller` it should have its own method for running a signal on server side. Also updated tests and docs

### DIFF
--- a/docs/api/universalcontroller.md
+++ b/docs/api/universalcontroller.md
@@ -20,7 +20,8 @@ const controller = UniversalController({
 ```
 
 ## Methods
-### run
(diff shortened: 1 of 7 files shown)


Guria:
are any breaking changes introduced here?

christianalfoni:
Should not be any breaking here. The `run` will just not be populated with a name for the debugger anymore.While `runSequence` is.

PR #1025, opened 2017-08-02
docs(website): remove beta, adjust to previous contribution by Michael

christianalfoni:
- Adjusted some content on frontpage, added debugger gif
- Uppercase main links
- All links blue
- Custom scroll spy
- Aligned content to left instead
- Removed BETA

screen shot 2017-08-02 at 11 19 01


bdjnk:
The new landing page is off center now.

screen shot 2017-08-02 at 12 45 31 pm

Also, scroll-spy has an interesting bug when sub-headers are the same as an ancestor.

screen shot 2017-08-02 at 12 50 53 pm

henri-hulski:
I think we're agree to make the TOC entries not blue. :-)

PR #1026, opened 2017-08-02
fix: only ignore top level `test` in .npmignore, fixes #1017

### LINKED ISSUE #1017: Cannot find module './lib/test' from 'test.js'
Version: cerebral 2.0.0

Import a test helper
```
import {runAction} from 'cerebral/test'
```

Content of node_modules/cerebral/test.js:
```
module.exports = require('./lib/test')
```
lib/test is missing ...

Maybe just something in an .ignore file filtering test.js files from package, but could not find one ;)

fmal:
test

### DIFF
--- a/packages/node_modules/@cerebral/firebase-admin/.npmignore
+++ b/packages/node_modules/@cerebral/firebase-admin/.npmignore
@@ -1,3 +1,3 @@
 .babelrc
-src
-test
+/src
+/test
--- a/packages/node_modules/@cerebral/firebase/.npmignore
+++ b/packages/node_modules/@cerebral/firebase/.npmignore
@@ -1,3 +1,3 @@
 .babelrc
-src
-test
+/src
+/test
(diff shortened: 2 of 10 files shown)


gaspard:
Why would we need to run the tests on a published module ?

If this is to be fixed, it has to be done in all packages (they have the same 'test' ignore thing).

henri-hulski:
@gaspard without this fix you cannot import the test helper from Cerebral. 
See #1017.

gaspard:
Ah OK makes sense, so can we update the `.npmignore` in other packages as well ? And also fix "src" and "src" which should be "/src" ?

I don't know the details of `.npmignore` but maybe there is a glob thing and we could do:

```
*.test.js
.babelrc
/src
/test
```

Globs, then files, then  absolute paths (most generic to most specific). It would be awesome :-)

fmal:
@gaspard good idea, done 😃

PR #1032, opened 2017-08-07
chore(release): add travis next release script

christianalfoni:
- Updated repo-cooker to handle branch specific logic
- Updated travis file to run publish on master and next branch
- Dry run on both master and next, looks good!


.travis.yml
```diff
@@ -29,11 +29,11 @@ after_success:
     npm run coverage:upload;
   fi
   if [[ $TRAVIS_BRANCH == 'release' && $TRAVIS_PULL_REQUEST == 'false' ]]; then
```
gaspard: Not sure about the "release" branch. I changed this weeks ago but I think we reverted to using the "master".

gaspard: Yes @edgesoft confirms, we should do the automatic release from "master". So this can be merged with the previous `if`.

Guria: hmm. I don't like having any logic in travis.yml scripts now when we have repo-cooker.
I'd like to see here only repo-cooker script invocation. Th rest of the logic shoul go into actions.

gaspard: Yes, seems like a good idea. We then need a new action in repo-cooker to get name of current branch.

package.json
```diff
@@ -45,6 +44,7 @@
     "raf": "3.3.0",
     "react": "15.4.2",
     "react-dom": "15.4.2",
+    "repo-cooker": "^1.0.0",
```
gaspard: This looks gooooooood !! :-) haha 👍

gaspard:
```yaml
# no more before_install
after_success:
- |
  if [[ $TRAVIS_BRANCH == 'master' || $TRAVIS_PULL_REQUEST == 'true' ]]; then
    npm run coverage;
    npm run coverage:upload;
  fi
  if [[ $TRAVIS_BRANCH == 'master' || $TRAVIS_BRANCH == 'next' ]] && [[ $TRAVIS_PULL_REQUEST == 'false' ]]; then
    echo "//registry.npmjs.org/:_authToken=\${NPM_TOKEN}" > .npmrc
    npm run release -- --publish
    # Maybe deploy could be part of the repo-cooker script...
    # git remote add auth https://cerebraljs:${GH_TOKEN}@github.com/cerebral/cerebral;
    # git config --global user.email "cerebraljs@gmail.com";
    # git config --global user.name "Cerebral JS";
    # npm run deploy && echo 'Website updated.';
  fi
```

scripts/release/index.js
```diff
@@ -16,14 +13,33 @@ cooker.run('publish', [
   cook.relatedPackagesByPackage,
   cook.getCurrentVersionByPackage,
   cook.evaluateNewVersionByPackage,
+  cook.byBranch,
```
Guria: should getLatestReleaseHash depend on branch? how we deal with it for different branches?

scripts/release/index.js
```diff
@@ -16,14 +13,33 @@ cooker.run('publish', [
...
+  {
+    next: cook.remap(
+      'newVersionByPackage',
+      (version, props) => `${version}-${props.hash.slice(0, 6)}`
```
Guria: let use datetime here, because hash caused problems with semver

scripts/release/index.js
gaspard: We base all the @next versions on "latest" so this works correctly: we take latest "master release", compute semver and then apply semver on top of latest "master release version". Good for me.

scripts/release/index.js
gaspard: Even a short hash ? So I propose to use

```js
version => `${version}-${new Date().getTime()}`
```

gaspard: I liked the hash idea... What is wrong with hash ?

Guria: it causes problem with ~ and ^ version resolution.

scripts/release/index.js
Guria: next can be released several times before getting into master (aka latest)

Guria: ahh, got it. it's only to evaluate correct version. ok then

scripts/release/index.js
christianalfoni: Updated to datime

PR #1041, opened 2017-08-10
chore(release): use cook and debugger only on dryRun

christianalfoni:


scripts/release/cooker.js
```diff
@@ -7,7 +7,7 @@ process.env.REPO_COOKER_GITHUB_TOKEN = process.env.GH_TOKEN
 
 export const cooker = Cooker({
   devtools:
-    !process.env.CI &&
+    dryRun &&
       new Devtools({
```
gaspard: Should be just the devtools settings: also no need to import Devtools.

```js
// This is what Cooker does:
const tools = new Devtools({
  host: options.devtools ? options.devtools.host : 'localhost:9090',
})
```

christianalfoni: oh, fixed! :-)

Guria:
Seems debugger must be fixed in repo-cooker itself:
https://github.com/cerebral/repo-cooker/blob/259c57c725331b3f1c4057924851eac71f7d953d/src/Cooker/index.js#L22

christianalfoni:
In what way @Guria ? Should only add devtools if devtools option passed in and not in production environment... which it does now?

gaspard:
@christianalfoni Yes, I think we should only check for truthy value in repo-cooker. I'll add an issue with the other things to fix.

PR #1045, opened 2017-08-12
chore(release): add website deploy to master script

christianalfoni:
Makes master branch also run the website publishing script

### DIFF
--- a/scripts/release/index.js
+++ b/scripts/release/index.js
@@ -38,6 +38,7 @@ cooker.cook('publish', [
       cook.pushTagToRemote,
       cook.createReleaseNotes(releaseNotesTemplate),
       cook.createGithubRelease,
+      cook.runNpmScript('deploy'),
     ],
     otherwise: [],
   },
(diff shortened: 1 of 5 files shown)


scripts/release/index.js
```diff
@@ -38,6 +38,7 @@ cooker.cook('publish', [
       cook.pushTagToRemote,
       cook.createReleaseNotes(releaseNotesTemplate),
       cook.createGithubRelease,
+      cook.runNpmScript('deploy'),
```
Guria: Have you tested it? I am not sure this will work. At least website dependencies is not installed automatically

christianalfoni: I have no idea how it works. You did the script and @gaspard did the new setup stuff :) @gaspard , you have any comments here?

gaspard: I think we need to install the devDependencies required for all this "shx", "run-p", "run-s", "gh-pages". Should be tried locally first with "npm run deploy" (eventually disconnecting network to test this)...

We could list the devDeps in "website/package.json" and this should update the main package.json file...

christianalfoni: cc @Guria , got any comments on this? Not sure how far I will get offline and do not understand how this script ever worked without having the dependencies?

gaspard:
OK, I updated the deps for this and it should work now.

PR #1047, opened 2017-08-15
add missing arg to Controller ts declaration

nd0ut:


Guria:
@nd0ut please consider adding suggested test as separate pr

PR #1049, opened 2017-08-22
feat(Devtools): allow mutation, rather document bad external mutations

christianalfoni:
Due to challenges with development vs production I suggest we remove the freezing feature. Currently it copies and freezes state when you extract it in actions. That is great, but you could easily insert this copied frozen state into the state tree again... which is really bad.

I believe it is better to allow mutations outside of the state API, but document that it will not cause a render and the debugger will not pick it up. This is just like Redux. It should be a discipline and you will never get different experience in production.


sladiri:
Would it be sensible to use a Proxy for state during development? Maybe it could log a warning if someone mutates the object properties directly?

PR #1050, opened 2017-08-23
chore(monorepo): use the latest repo-cooker for now

gaspard:
This will allow us to not change Cerebral sources if we need to fix repo-cooker.


gaspard:
It prevents us from having to change monorepo each time we fix a bug in repo-cooker. I propose to use this for now but use a fixed version as soon as we have created one or two releases...

PR #1052, opened 2017-08-25
Added view library for angular 2/4

andrewvmail:
This will allow angular 2/4 to use cerebral as their state management. We are currently using it in a Cordova Ionic app.


packages/node_modules/cerebral/.babelrc
```diff
@@ -1,6 +1,6 @@
 {
   "presets": ["es2015", "react"],
-  "plugins": ["version-transform"],
+  "plugins": ["version-transform", "transform-decorators-legacy"],
```
henri-hulski: `transform-decorators-legacy` has to be added to devDependencies. 
At the moment we get the following error on Travis :
```
ReferenceError: Unknown plugin "transform-decorators-legacy" specified in "/home/travis/build/cerebral/cerebral/packages/node_modules/cerebral/.babelrc" at 1, attempted to resolve relative to "/home/travis/build/cerebral/cerebral/packages/node_modules/cerebral"
```

westcoastavgallery: Okay added missing devDepencies, I missed committing that.

andrewvmail: Okay @henri-hulski I added the missing deps i forgot to commit that earlier. not sure on the eslint warning i didn't really change anything on that front.

henri-hulski: Now we get
```
Error: Cannot find module 'core-js/es7/reflect'
```
and also some prettier linting errors.

docs/api/components.md
```diff
@@ -292,6 +292,78 @@ export default connect({
...
+@Component({ 
+  selector: 'app-component',
+  template: `
+    <div (click)="onClick()">{{foo}}</div>
+  `,
+  changeDetection: ChangeDetectionStrategy.OnPush // disable angular change detection
+})
+@Cerebral({
+  myName: state(['foo']),
+  onClick: signal(['mySignal'])
```
gaspard: What is the reason to have an exceptional way of using tags for this API ? Why not use tagged templates ?

```js
myName: state`foo`,
onClick: signal`mySignal`,
```

packages/node_modules/cerebral/tags.d.ts
```diff
@@ -18,7 +18,7 @@ export class Tag {
...
 
-export function props(path: TemplateStringsArray, ...values: any[]): Tag
-export function signal(path: TemplateStringsArray, ...values: any[]): Tag
-export function state(path: TemplateStringsArray, ...values: any[]): Tag
-export function string(path: TemplateStringsArray, ...values: any[]): Tag
\ No newline at end of file
+export function props(path: any[], ...values: any[]): Tag
+export function signal(path: any[], ...values: any[]): Tag
+export function state(path: any[], ...values: any[]): Tag
+export function string(path: any[], ...values: any[]): Tag
```
gaspard: What is wrong with TemplateStringsArray ? I don't see why adding an API should revert TS types here.

packages/node_modules/cerebral/test/setup.js
```diff
@@ -1,5 +1,24 @@
...
 
+require('core-js/es6')
+require('core-js/es7/reflect')
+require('zone.js/dist/zone-node')
+require('zone.js/dist/long-stack-trace-zone')
+require('zone.js/dist/proxy')
+require('zone.js/dist/sync-test')
+require('zone.js/dist/async-test')
+require('zone.js/dist/fake-async-test')
+require('zone.js/dist/proxy')
```
gaspard: zone.js patches a lot of async methods. Not sure I like the fact that all testing for cerebral is now tinted by this.

If we need such deep customisation of the testing environment for Angular, I propose to simply create a new package for angular. The only change for the user is this:

```ts
import { connect } from '@cerebral/angular'
// instead of
import { connect } from 'cerebral/angular'
```

Maybe all views should live in their own packages. '@cerebral/react', '@cerebral/xxx'. @christianalfoni what do you think ?

packages/node_modules/cerebral/tags.d.ts
andrewvmail: I reverted back to TemplateStringsArray and updated docs based on comment on Discord.

packages/node_modules/cerebral/.babelrc
andrewvmail: added devDeps for angular. should be good to run now.

packages/node_modules/cerebral/test/setup.js
henri-hulski: The views are now moved to "@cerebral/". 
So Angular should also move there in this PR.

andrewvmail: Sounds good will work on that.

andrewvmail: I moved the angular views into @cerebral now

henri-hulski:
@andrewvmail Seems like there was something wrong with the rebase. The PR should contain only your commits. 
Maybe you did a merge instead. 
The right way is to do from PR branch 
```sh
git rebase next
```
And after fixing conflicts 
```sh
git push --force
```

PR #1055, opened 2017-08-28
TypeScript enhancements: ready

gaspard:
- [x] wait for #1063 to be in @next and then rebase
- [x] Types for all "state" methods
- [x] Types for all operators (if not already there)
- [x] Tests involving all "state" methods
- [x] Tests involving all operators


gaspard:
@nd0ut If you can test this branch before we pull it in, it would be great :-) !

PR #1059, opened 2017-08-30
feat(firebase): add typescript definitions, add payload to put

gaspard:
- [x] TypeScript types and tests for all methods in provider
- [x] TypeScript types and tests for all operators
- [x] Add `payload` to put action/operator for progress callback
- [x] Small documentation enhancements
- [x] Export `deleteFile` instead of the potentially illegal keyword `delete` (old export still works)


gaspard:
Depends on #1055

PR #1063, opened 2017-08-31
refactor(Views): move react to @cerebral/react

gaspard:
Tested with my app and works fine. Haven't been able to test legacy import since I am using typescript and "cerebral/react" doesn't have types now.


Guria:
I wonder if reexport in d.ts will work.

PR #1067, opened 2017-09-01
refactor(Views): move views to @cerebral

henri-hulski:


packages/node_modules/cerebral/package.json
```diff
@@ -19,7 +19,12 @@
   "homepage": "http://cerebral.github.io/cerebral-website",
   "dependencies": {
     "eventemitter3": "^2.0.2",
-    "function-tree": "next"
+    "function-tree": "next",
+    "@cerebral/react": "next",
+    "@cerebral/inferno": "next",
+    "@cerebral/preact": "next",
+    "@cerebral/vue": "next",
+    "@cerebral/angularjs": "next"
```
christianalfoni: Should not be dependencies? The views are optional... might not even add a view :-)

henri-hulski: I thing we need these imports so we can reexport them while the old imports are deprecated.
They will be removed when deprecation will be removed.

packages/node_modules/@cerebral/inferno/README.md
```diff
@@ -0,0 +1,55 @@
+# @cerebral/inferno
+
+[Inferno](http://infernojs.org) view for cerebral.
+
+**NPM**
+
+`npm install @cerebral/inferno inferno inferno-component inferno-create-element babel-plugin-inferno`
+
+```js
+import Inferno from 'react'
```
edgesoft: Is this correct?

henri-hulski: I just copied it from docs/api/components.md, but you're right it doesn't look correct.

PR #1071, opened 2017-09-04
Baobab

christianalfoni:
- Refactored to `BaseModel`, allowing the state and state changing methods to be altered
- Added @garth Baobab model implementation
- Fixed tests
- Added docs


packages/node_modules/@cerebral/baobab/src/index.js
```diff
@@ -0,0 +1,85 @@
...
+        values.forEach(value => {
+          this.state.merge(path, value)
+          for (let prop in value) {
+            this.changedPaths.push({
+              forceChildPathUpdates: true,
+              path: path.concat(prop),
+            })
+          }
+        })
+      }
```
garth: With merge, had to add a check if value exists to mimmic behaviour of standard model

```js
merge(path, ...values) {
    if (!this.state.exists(path)) {
      this.state.set(path, {})
    }
    values.forEach(value => {
      this.state.merge(path, value)
      for (let prop in value) {
        this.changedPaths.push({
          forceChildPathUpdates: true,
          path: path.concat(prop),
        })
      }
    })
  }
```

christianalfoni: Okay, updated :)

PR #1072, opened 2017-09-04
docs(website): add views docs and typescript highlighter

christianalfoni:


packages/website/builder/config.json
```diff
@@ -40,10 +40,25 @@
...
+    }, {
+      "name": "inferno",
+      "path": "../../node_modules/@cerebral/inferno/README"
+    }, {
+      "name": "preact",
+      "path": "../../node_modules/@cerebral/preact/README"
+    }, {
+      "name": "angularjs",
+      "path": "../../node_modules/@cerebral/angularjs/README"
+    }],
```
gaspard: We need to keep these sorted to help with navigation... Order is:

```
angularjs
inferno
preact
react
vue
```

packages/website/builder/config.json
```diff
@@ -40,10 +40,25 @@
...
     "addons": [{
       "name": "router",
       "path": "../../node_modules/@cerebral/router/README"
```
gaspard: Could also sort these ones while we are here...

```
firebase
firebase-admin
forms
http
router
shortcuts
storage
useragent
```

packages/website/builder/config.json
christianalfoni: Agreed, updated :)

packages/website/builder/config.json
christianalfoni: Agreed, updated :) Removed `function-tree`, I think it deserves its own "thing". Not sure what, will think on it

christianalfoni:
Okay, summary.

I changed to alphabetical order on views, but have not changed addons yet. I will make separate PR where I make everything alphabetical on API, also changing how collapsing/expanding works

PR #1074, opened 2017-09-06
feat(angular) add view library for angular 2/4

gaspard:
Superseeds #1052


christianalfoni:
Okay, so I made some changes.

A couple of questions:

1. Do I really have to add angular tests to the eslint ignore file? It does not work to add ignore to the file itself for some reason

2. Locally I got failing tests on Cerebral and React... no idea why

3. Now that TypeScript handles template literals... any reason to have support for the array version?

And yeah... I messed up the PR with not pushing the rebase before adding fixes :(

henri-hulski:
To 2. I also get often some failing tests normally related to some timeouts.
Don't know exactly what's going on, but when trying several times either all tests passes or at least some other tests fails. Though should actually be fixed. Maybe add some more time for tests somewhere.

andrewvmail:
Hi @christianalfoni, I just used the new codes and its working in my project. For number 3 angular ahead of time compiler will spit this out "Error: Error encountered resolving symbol values statically. Expression form not supported" if use template literal syntax ... and in some framework like Ionic aot is enabled by default when building for production (disabling - https://github.com/ionic-team/ionic-app-scripts/issues/145) ... 

One other thing i just recently found out is to make the project compile with aot you also need these added (see snippet below) into the view controller class body... im not sure if creating a cerebral base class to extend is a good idea here since all view controllers will have those 3 lines of boilercode in the body..
```
ngOnDestroy() {}
ngOnInit() {}
ngAfterViewInit() {}
```
see ticket https://github.com/angular/angular/issues/16023

christianalfoni:
Hi @andrewvmail and thanks for the comments!

Yeah, extending here seems like a good idea cause then it will be easier for us to ensure that both controller and cdr is passed into `super()` call.

Let me make an update for this and see what you think :-)

packages/node_modules/@cerebral/angular/package.json
```diff
@@ -14,6 +14,7 @@
   },
   "homepage": "http://cerebral.github.io/cerebral-website",
   "dependencies": {
+    "@types/mocha": "^2.2.43",
```
henri-hulski: Are mocha types still necessary after removing the tests?

andrewvmail: working on adding the test back now

PR #1079, opened 2017-09-08
chore(release): change dependency on cerebral from normal to peer

gaspard:
We do not want to have circular depencencies because this prevents release with repo-cooker.

### DIFF
--- a/packages/node_modules/@cerebral/http/package.json
+++ b/packages/node_modules/@cerebral/http/package.json
@@ -23,14 +23,11 @@
     "url": "https://github.com/cerebral/cerebral/issues"
   },
   "homepage": "https://github.com/cerebral/cerebral/tree/master/packages/http#readme",
-  "dependencies": {
-    "function-tree": "^1.0.0-beta.3"
-  },
   "peerDependencies": {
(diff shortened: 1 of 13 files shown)


packages/node_modules/@cerebral/http/package.json
```diff
@@ -24,13 +24,13 @@
   },
   "homepage": "https://github.com/cerebral/cerebral/tree/master/packages/http#readme",
   "dependencies": {
-    "function-tree": "^1.0.0-beta.3"
+    "function-tree": "next"
```
henri-hulski: Why we need here function-tree as dependency when cerebral has it already?

packages/node_modules/@cerebral/storage/package.json
```diff
@@ -23,10 +23,10 @@
     "url": "https://github.com/cerebral/cerebral/issues"
   },
   "homepage": "https://github.com/cerebral/cerebral/tree/master/packages/storage#readme",
-  "peerDependencies": {
-    "cerebral": "next"
-  },
   "devDependencies": {
     "cerebral": "^2.0.0-beta.3"
```
henri-hulski: This should be next.

PR #1086, opened 2017-09-10
Added back tests for Angular view using Typescript

andrewvmail:


christianalfoni:
Great stuff @andrewvmail , though did you try to rather `extend` the component class? And just do normal injection as in the docs? I think that should work now? Do not have to do the hack :)

andrewvmail:
@christianalfoni Just tried it yes the extend work looks good and I have a test running with it successfully although the test has to be mocked up a bit because of some quirks. Not sure what to do with the overall mono repo dependencies system that the travis check relies so the build keeps failing. I notice some times on my computer some artifact is left in dist directory and the subsequent test fails until i deleted that dist folder. I'm not sure what's causing that right now.... but in case you run into that, check if the dist folder exist.

christianalfoni:
Okay, merged it in to keep working on it :)

PR #1090, opened 2017-09-11
Fix release with new repo-cooker, fixes view error message, change logo size

gaspard:
The new repo-cooker now supports circular dependencies (Yeah !) so we should not filter out dependencies because this leads to invalid "@next" releases (peer dependency mess or worse multiple cerebral versions installed and breaking apps).

Made the log 25vh instead of 50vh because it is really too big now and pushes the important links way down....

Please let me pull this PR in when it is approved (need to bust cache for repo-cooker).

### DIFF
--- a/packages/node_modules/cerebral/src/views/View.js
+++ b/packages/node_modules/cerebral/src/views/View.js
@@ -271,9 +271,7 @@ class View {
       }
 
       if (!this.dependencies[propKey].getTags) {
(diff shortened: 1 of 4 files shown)


christianalfoni:
Is it something wrong with the test here?

gaspard:
@christianalfoni OK, fixed typo.

PR #1097, opened 2017-09-12
Remove deprecations

christianalfoni:


henri-hulski:
I think #992 should either go in here or be fixed by himself.

packages/node_modules/cerebral/src/test/index.test.js
```diff
@@ -53,114 +53,6 @@ describe('test helpers', () => {
     })
   })
 
-  describe('runSignal', () => {
```
henri-hulski: I think the test for `runSignal` should not be removed. Only for `RunSignal factory`.

henri-hulski: Ok I have fixed that.

PR #1109, opened 2017-09-15
Refactor to new @cerebral/react package imports

christianalfoni:


packages/demos/demo/package.json
```diff
@@ -11,15 +11,15 @@
...
   ],
   "devDependencies": {
-    "node-sass": "^4.5.3",
-    "extract-text-webpack-plugin": "^2.1.0"
+    "extract-text-webpack-plugin": "^2.1.0",
+    "node-sass": "^4.5.3"
   },
   "dependencies": {
-    "bulma": "^0.2.3",
-    "cerebral": "next",
```
gaspard: Since we directly import from 'cerebral' (operators, tags, controller), I find it a little weird that we no longer have it as dependency. But this is maybe just a pattern I must get used to.

PR #1110, opened 2017-09-16
Fixes firebase & some types

gaspard:
**firebase**

- Fixed a bug in firebase preventing proper error reporting.
- Fixed `off` method when it is not used with wildcard.

**cerebral**

- Fixed `resolve` type in cerebral (was missing optional overrideProps).


gaspard:
Had to rebase on top of latest changes. @christianalfoni please review again...

PR #1121, opened 2017-09-20
Move tags and resolve provider to function-tree

christianalfoni:
This is what I have done:

- function-tree now has the `ResolveProvider` and all logic related to tags
- Cerebral proxies these exports and adds some custom ones
- added new class `TagValue`. This is a class that identifies that a value can be extracted in a template tag. Now `Compute` extends `TagValue`... meaning Compute is no longer "a thing" related to resolver or tags, it has become generic. Also `Tag` extends `TagValue`.
- Removed custom logic in tags for how to extract things from context. Now this is defined when the tag is created:

```js
const props = createTemplateTag('state', (path, context) => {
  return context.state.get(path)
})
```
- Deprecated `context.resolve.isCompute`, now use `context.resolve.isTagValue`
- `overrideProps` is noe `overrideContext`. Meaning:

```js
function someAction ({resolve}) {
  resolve.value(someTag, {
    props: {foo: 'bar'}
  })
}
```
This was necessary because of the new generic nature of tags and resolver. Will allow for overriding other stuff as well.
- Removed options from tags. `isStateDependency` was not necessary, could just check for tag type of "state". Also `hasNoValue` option was no longer necessary as how value is extracted is now defined with the definition of a tag.

This PR needs to be tested in a real project, but all tests are running green at least.


Guria:
@christianalfoni it seems it is good time to provide convertObjectWithTemplates functionality in ResolveProvider:

```
Searching for convertObjectWithTemplates in/packages/node_modules (Found 34 matches in 18 files)

/packages/node_modules/@cerebral/firebase/src/helpers.js:
	175: export function convertObjectWithTemplates(obj, resolve) {

/packages/node_modules/@cerebral/firebase/src/operators/onChildAdded.js:
	1: import { convertObjectWithTemplates } from '../helpers'
	8:       convertObjectWithTemplates(options, resolve)

/packages/node_modules/@cerebral/firebase/src/operators/onChildChanged.js:
	1: import { convertObjectWithTemplates } from '../helpers'
	8:       convertObjectWithTemplates(options, resolve)

/packages/node_modules/@cerebral/firebase/src/operators/onChildRemoved.js:
	1: import { convertObjectWithTemplates } from '../helpers'
	8:       convertObjectWithTemplates(options, resolve)

/packages/node_modules/@cerebral/firebase/src/operators/onValue.js:
	1: import { convertObjectWithTemplates } from '../helpers'
	8:       convertObjectWithTemplates(options, resolve)

/packages/node_modules/@cerebral/firebase/src/operators/signInWithFacebook.js:
	1: import { createReturnPromise, convertObjectWithTemplates } from '../helpers'
	6:       firebase.signInWithFacebook(convertObjectWithTemplates(options, resolve)),

/packages/node_modules/@cerebral/firebase/src/operators/signInWithGithub.js:
	1: import { createReturnPromise, convertObjectWithTemplates } from '../helpers'
	6:       firebase.signInWithGithub(convertObjectWithTemplates(options, resolve)),

/packages/node_modules/@cerebral/firebase/src/operators/signInWithGoogle.js:
	1: import { createReturnPromise, convertObjectWithTemplates } from '../helpers'
	6:       firebase.signInWithGoogle(convertObjectWithTemplates(options, resolve)),

/packages/node_modules/@cerebral/firebase/src/operators/task.js:
	1: import { createReturnPromise, convertObjectWithTemplates } from '../helpers'
	8:         convertObjectWithTemplates(payload, resolve)

/packages/node_modules/@cerebral/firebase/src/operators/update.js:
	1: import { createReturnPromise, convertObjectWithTemplates } from '../helpers'
	6:       firebase.update(convertObjectWithTemplates(updates, resolve)),

/packages/node_modules/@cerebral/firebase/src/operators/value.js:
	1: import { convertObjectWithTemplates, createReturnPromise } from '../helpers'
	8:         convertObjectWithTemplates(options, resolve)

/packages/node_modules/@cerebral/http/src/utils.js:
	70: export function convertObjectWithTemplates(obj, resolve) {

/packages/node_modules/@cerebral/http/src/operators/httpDelete.js:
	1: import { convertObjectWithTemplates, processResponse } from '../utils'
	8:         convertObjectWithTemplates(query, resolve)

/packages/node_modules/@cerebral/http/src/operators/httpGet.js:
	1: import { convertObjectWithTemplates, processResponse } from '../utils'
	6:       http.get(resolve.value(url), convertObjectWithTemplates(query, resolve)),

/packages/node_modules/@cerebral/http/src/operators/httpPatch.js:
	1: import { convertObjectWithTemplates, processResponse } from '../utils'
	6:       http.patch(resolve.value(url), convertObjectWithTemplates(body, resolve)),

/packages/node_modules/@cerebral/http/src/operators/httpPost.js:
	1: import { convertObjectWithTemplates, processResponse } from '../utils'
	6:       http.post(resolve.value(url), convertObjectWithTemplates(body, resolve)),

/packages/node_modules/@cerebral/http/src/operators/httpPut.js:
	1: import { convertObjectWithTemplates, processResponse } from '../utils'
	6:       http.put(resolve.value(url), convertObjectWithTemplates(body, resolve)),

/packages/node_modules/@cerebral/http/src/operators/httpUploadFile.js:
	1: import { convertObjectWithTemplates, processResponse } from '../utils'
	7:     const options = convertObjectWithTemplates(optionsValue, resolve)

Found 34 matches in 18 files 
```

christianalfoni:
Yes indeed @Guria, let me fix that asap :)

christianalfoni:
@Guria Okay, this should be part of the commit now :-)

packages/node_modules/function-tree/README.md
```diff
@@ -341,6 +341,121 @@ const ft = new FunctionTree([
...
+```js
+import {string, props} from 'function-tree/tags'
+
+export default [
+  shout(string`Hello ${props`name`}`)
+]
+```
+
+#### createTemplateTag
+Allows you to add new tags, targeting your custom providers.
```
gaspard: This is really cool 👍  ! And API is super simple. Well done.

packages/node_modules/function-tree/README.md
```diff
@@ -341,6 +341,121 @@ const ft = new FunctionTree([
...
+
+```js
+import {createTemplateTag} from 'function-tree/tags'
+
+export const myTag = createTemplateTag('myTag', (path, context) => {
+  return context.myProvider.get(path)
+})
+```
```
gaspard: Maybe add usage ?

```js
someChain = [
  set(state`some.path`, myTag`other.path`)
]
```

packages/node_modules/function-tree/README.md
```diff
@@ -341,6 +341,121 @@ const ft = new FunctionTree([
...
+export default (tag) => {
+  return new Uppercaser(tag)
+}
+```
+
+```js
+import uppercaser from './uppercaser'
+
+export default [
+  shout(string`Hello, ${uppercaser(props`name`)}`)
```
gaspard: Not sure I understand here. Is `uppercaser` a tag now ? Or is it just a wrapper around another tag ?

packages/node_modules/function-tree/src/tags/index.js
```diff
@@ -0,0 +1,46 @@
...
+
+    return convertedObject
+  }, {})
+}
+
+export function extractValueWithPath(obj, path) {
+  return path.split('.').reduce((currentValue, key, index) => {
+    if (index > 0 && currentValue === undefined) {
+      throw new Error(
+        `A tag is extracting with path "${path}", but it is not valid`
```
gaspard: Error messages are phrases. Needs a `.` at the end (same for other error messages below)...

packages/node_modules/function-tree/README.md
gaspard: How. OK. TagValue is just a wrapper around `getValue`. I tried to implement such dynamic values by creating custom tags but it was a mess. Very cool idea to make this more accessible.

But I think the name `TagValue` is misleading: it has nothing to do with tagged templates. I would call this `DynamicValue` because this is what it gives us: a value that is resolved when signal is executed.

Here are two examples that I stumbled on and for which I wanted such a thing:

```js
class Now extends DynamicValue {
  getValue() {
    return Date.now()
  }
}
export const now = new Now()

class Uuid extends DynamicValue {
  getValue() {
    return uuid.v4()
  }
}
export const uuid = new Uuid()

const mychain = [
  set(props`now`, now),
  set(props`uid`, uuid)
]
```

Also I think there is no need to call `super` in examples...

packages/node_modules/function-tree/src/providers/Resolve.js
```diff
@@ -0,0 +1,59 @@
...
+
+      return arg
+    },
+    path(arg) {
+      if (arg instanceof Tag) {
+        return arg.getPath(context)
+      }
+
+      throw new Error(
+        'You are extracting a path from an argument that is not a Tag'
```
gaspard: Same here and with other messages: missing `.` at end of error phrase...

Guria:
Commit message should list all breaking changes there. Including changed imports, a way to create own tags, etc

christianalfoni:
@gaspard Gone through and updated with your comments, thanks a bunch :)

Will now split up into multiple commits

christianalfoni:
Hold this PR until next is releasef in master

gaspard:
PS: I did not export `ContextualValue` from `cerebral/tags` as I consider these re-exports a deprecation thing and since `ContextualValue` is new it should only be exported from function-tree I think.

gaspard:
OK, refactored to use `ResolveValue` and `resolveObject`. Added in operators when they detect objects with ResolveValues.

Guria:
@gaspard Great job done here

What about breakings list? I suppose the way to define own tag is changed dramatically at least. 

Which packages major versions would be bumped?

gaspard:
@Guria No, I don't think we have breaking changes here. People used to create tags by extending the Tag class and this still works. @christianalfoni might have a better idea on breaking changes.

If cerebral has a breaking change, all packages which depend on 'cerebral' will bump major version (= probably all).

Guria:
> If cerebral has a breaking change, all packages which depend on 'cerebral' will bump major version (= probably all).

It seems a bit unnecessary to bump every package on it. I would expect that untouched packages wouldn't published at all in such case.

christianalfoni:
@Guria Problem is that we do not know which packages are not affected by the breaking change. We just know that "something breaks", we have no idea if it actually breaks any of the others. So it is just to be on the safe side.

But I will set up a hangout so we can discuss versioning strategy. Currently it is confusing for people, cause the packages commited to does not reflect the version bumping.

At least now we have some experience and can move on :)

There should not be any breaking change here

PR #1126, opened 2017-09-21
Next.js support

christianalfoni:
This is basically how it looks in `next.js`:

```js
import React from 'react'
import {Controller, UniversalController} from 'cerebral'
import Devtools from 'cerebral/devtools'
import {Container} from '@cerebral/react'
import modules from '../modules'
import Page from '../components/Page'

export default class extends React.Component {
  static async getInitialProps({pathname}) {
    const controller = UniversalController({modules})

    return controller.runSequence([
      function setPage ({state}) {
        state.set('app.currentPage', 'home')
      }
    ])
      .then(() => {
        return {
          stateChanges: controller.getChanges()
        }
      })
  }
  render() {
    return (
      <Container controller={Controller({
        devtools: typeof window === 'undefined' ? null : Devtools({host: 'localhost:8787'}),
        stateChanges: this.props.stateChanges,
        modules
      })}>
        <Page />
      </Container>
    )
  }
}
```

### DIFF
--- a/docs/api/controller.md
+++ b/docs/api/controller.md
@@ -11,7 +11,11 @@ const controller = Controller({
   signals: {},
 
   // Defines the top level modules
-  modules: {}
+  modules: {},
+
+  // A map of state changes to run before instantiation,
+  // where the key is the path and value is the state value
+  stateChanges: {}
 })
 
 export default controller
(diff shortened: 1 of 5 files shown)


MartanLV:
haha just found this! A bit too late https://github.com/zeit/next.js/pull/2970

PR #1127, opened 2017-09-21
feat(storage): add support for async local storage

JohannesAnd:


christianalfoni:
This can actually be handled without an option, created this PR: https://github.com/cerebral/cerebral/pull/1134. Thanks for inspiration @JohannesAnd :)

PR #1113, opened 2017-09-16
Turn storage provider into a module + api to re-hydrate upon refresh

andrewvmail:
An improvement to avoid getter setter ```localStorage``` boilercodes in the controller options. 

Referencing #1064

PR #1134, opened 2017-09-24
add async storage handling

christianalfoni:
BREAKING CHANGE:

remove does no longer return anything


Guria:
probably should be combined with #1113

christianalfoni:
Oh yes, good point! I will do this!

PR #1136, opened 2017-09-25
feat(http): pass through options from operators

Guria:
`@cerebral/http` operators now accept and pass through options.

### DIFF
--- a/packages/node_modules/@cerebral/http/README.md
+++ b/packages/node_modules/@cerebral/http/README.md
@@ -175,7 +175,7 @@ function someDeleteAction ({http}) {
 
 ### operator
 ```js
-import {httpPost} from '@cerebral/http/operators'
+import {httpDelete} from '@cerebral/http/operators'
 import {state} from 'cerebral/tags'
 
 export default [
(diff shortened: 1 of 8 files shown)


christianalfoni:
Great stuff! Though would suggest `options` should also resolve as object template. Cause you might want to pass in options using tags :) But other than that, looks great!

Guria:
@christianalfoni not sure about it. And I also have a personal need to pass there Symbol as key. It will not survive after resolving.

christianalfoni:
@Guria It only resolves when there is actual a `ResolveValue` instance, so if you are passing in a Symbol there it will stay:

```js
httpGet('/user', {
  foo: Symbol,
  bar: props`bar`
}) 
```

Only `bar` will actually be resolved, while Symbol stays the same :)

Guria:
I have used Symbol as key. And it was lost after passing resolve function.

christianalfoni:
Then that is a bug @Guria, resolvers should only touch `ResolveValue`types. Let us fix this when we got the new `ResolveValue` in there so we do not have to fix it twice :)

Guria:
@christianalfoni that's not a bug as I can understand. It's intended omit symbol keys when object keys copied.

Guria:
@christianalfoni let's merge it in as is without resolving options object.

PR #1137, opened 2017-09-25
test(Controller): failing test for Error constructor other then native

Guria:


christianalfoni:
What is this? Is it TypeScript that can not understand `extend` of Error?

Guria:
I don't know :(
I haven't figured it out yet

gaspard:
`ErrorConstructor` has no link to `Error` as strange as this looks. So TypeScript has no way of knowing that a CustomError extends Error.

Maybe stackoverflow knows... https://stackoverflow.com/questions/46426645/how-to-use-typescript-map-with-error-handlers

gaspard:
I fixed the type. But there is a gotcha when creating the map (see test).

packages/node_modules/cerebral/test/typescript/controller.ts
```diff
@@ -30,7 +30,10 @@ const controller: ControllerClass = Controller({
     module1,
     module2
   },
-  catch: new Map([
+  // We have to type map here or it will try to find a common
+  // interface for keys and none exists.
+  catch: new Map<CatchErrorConstructor, SignalChain>([
```
Guria: @gaspard I suppose now we can remove type here. right?

gaspard: No, we cannot remove the type because TS will try to guess the common type of keys and will fail due to current TS limitations...

gaspard: image

gaspard: But maybe we can find a fix. Let me try.

gaspard: The issue is TS trying to find a common constructor signature and they are all different so it complains.

Guria: @gaspard. hmm. but it works on my side.
image
image

no complies, compiled without errors

gaspard: It works because you have only one key in the Map (HttpProviderError). As soon as you have more then one, TS tries to resolve the constructor signature and fails.

gaspard:
I rebased. This is OK for now.

PR #1139, opened 2017-09-25
WIP: RFC: alternative way to expose services with providers

Guria:
@cerebral/collaborators request for comments

### DIFF
--- a/packages/node_modules/@cerebral/http/src/HttpService.js
+++ b/packages/node_modules/@cerebral/http/src/HttpService.js
@@ -1,0 +1,123 @@
+import request from './request'
+import { urlEncode, mergeWith, createResponse } from './utils'
+import FileUpload from './fileUpload'
+import DEFAULT_OPTIONS from './DEFAULT_OPTIONS'
+
+export default class HttpService {
+  constructor(controller, passedOptions) {
+    this.controller = controller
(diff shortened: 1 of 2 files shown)


christianalfoni:
Hm... not sure about this as we already have `provide`?

```js
provide('http', {

})
```

It caches and adds debugger and all... Anything more low level than that I do not think we should be opinionated about. Could maybe allow:

```js
provide('http', (context) => {
  return {} // Actual API returned
})
```

Which also caches, the returned result, but then you also have access to the context in those methods

Guria:
Hmm. I have checked `provide` again. It doesn't any caching. It assigns and wraps object on every single action execution.

christianalfoni:
Yeah, there is no reason to cache on `provide` because you return the object that will always be used.

You can not do debugger wrap just once, because the wrapper needs information about the execution, to identify what action triggered the side effect. So it needs a fresh `context` every time.

I guess we could do it a smarter way though. So it wraps once, but the `context` it points to is rather swapped out. But that could cause weird issue with asynchronicity. There might be two actions running the same side effect, if we have only one `context` available at any time the debugger could show the wrong information.

Have to think more on this, but yeah... I think the suggested callback version where `context` is provided would eliminate the need for the low level one... at least as public API.

PR #1152, opened 2017-10-04
fix(storage): Setting key value to undefined should unset key

garth:
When setting a value to undefined it gets serialised to 'undefined' string, then when parsing the storage value back to json later it fails since 'undefined' is truthy.

Instead it should not be possible to set undefined, setting undefined now removes the key.

This also fixes sync with unset state values.


garth:
Here's a temporary fix until the PR is merged and released, just add the following to your html:

```html
<iframe srcdoc="<html><head><script>
  window.addEventListener('storage', e => e.newValue === 'undefined' && window.localStorage.removeItem(e.key), false);
</script></head></html>"></iframe>
```

PR #1155, opened 2017-10-05
Add 'signals' tag to target all the signals of the specified module

nd0ut:
Hello! I thought it was a good idea to allow all module signals to be pushed right into the view without needing to specify each signal separately.

Example of usage:
```
@connect({
  moduleSignals: signals`someModule`
})
...
props.moduleSignals.moduleSignal(payload)
```

What do you think about it?


gaspard:
@nd0ut Please redo the implementation on top of current "next" branch. I tried to rebase but too many things have moved and it is simpler to just recreate the change, copy-pasting code for "signal" to "signals".

PR #1156, opened 2017-10-06
Include addRoutes method on router module

garth:
Adds the option to add new routes after the router module has been initialised. Useful for lazy loading app parts and for app security.

Not tested, just a concept!!!

### DIFF
--- a/packages/node_modules/@cerebral/router/README.md
+++ b/packages/node_modules/@cerebral/router/README.md
@@ -340,3 +340,40 @@ const controller = Controller({
   }
 })
 ```
+
+#### Dynamically add routes
+
+Apps that use code splitting for security reasons may not want to define all routes until after the user has been verified.
+
+The `addRoutes()` method allows routes to be added after the app has been initialized.
+
+Define your initial set of routes as normal:
+```js
+import Router from '@cerebral/router';
+
+export default Router({
+  routes: [
+    { path: '/', signal: 'homeRouted' },
+    { path: '/login', signal: 'loginRouted' }
+  ]
+});
+```
+
+Then later you can call `addRoutes`:
+```js
+import router from './router';
+
+router.addRoutes([
+  { path: '/now-you-see-me', signal: 'hiddenModule.hiddenRouted' }
+]);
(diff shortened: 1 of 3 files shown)


Guria:
Yeah we definitely need some ways to do that. Would be great to see some usage examples at least.

garth:
I'm currently hacking up a working example, so should be able to update the PR with working code and a sample.

garth:
OK, I've updated the code with something that works now (tested in my app with a custom router).

Usage is very simple, define your initial set of routes as normal:
```js
import Router from '@cerebral/router';

export default Router({
  routes: [
    { path: '/', signal: 'app.homeRouted' },
    { path: '/login', signal: 'app.loginRouted' }
  ]
});
```

Then after successful login you can do something like
```js
export default async function loadSecretApp({ state }) {
    // let webpack know to make a separate chunk with the secret stuff - be sure to
    // secure this static asset to only allow authenticated clients to access
    await import('../secretAppEntryPoint');
  }
}
```

and finally within secretAppEntryPoint.js add your additional modules and routes
```
import controller from './controller';
import router from './modules/router';
import SomeTopSecretModule from './modules/SomeTopSecretModule';

// add secret modules
controller.addModule('secretModule', SomeTopSecretModule());

// add secret routes
router.addRoutes([{ path: '/now-you-see-me', signal: 'secretModule.hiddenRouted' }]);
```

You also have to import your components in the entry point file, but that's not really a cerebral issue.

The only gotcha is that you might need to refresh the current route when reloading the app. Due to the order of evens, the router will fire (without the secret routes) before the routes and modules have loaded, This can be handled with an action that should be called after the loadSecretApp action
```js
export default function reloadRoute({ router }) {
  const path = document.location.pathname;
  router.redirect(path);
}
```

garth:
@Guria If you are OK with this PR then I will add docs?

Guria:
Interesting challenge with reloading. I am ok with current solution, let's try it and explore more.

garth:
I've added docs to the "Experimental" section of the router docs.

PR #1158, opened 2017-10-09
feat(cerebral): expose the controller in CerebralTest response

garth:
Having direct access to the controller allows you to use it in render tests

```js
    const cerebral = CerebralTest({ modules: { login: LoginModule() } });
    const wrapper = mount(
      <Container controller={cerebral.controller}>
        <Login />
      </Container>
    );
    expect(wrapper.find('.submit-button')).toHaveProp('disabled', false);
```

### DIFF
--- a/docs/introduction/test.md
+++ b/docs/introduction/test.md
@@ -118,13 +118,18 @@ it('should increment numbers in state', () => {
 
 ## Signals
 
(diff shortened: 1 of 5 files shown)


christianalfoni:
https://github.com/cerebral/cerebral/pull/1164

PR #1163, opened 2017-10-14
feat(View): allow props to be a function that gets run executed with ext props

schotime:
This is just a proposal PR that allows dependencies to be passed as a function like we used to.

Allows:

```js
connect(extProps => ({
   name: state`name`,
   location: state`locations.${extProps.locNumber}`
}), Component);
```

Extremely useful for my fully typed implementation that is very close.


schotime:
Closed for now as I have found a very nice workaround.

PR #1166, opened 2017-10-17
fix(function-tree): add support for falsey values in tag path string builder

mrvelic:


mrvelic:
@schotime and I discovered this when passing an array index into a path builder from a map. The zero value was simply removed from the resulting path

PR #1171, opened 2017-10-25
ts: remove deprecated typings lib usage

br1anchen:
Remove unnecessary `typings` dependency based on typings [publishing](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html) for Typescript

### DIFF
--- a/package.json
+++ b/package.json
@@ -151,7 +151,6 @@
     "ts-node": "3.3.0",
     "typescript": "2.5.2",
     "typescript-definition-tester": "0.0.5",
-    "typings": "2.1.1",
     "vue": "2.3.3",
     "webpack": "2.5.1",
     "webpack-dev-server": "2.4.5",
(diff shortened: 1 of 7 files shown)


gaspard:
Hmmm. I think we should remove these fields completely because we only expose part of them, like "tags.d.ts" is not exposed... 

From the TypeScript docs: 

```
Note that the "typings" field is synonymous with "types", and could be used as well.

...
Also note that if your main declaration file is named index.d.ts and lives at the root of the package (next to index.js) you do not need to mark the "types" property, though it is advisable to do so.
```

br1anchen:
@gaspard I reverted changes filed name `typings` to `types` since it is not necessary.
I do think that we should keep `typings` field as type definition file entry point in this case.
typings in `tags.d.ts` should be imported/included in module level `index.d.ts`, 
if some of types are internal types, we should include them as internal types in the same module namespace, in order to make `tsc` to infer the correct type

gaspard:
Actually, `tags.d.ts` is used when we import from `cerebral/tags`. Seems to work fine, so yes, let's keep this so :-)

gaspard:
Thanks for the contribution ! 👍

PR #1173, opened 2017-10-27
ts: props type generic for Controller.getSignal

manneredboor:
Added the generic for make definable type of signal, returning from ControllerClass.getSignal

### DIFF
--- a/packages/node_modules/cerebral/index.d.ts
+++ b/packages/node_modules/cerebral/index.d.ts
@@ -79,7 +79,7 @@ export interface ControllerClass extends FunctionTree {
   getModel(): StateModel
   getState(path?: string): any
   runSignal(name: string, signal: SignalChain, payload: any): void
-  getSignal(path: string): Signal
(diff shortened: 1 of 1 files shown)


packages/node_modules/cerebral/index.d.ts
```diff
@@ -79,7 +79,7 @@ export interface ControllerClass extends FunctionTree {
   getModel(): StateModel
   getState(path?: string): any
   runSignal(name: string, signal: SignalChain, payload: any): void
-  getSignal(path: string): Signal
+  getSignal<T=any>(path: string): Signal<T>
```
Guria: What I don'like here about typing systems like ts and flow is that is could be treated as breaking change :(
But bump major version on each such change is overkill

gaspard:
Nice contribution. @manneredboor Can you rephrase your commit so that it matches the pattern (used when publishing):

```
fix(types): some message here
```

manneredboor:
@gaspard Done!

PR #1175, opened 2017-11-01
fix(website): ensure the nav menu renders nicely on mobile systems

bdjnk:
Some mobile systems evidently don't have fonts which include [⤣](http://www.fileformat.info/info/unicode/char/2923/index.htm).

Additionally, evidently there are bugs ([for example](https://stackoverflow.com/questions/38781089/font-family-monospace-monospace#38781315)) on some systems relating to monospace font rendering regarding line height and other typographical niggles.

This pull request addresses these problems.


packages/website/builder/css/nav.css
```diff
@@ -165,7 +163,7 @@
 
 .nav_toggle,
 .nav_toggle-label::before {
-  font-family: "Courier New", monospace;
+  font-family: monospace, monospace;
```
henri-hulski: I think one `monospace` should be enough.

packages/website/builder/css/nav.css
```diff
@@ -210,13 +208,22 @@
   color: #cacaca;
   position: absolute;
   right: 0;
-  font-family: "Courier New", monospace;
+  font-family: monospace, monospace;
```
henri-hulski: Same here.

packages/website/builder/css/main.css
```diff
@@ -101,3 +101,20 @@ pre > code {
   font-size: 14px;
   line-height: 24px;
 }
+
+@font-face {
+    font-family: 'cerebral icons';
+    src: url('/images/cerebral-icons.woff') format('woff');
```
henri-hulski: Is woff format enough for all supported browsers?

packages/website/builder/css/nav.css
bdjnk: No, there's a bug and this is the hack fix. One monospace does the wrong thing in various browsers which misbehave.

packages/website/builder/css/nav.css
bdjnk: Same here ;)

packages/website/builder/css/main.css
bdjnk: [Yes.](http://caniuse.com/#feat=woff)

packages/website/builder/css/nav.css
henri-hulski: I see.

PR #1178, opened 2017-11-02
use action instead of unprotect/protect

mattiamanzati:
Not sure if it will work, code edited in the browser without testing, but ideally it should :)

### DIFF
--- a/packages/node_modules/@cerebral/mobx-state-tree/src/Model.js
+++ b/packages/node_modules/@cerebral/mobx-state-tree/src/Model.js
@@ -30,7 +30,7 @@ class Model {
         )
       )
       .actions(self => ({
-         changeInBatch(fn){
+         runInBatch(fn){
             fn(self) 
          }
       })
--- a/packages/node_modules/@cerebral/mobx-state-tree/src/Model.js
+++ b/packages/node_modules/@cerebral/mobx-state-tree/src/Model.js
@@ -29,6 +29,11 @@ class Model {
           }
         )
       )
+      .actions(self => ({
+         changeInBatch(fn){
+            fn(self) 
+         }
+      })
       .create(state)
   }
   updateIn(path, cb, forceChildPathUpdates = false) {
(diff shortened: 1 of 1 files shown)


mattiamanzati:
Also note that instead of traversing the model, you could use something like

applyPatch(this.state, {op: "replace", path: "/path/with/slashes/instead/of/dots", value: newValue)

christianalfoni:
Hi man and thanks for the suggestion! :)

I believe this fix will have the same problem cause each individual state change will be treated as running one MST action, meaning each individual state change will protect/unprotect. That said! I think I have a fix:

First of all, whenever a signal execution ends or an async function starts I will protect the MST... 

```js
    controller.on('end', () => protect(this.state))
    controller.on('asyncFunction', () => protect(this.state))
```

Then whenever a Cerebral action does a state change I will `unprotect` it.

Basically this means whenever a state change is made through a cerebral action it will be possible to do so. But only at the end of a signal execution or something async happens in a signal it will protect it again, flushing out changes to components... as I understand it.

Though how does this work in MST? Like:

```js
.actions(self => ({
  doThis() {
    self.foo = 'bar'
    self.bar = 'foo'
    // The two previous changes before components are notified?
    return async(). then(() => {
      self.mip = 'mop'
     // But here? How does MST control when to protect/unprotect?
    })
  }
}))
```

Need to test out my solution first, thanks for suggestions and info! :)

christianalfoni:
Or wait... hm... I think I am a bit confused. Cause now testing this in isolation.

```js
[
  increment, // unprotect/protect
  increment, // unprotect/protect
  increment // unprotect/protect
]
```

These cause one render to happen. Though I have seen scenarios where this is not the case.... you know... let me try your approach here and see how it works :)

christianalfoni:
Okay, I tested this an it works :) It seems like best solution as it will hand off all this logic to mobx-state-tree! Great stuff, thanks!

christianalfoni:
Made some tiny changes: https://github.com/cerebral/cerebral/pull/1179

mattiamanzati:
You're welcome! :D

PR #1179, opened 2017-11-02
fix(mobx): wrap state changes in MST action

christianalfoni:


henri-hulski:
@christianalfoni could you fix the base branch?

PR #1182, opened 2017-11-03
feat(firebase-admin): depend on latest version of firebase-admin

christianalfoni:
BREAKING CHANGE:

Might break where firebase-admin is used directly. Be sure to depend on same version in own

package.json if any

### DIFF
--- a/packages/node_modules/@cerebral/firebase-admin/package.json
+++ b/packages/node_modules/@cerebral/firebase-admin/package.json
@@ -10,10 +10,12 @@
   },
   "dependencies": {
     "faye-websocket": "^0.11.1",
-    "firebase": "^3.9.0",
-    "firebase-admin": "^4.0.4",
-    "firebase-queue": "^1.6.0",
     "proxyquire": "^1.7.11"
+  },
+  "peerDependencies": {
+    "firebase": "^5.4.3",
+    "firebase-admin": "^4.0.4",
+    "firebase-queue": "^1.6.0"
   },
   "directories": {
     "test": "tests"


packages/node_modules/@cerebral/firebase-admin/package.json
```diff
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "faye-websocket": "^0.11.1",
-    "firebase": "^3.9.0",
+    "firebase": "^5.4.3",
```
Guria: Probably worth to put it in peerDeps, so the user controls used versions.

christianalfoni: Yeah, thought about it... but is it correct usage? Cause then we do not control when the API might break due to breaking in firebase. I think this is most correct, but most maintenance of course

Guria: We still control by defining range for what we know works for us

christianalfoni: Hm, I thought peerDependency was when lib did not explicitly require the host library... like middleware stuff... but made the change. Does not matter to me 😄

packages/node_modules/@cerebral/firebase-admin/package.json
```diff
@@ -10,10 +10,12 @@
   },
   "dependencies": {
     "faye-websocket": "^0.11.1",
-    "firebase": "^3.9.0",
+    "proxyquire": "^1.7.11"
+  },
+  "peerDependencies": {
+    "firebase": "^5.4.3",
     "firebase-admin": "^4.0.4",
     "firebase-queue": "^1.6.0",
```
henri-hulski: @christianalfoni you forgot to remove the comma at the end of the line.
Like this Travis is erroring.

henri-hulski:
Also the base branch has to be fixed.

christianalfoni:
This is rather part of: https://github.com/cerebral/cerebral/pull/1198

PR #1190, opened 2017-11-09
feat(firebase-admin): added listUsers

tomhah:
Added listUsers function from firebase-admin and updated the firebase-admin npm package

### DIFF
--- a/packages/node_modules/@cerebral/firebase-admin/README.md
+++ b/packages/node_modules/@cerebral/firebase-admin/README.md
@@ -44,6 +44,14 @@ function authenticate (context) {
 }
 ```
 
+### listUsers
+Lists all users from Firebase authentication. Outputs {users: [], nextPageToken: 'token'}.
+```js
+function listAllUsers (context) {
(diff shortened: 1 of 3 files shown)


henri-hulski:
I see that now the commit history gets a little mixed up.
It would also be good to merge the 2 PRs as otherwise they will conflict. 

So I think the easiest would be to close this one and add a second commit to #1191 for listUsers.

henri-hulski:
Here is a conflict with the base branch. Could you rebase?
There should also be only 2 commits here.

christianalfoni:
@henri-hulski How exactly to solve this? Just rebase?

henri-hulski:
@christianalfoni Not sure what happened here. My first proposition was to close this one and add the `feat(firebase-admin): added listUsers` commit to #1191.
This should be straightforward.

christianalfoni:
Just created new PR, sorry @tomhah, too much git 💩 Moved it here: https://github.com/cerebral/cerebral/pull/1198

Issue #1184, opened 2017-11-05
RFC: Consistent API


christianalfoni:
Okay, so my thoughts came together this morning :)

I want to change the API surface of Cerebral to:

```js
import { Controller, Module, Provider, Compute } from 'cerebral'
```

These are the constructs that Cerebral consists of and you use them like this:

### Controller
The controller takes a module and options.

```js
const controller =Controller(rootModule, {
  devtools: null
})
```

### Module
The module is how you structure your application.

```js
const app =Module({
  state: {},
  signals: {},
  providers: {},
  modules: {},
  catch: null
})
```

1. Providers are now an object instead of an array. This makes a more sense as providers needs a name and we do not need this `provide` factory:

```js
const app =Module({
  state: {},
  providers: {
    http:HttpProvider({})
  }
})
```

2. The only way to `catch` an error now is at the module level. Errors propagate up to parent modules if not caught. Meaning there is no signal specific error catching or "global". "Global" would be your root level module.

You still have a callback version
```js
const app =Module(({ name, path, controller ) => ({
  state: {}
}))
```

### Provider
Providers will now have a stronger concept. The concept is being the API between your application and your tools of choice. It can only be defined in one way, an object with methods. This makes things consistent and also enforces the idea that you think of a provider as a "bridge" between your app and the tools your are using.

```js
const myProvider = Provider({
  foo() {}
})
```

With existing context:
```js
const myProvider = Provider(({ http }) => ({
  foo() {
    return http.get()
  }
}))
```

All providers defined this way is automatically cached and wrapped for debugger purposes. You can still just add a tool if you want to:

```js
Module({
  providers: { uuid }
})
```

But these are not wrapped.

### Compute

This is just a change to be consistent with the other APIs:

```js
const myComputed =Compute(state`foo`, (foo) => ())
```

### Catch
There is no need to use `Map` for catching errors. Also catching is now moved to one spot... the module.

```js
Module({
  signals: {},
  catch: [
    [Error, sequences.catchError]
  ]
})
```

Allrighty... please comment and I will get to work on this as soon as we agree. It is of course a breaking change, but it is small. This is what needs to be changed in the apps:

1. Change controller configuration to take in a top level module
2. Wrap all modules in Module constructor
3. Change configuration of providers to use object

Here is a video as well: https://drive.google.com/open?id=0B1pYKovu9UpyT1NTWVlHN0c2ZTA

henri-hulski:
[...]

PR #1193, opened 2017-11-12
WIP: Canary

christianalfoni:
This is the implementation of RFC: https://github.com/cerebral/cerebral/issues/1184

```js
import { Controller, Module, Provider, Compute, CerebralError } from 'cerebral'
```

These are the constructs that Cerebral consists of and you use them like this:

### Module
The module is how you structure your application.

```js
const app = Module({
  state: {},
  signals: {},
  providers: {},
  modules: {},
  catch:[]
})

// Or callback
const app = Module(({name, path, controller}) => ({
  state: {}
})
```

1. Providers are now an object instead of an array. This makes a more sense as providers needs a name and we do not need this `provide` factory:

```js
const app = Module({
  state: {},
  providers: {
    http: HttpProvider({})
  }
})
```

2. The only way to `catch` an error now is at the module level. Errors propagate up to parent modules if not caught. Meaning there is no signal specific error catching or "global". "Global" would be your root level module.

```js
const app = Module({
  state: {},
  catch: [
    [HttpProviderError, sequences.catchHttpError]
  ]
})
```

### Controller
The controller takes a module as first argument, and other options as a second argument.

```js
const controller = Controller(rootModule, {
  devtools: null
})
```

### Provider
Providers will now have a stronger concept. The concept is being the API between your application and your tools of choice. It can only be defined in one way, an object with methods. This makes things consistent and also enforces the idea that you think of a provider as a "bridge" between your app and the tools your are using.

```js
const myProvider = Provider({
  foo() {}
})
```

You can point to the existing context using `this.context`. It works this way to simplify with just one API surface and we can do prototypal optimizations under the hood.
```js
const api = Provider({
  getUser() {
    return this.context.http.get('/user')
  }
})
```

All providers defined this way is automatically optimized and wrapped for debugger purposes. If you just want to use a 3rd party tool directly, you can still do that by just attaching it:

```js
Module({
  providers: { uuid }
})
```

### Compute

This is just a change to be consistent with the other APIs:

```js
const myComputed = Compute(state`foo`, (foo) => ())
```

### CerebralError
To more easily create different error types you can now define your errors by extending the CerebralError:

```js
import { CerebralError } from 'cerebral'

export class AppError extends CerebralError {}
export class ApiError extends CerebralError {}
```

### Module tag
You can now use a `module` tag to point to the state of the module running the signal:

```js
[
  set(module`foo`, 'bar')
]
```

### Migration

1. Change controller configuration to take in a top level module
2. Wrap all modules in Module constructor
3. Change configuration of providers to use object
4. Change any custom providers to use the Provider factory


docs/api/module.md
```diff
@@ -1,40 +1,42 @@
 # Module
 
-Modules help you structure your state and signals into logical units. You can think of them as namespaces for state and signals.
+Modules are the core building blocks of your application. This is where you define state, signals, other modules etc.
```
henri-hulski: Maybe better 'submodules' instead of 'other modules'.

docs/api/tags.md
```diff
@@ -12,6 +12,7 @@ state`items.${props`itemKey`}`
 This allows you to express signals and component dependencies more effectively. There are 4 tags in Cerebral.
 
 - **state** - Used in signals and connect to target state
+- **module** - Used in signals and connect to target state of the module running the signal
```
henri-hulski: How do you use use 'module' in connect as the view is not bound to a specific module.

docs/api/operators.md
```diff
@@ -5,13 +5,13 @@ Read more about operators in the [Cerebral in depth - Operators](https://www.jsb
 
 ## State operators
 
-The methods for changing state within actions is also available as operators. All state operators support using both **state** and **props** tags as values.
+The methods for changing state within actions is also available as operators. All state operators support using **state**, **module** and **props** tags as values.
```
henri-hulski: Should be "within actions **are** ...".

docs/introduction/providers.md
```diff
@@ -26,62 +27,49 @@ The [devtools](/docs/api/devtools) also adds its own provider called **debugger*
...
+    http
+  }
 })
 ```
 
-When using the **provide** factory the provider will automatically be wrapped by the Cerebral devtools, to track its usage in the debugger.
+**It does not matter what module exposes the provider, all providers are global. That means any provider defined is available in any action in the application.**
 
-Now this provider is available to any action:
+You can use providers for pretty much anything, though typically it is to handle some kind of side effect. Examples of providers is [@cerebral/storage](https://github.com/cerebral/cerebral/tree/master/packages/node_modules/@cerebral/storage), [@cerebral/firebase](https://github.com/cerebral/cerebral/tree/master/packages/node_modules/@cerebral/firebase) and [@cerebral/http](https://github.com/cerebral/cerebral/tree/master/packages/node_modules/@cerebral/http).
```
henri-hulski: Should be "Examples of providers **are** ...".

docs/api/tags.md
christianalfoni: @henri-hulski It is only for use in operators. should note that :)

docs/api/module.md
christianalfoni: agree, adjusting :)

packages/node_modules/@cerebral/mobx-state-tree/README.md
```diff
@@ -44,48 +42,41 @@ import { types } from 'mobx-state-tree'
...
-  // on the module, though nested models can of course also have views. Note
-  // that views are only to GET state, not to change it. Meaning that in actions
-  // do not use views with for example state.set etc., only state.get
-  views(state) {
-    return {
-      currentBox() {
-        return state.boxes[state.currentBoxId]
-      }
+  // Define getters
+  // store.upper // "BAR"
```
henri-hulski: @christianalfoni the method in the comment is different from the function name ("upper" -> "upperName").

packages/node_modules/@cerebral/mobx-state-tree/README.md
```diff
@@ -44,48 +42,41 @@ import { types } from 'mobx-state-tree'
...
+  getters: {
+    upperName() {
+      return this.name.toUpperCase()
+    }
+  },
+  // Define computed (receives a value to compute)
+  // store.helloFromFoo("someName")
+  computed: {
+    helloFrom(name) {
+      return `Hello ${this.name}, I am ${name}`
```
henri-hulski: Here also:
"helloFromFoo" -> "helloFrom"

docs/api/signal.md
```diff
@@ -193,45 +173,24 @@ function myAction ({path}) {
...
 ```js
-function myAction ({someProvider, path}) {
+function myAction ({ someProvider, path }) {
   return someProvider.doAsync()
-    .then((result) => path.bananas({data: result.data}))
+    .then((result) => path.bananas({ data: result.data }))
 }
 ```
-
-## Errors
```
Guria: leave and add removal note?

docs/api/tags.md
```diff
@@ -2,17 +2,24 @@
...
-import {state, props} from 'cerebral/tags'
+import { state, props } from 'cerebral/tags'
 
 state`items.${state`currentItemKey`}`
 
 state`items.${props`itemKey`}`
 ```
 
-This allows you to express signals and component dependencies more effectively. There are 4 tags in Cerebral.
+There are 6 tags in Cerebral.
```
Guria: Seems it's better not to count tags here :)

Guria:
Changed to Approve, docs shouldn't block getting into `next`

packages/node_modules/@cerebral/mobx-state-tree/README.md
christianalfoni: 👍

docs/api/action.md
```diff
@@ -122,12 +132,5 @@ function someActionFactory(someArgument) {
 You have access to the controller instance on the context:
 
 ```js
-function someAction ({controller}) {}
-```
-
-## Execution
```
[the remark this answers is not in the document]
christianalfoni: I just removed docs as this is very low level. Do not see any reason to use it for application development.

docs/api/signal.md
christianalfoni: We have it in migration docs, think that is enough :)

PR #1197, opened 2017-11-19
docs(website): add official typescript docs

christianalfoni:
Okay, so I need help with this :)

I think we should write official typescript docs recommending how to write Cerebral with typescript. These are the things I hope we can get in there with as much type stuff and autosuggestions as possible:

- Signals
- Actions
- State
- Connect

I think the docs would be nice with a short summary on top of what Typescript helps you with and then specific subtitles of API it affects and example.

Think this would be a great start for the official Typescript story :)

### DIFF
--- a/docs/introduction/typescript.md
+++ b/docs/introduction/typescript.md
@@ -1,0 +1,1 @@
+# Typescript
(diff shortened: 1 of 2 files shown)


christianalfoni:
Closing this because of `fluent`

PR #1205, opened 2017-12-04
maxLength: check isExisty before checking length

joseds:
maxLength checks on null or undefined lead to an exception. Note that contrary to minLength, we don't need to check isEmpty since the length is defined for an empty string and its length 0 works fine with maxLength.

### DIFF
--- a/packages/node_modules/@cerebral/forms/src/rules.js
+++ b/packages/node_modules/@cerebral/forms/src/rules.js
@@ -80,7 +80,9 @@ const rules = {
     return value === get(state`${field}.value`)
   },
   maxLength(value, length) {
-    return value.length <= length
+    return (
+      !rules.isExisty(value) || value.length <= length
+    )
   },
   minLength(value, length) {
     return (


packages/node_modules/@cerebral/forms/src/rules.js
```diff
@@ -80,7 +80,9 @@ const rules = {
     return value === get(state`${field}.value`)
   },
   maxLength(value, length) {
-    return value.length <= length
+    return (
+      !rules.isExisty(value) || value.length <= length
```
henri-hulski: I think we could also add
`|| rules.isEmpty(value) `
like in the `regexp` and `minLength` rules.

edgesoft: @henri-hulski That is a very good suggestion. Also add null, undefined test to the rules.js test file....

henri-hulski: `isExisty` actually checks null and undefined.

edgesoft: Yes, but we don't know if all rules have isExisty....

edgesoft: And if they should...

henri-hulski: Most of the rules already have isExisty through regexp.
But yeah maybe it should be added to all rules (where it makes sense).

edgesoft: Ok, just commented on the undefined null checks in test..... Think we need more of these to catch errors...

PR #1207, opened 2017-12-05
Fix inferno typings

br1anchen:
I think both `cerebral` and `inferno` do not have fully covered types currently.
I try to add types based on `@cerebral/inferno` use cases now, would like to have some reviews about it.
Since it is a bit challenging for me to improve typescript support for `@cerebral/inferno` based on less experience of `cerebral` and `inferno`


packages/node_modules/@cerebral/inferno/index.d.ts
```diff
@@ -1,12 +1,31 @@
...
-    getChildContext(): any
-    render(): InfernoChildren
+interface InfernoComponentClass<P> {
+    new (props?: P, context?: any): InfernoComponent<P, ComponentState>;
 }
-export function Container(opts: any): any
 
-export function connect(props: any, comp: (props: any) => void): any
-export function connect(props: any, comp: any): any
+interface FunctionalInfernoComponent<P> {
```
gaspard: We could remove the `Inferno` in types since the package already indicates that we are dealing with inferno. I see that the 'inferno' package has such types but we do not have this convention here for other packages. You can use the simplest types possible, like `Props`, `StatelessComponent`, `ComponentClass`, etc. This makes these typings easier to understand and maintain.

packages/node_modules/@cerebral/inferno/index.d.ts
```diff
@@ -1,12 +1,31 @@
...
+interface SignalsMap {
+    [signalName: string]: SignalChain
+}
+
+export const StateContainer: InfernoComponentClass<{state: any, signals: SignalsMap, children?: InfernoChildren}>
+export const Container: InfernoComponentClass<{controller: ControllerClass, children?: InfernoChildren}>
+
+// TODO: expose PropsMap from cerebral
+interface PropsMap {
+    [prop: string]: ResolveValue<any>
```
gaspard: No need for <any>, it can be omitted (makes the types simpler)

packages/node_modules/@cerebral/inferno/index.d.ts
```diff
@@ -1,12 +1,31 @@
...
+}
+
+export function connect<P extends PropsMap>(props: P, comp: InfernoCompnentType<Record<keyof P, any>>): InfernoComponentClass<P>
```
gaspard: This does not work because passed props are not taken into account. I know it is tempting to use the props map as source for types but until we have a solution (like [this](https://gist.github.com/gaspard/773647c5438bce74b166d4f0708a933c) one) to have proper types, having `any` does more harm then good.

Better be explicit and let users put any in there if they want to:

```ts
// Props are signal/state props, EProps are external props
export function connect<Props, EProps = {}>(propsMap: PropsMap, comp: Component<Props & EProps>): ComponentClass<EProps>
```

Also having the same type signature between `react` and `inferno` makes it easier to switch from one to the other.

packages/node_modules/@cerebral/inferno/package.json
```diff
@@ -14,17 +14,21 @@
...
-  "dependencies": {
-    "cerebral": "next"
+    "babel-plugin-inferno": "^3.2.0",
+    "cerebral": "^3.6.0-1511263649330",
+    "function-tree": "^2.2.1",
+    "inferno": "^3.10.1",
+    "inferno-component": "^3.10.1",
+    "typescript": "^2.6.2"
   },
+  "dependencies": {},
```
gaspard: Dependencies should include `"cerebral": "next"`.

Dev-dependencies should not include `cerebral` or `function-tree` (this one comes along with `cerebral`).

No need to include `typescript` here, we keep in in the monorepo root and it makes it easier to update without having to go into each package changing the dependency.

packages/node_modules/@cerebral/inferno/package.json
```diff
@@ -14,17 +14,21 @@
...
   "scripts": {
     "build": "cross-env BABEL_ENV=production babel src/ --out-dir=lib/ -s",
     "coverage": "nyc --reporter=lcov --reporter=json npm run test",
     "prepublish": "npm run build",
     "test": "mocha -r test/setup --compilers js:babel-register \"src/**/*.test.js\"",
-    "test:watch": "npm run test -- --watch"
+    "test:watch": "npm run test -- --watch",
+    "test:typings": "tsc -p tsconfig.json"
```
gaspard: Please look at how typings are run from @cerebral/react package. We like to run them along with all the other tests.

packages/node_modules/@cerebral/inferno/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
```
gaspard: We don't use tsconfig, please look at @cerebral/react for how we run the tests...

packages/node_modules/@cerebral/inferno/index.d.ts
br1anchen: Good to know that there are `External Props` for `connect` in this case.
I think generic type `P` will be inferred from `props` argument, which will be used to enforce/type `comp`. 
The problem here is that all `Tag` functions in `props` object is typed as `(path: TemplateStringsArray | string[], ...values: any[]): Tag<T>`, which `Tag<T>` is Object rather than function here.
Then in test example:
```
  ({foo, click}) => {
    return <div onClick={click}>{foo}</div>
  }
```

`click` type will be `Tag<T>` which is not assignable to `EventHandler`... that's reason, a mapped type `Record<keyof P, any>` is in use.
@gaspard I might misunderstand `Tag<T>` here?

br1anchen:
There is no third party `@types` module for `Inferno` JSX. It is still open [issue](https://github.com/infernojs/inferno/issues/686) from `Inferno`.
The reason I added it because we can just remove it when `Inferno` support it natively, otherwise typings we made here will not help developer using JSX in most cases. I agree that it might be better we drop external `InfernoJSX.d.ts` here, only necessary typings from `InfernoJSX.d.ts`.
I guess we just have to hope JSX typing support will come along with `Inferno` in the future.

christianalfoni:
@br1anchen there seems to be some linting issues here... hm... wonder how that got in. You have chance to check that out? Then we can move it in :)

christianalfoni:
@br1anchen Seems a rebase is needed here. Let me know if you need any help. Maybe we should put in `canary` first as there might be some types that needs to be changed anyways. Rather rebase when `canary` is in. Will merge in soon :)

br1anchen:
@christianalfoni branch is rebased from `next` now. 
Funny issue when `origin` and `upstream` not syncing at local :/

christianalfoni:
Sweet! It just made it, great :)

PR #1211, opened 2017-12-12
1210 Fix two bugs in http provider

### LINKED ISSUE #1210: http :: Issue Passing Options To httpGet And With Progress Calculation
I'm trying to set the `onProgress` option when using `httpGet` operator but its not working. In Cerebral debugger it looks like the template tag is not processed properly (though I'm only guessing as I'm not familiar with how things are done inside Cerebral).

```javascript
httpGet( props`url`, props`query`, props`http_options` )
```

20170-23_04-06-59-pm2

Looking at the code, it seems like the options should be passed through `convertObjectWithTemplates()` like the query is on the line above it (again, just guessing).

https://github.com/cerebral/cerebral/blob/e6a037c85451e57396f3a8f5008dd220ca4e2410/packages/node_modules/%40cerebral/http/src/operators/httpGet.js#L7-L11

**Update:** I also found a bug in the progress calculation that caused it to always return `0` or `1`. I submitted a PR to fix both issues.

lots0logs:
Fixes #1210

### DIFF
--- a/packages/node_modules/@cerebral/http/src/operators/httpGet.js
+++ b/packages/node_modules/@cerebral/http/src/operators/httpGet.js
@@ -7,7 +7,7 @@ function httpGetFactory(url, query = {}, options) {
       http.get(
         resolve.value(url),
         convertObjectWithTemplates(query, resolve, 'http.httpGet'),
-        convertObjectWithTemplates(options, resolve, 'http.httpGet'),
+        convertObjectWithTemplates(options, resolve, 'http.httpGet')
       ),
       path
     )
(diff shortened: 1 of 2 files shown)


henri-hulski:
2 tests are failing. Try to run `npm test` in "packages/node_modules/@cerebral /http" directory.
BTW there is also a linting error. You can check this with `npm run lint` in root directory. But when creating commit locally it should be caught during commit.
And try to always use the commit name conventions, as we use them for release notes. Easiest is to use `npm run commit`.

PR #1212, opened 2017-12-15
fix(Module): module should allow signals defined as a sequence, ts fixes

gaspard:


gaspard:
I'm fixing the failing tests.

PR #1213, opened 2017-12-17
fix(Controller): universal controller must adjust to new API surface

christianalfoni:


Guria:
Are Docs updated to cover this change?

PR #1216, opened 2017-12-18
fix(cerebral): remove direct 'lib' imports to avoid double linking

gaspard:
created a 'cerebral/internal' with all exports from 'lib' to avoid loading both 'es' and 'lib' compilations.

```js
// Replaced these imports
import BaseModel from 'cerebral/lib/BaseModel'
// With this
import { BaseModel } from 'cerebral/internal'
```

We really need to avoid importing directly from 'lib' in the future now that we have 'es'.


henri-hulski:
This PR seems also to contain a refactor of function-tree types.
Maybe better to put this in a separate commit.

gaspard:
@henri-hulski Yes, good idea.

PR #1218, opened 2017-12-19
docs(website): migration docs and smaller fixes

christianalfoni:


docs/migration/4_0.md
```diff
@@ -0,0 +1,141 @@
...
+[Documentation state](/docs/api/state)
+
+## Migration
+
+1. Change controller configuration to take in a top level module
+2. Wrap all modules in Module constructor
+3. Change configuration of providers to use object
+4. Change any custom providers to use the Provider factory
+5. Rename use of `compute` to `Compute`
+5. Any error catching now needs to be at module level without `new Map(...)`
```
henri-hulski: Maybe add to extend errors from `CerebralError`.
BTW you're using 2 times paragraph **5.** here.

packages/node_modules/@cerebral/mobx-state-tree/README.md
```diff
@@ -62,7 +62,7 @@ export default Module({
   // Define computed (receives a value to compute)
   // store.helloFromFoo("someName")
   computed: {
-    helloFrom(name) {
+    helloFromFoo(name) {
```
henri-hulski: Maybe use better `helloFrom` in comment as you're using it again later in `App.js`.
Above seems also be a mismatch. You have `store.upper` in comment but `upperName() ` as function.

PR #1233, opened 2017-12-25
chore(demos): fix demo publishing

christianalfoni:


henri-hulski:
Don't know which commits belong to this PR. Seems that it should be rebased.

Issue #1237, opened 2017-12-27
HTTP Provider's onProgress Callback Always Receives Either 0 or 1.


lots0logs:
This is caused by a bug on this line:
https://github.com/cerebral/cerebral/blob/c7f7131ed0c5f8958cf1e28004e0773d5d5402b9/packages/node_modules/%40cerebral/http/src/utils.js#L11

It should be:
```js
progress: Number((event.loaded / event.total) * 100).toFixed()
```

henri-hulski:
@lots0logs  Looks good! Do you want to create a Pull Request?

lots0logs:
@henri-hulski I rebased my PR on latest from `next` branch.

PR #1238, opened 2017-12-27
fix(http): fix logic bug that caused onProgress event to always emit 0 or 1

lots0logs:
ISSUES CLOSED: #1237


henri-hulski:
Seems that this PR needs to be rebased as it contains a commit which is already merged.

lots0logs:
@henri-hulski Done.

henri-hulski:
Great!

PR #1239, opened 2017-12-27
Update README.md

hipertracker:
fix syntax for C4

### DIFF
--- a/packages/node_modules/@cerebral/http/README.md
+++ b/packages/node_modules/@cerebral/http/README.md
@@ -150,7 +150,7 @@ const errorCatched = [
   displayError
 ]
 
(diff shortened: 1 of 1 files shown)


packages/node_modules/@cerebral/http/README.md
```diff
@@ -150,7 +150,7 @@ const errorCatched = [
   displayError
 ]
 
-Controller({
+Module({
   catch: new Map([
```
henri-hulski: `new Map()` should also be removed and "signal/global catch handlers" above should be "module catch handlers".

henri-hulski:
Will merge like this and fix it after.

PR #1241, opened 2017-12-27
Update index.d.ts

hipertracker:
Add missing TS definition for experimental "filterFalsy" option


hipertracker:
There is a small error. filterFalse should be optional.

hipertracker:
I created a second PR fixing that https://github.com/cerebral/cerebral/pull/1242

Issue #1244, opened 2017-12-28
@cerebral/router - experimental mapping always set a default state value to null


hipertracker:
```javascript
import { state } from 'cerebral/tags'
import Router from '@cerebral/router'

export default Router({
  routes: [
    {
      path: '/search',
      signal: 'app.searchRouted',
      map: {
        q: state`app.search.query`,
      },
    },
  ],
  onlyHash: false,
  query: true,
})
```
When the URL is opened it always destroys the original, default value in the state tree. The value is replaced always with `null`.

henri-hulski:
Fixed in #1253.

PR #1245, opened 2017-12-29
fix(router): preserve the default state value

hipertracker:
Fix the problem reported at https://github.com/cerebral/cerebral/issues/1244. The original code always set the default state value to null.

### DIFF
--- a/packages/node_modules/@cerebral/router/README.md
+++ b/packages/node_modules/@cerebral/router/README.md
@@ -297,22 +297,22 @@ const controller = Controller({
 
 #### computed mapping
 
-You can use a `compute` value here to run a computed in order to prepare
+You can use a `Compute` value here to run a computed in order to prepare
 the value passed to build the url.
 
 ```js
 map: {
-  urlKey: compute(/* ... */)
+  urlKey: Compute(/* ... */)
 }
 ```
 
-If you use a `compute` the router cannot map back from the url key to the
+If you use a `Compute` the router cannot map back from the url key to the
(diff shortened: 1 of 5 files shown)


Guria:
Great. Could we add a test to cover this case?

Guria:
Hmm, does url updated with value from store?

hipertracker:
Yes, the url is updated with the default value from store, but I cannot create a unit test to prove it. 

I have created another unit test showing for preserving the default state but how to amend it to this commit at Github?

Guria:
Just make a push from clone of your forked repo to patch-11 branch.

henri-hulski:
You have to do it from your local machine using git:
```sh
git clone git@github.com:hipertracker/cerebral.git
git checkout patch-11
```
Then you can eit amend to your commit or add new commits. When pushing it to the server (when amending you need a force push) it will be automatically added to the PR.
```sh
git push [--force]
```
Remember also to use `npm run commit` instead of `git commit` as this helps you to use the schema for the commit message which is after used for creating the release notes.
Otherwise we have to change your commit message later. 
```sh
git add .
npm run commit
```

henri-hulski:
Sorry I was hitting the wrong button. :-)

hipertracker:
Sorry, I send the "amend" and "push" commands with Tower git client instead of the terminal. The commit title has to be changed.

henri-hulski:
The PR should be rebased as it contains commits which are already merged but the commit message was adjusted afterwards.

PR #1251, opened 2018-01-07
Update README.md

jtasek:
Add a comma after isValidForm call to make the example work


henri-hulski:
Seems to be the same as #1250.
Closing.

PR #1252, opened 2018-01-13
feat(firebase): add getDownloadURL to firebase provider for storage

JohannesAnd:
Adds getDownloadURL to the firebase provider. This allows for generating file-urls based on path and filename. Typically used for generating image-urls to put in img-tags based on filename and path in firebase storage.


JohannesAnd:
There is a console.log in there, but I do not know how to remove it without introducing a new entry in the git history? What is the preferred approach?

PR #1253, opened 2018-01-15
fix(router): preserving the default state value

henri-hulski:
Fix the problem reported at #1244. The original code always set the default state value to null.

Moved from #1245 and cleaned up.


henri-hulski:
@Guria Could you have another look if it's ready now?

PR #1261, opened 2018-02-07
Fix typo in signals.md

evanblack:
Noticed a syntax typo while reading through the docs

### DIFF
--- a/docs/introduction/signals.md
+++ b/docs/introduction/signals.md
@@ -70,7 +70,7 @@ It is possible to diverge execution down specific paths.
 import * as actions from './actions'
 
 export const initialize = [
-  actions.getUser, {
+  actions.getUser: {
     success: actions.setUser,
     error: actions.setError
   }


evanblack:
Not sure why the tests failed. I changed 1 line in 1 of the markdown files...

evanblack:
Okay. This just seems like a really weird syntax for paths then  ¯\_(ツ)_/¯

christianalfoni:
@evanblack Yeah, the function and the object is related by order :) It is the only way to express it declaratively with arrays and objects. Like we could wrap it:

```js
[
  paths(myAction, {
    success: [],
    error: []
  })
]
```

But it was chosen to do it by order to make it as declarative as possible. But can totally understand the syntax throws you off at first. Though you quickly get used to it and appreciate it being as clean as possible :)

Typically prettier cleans it up like this:

```js
[
  myAction,
  {
    success: [],
    error: []
  }
]
```

Which makes it more clear how the syntax works

evanblack:
Thanks for the explanation! Makes sense!

PR #1264, opened 2018-02-10
ts(forms): add Provider return type

thaerlabs:
getting `'FormsProvider', which lacks return-type annotation, implicitly has an 'any' return type` error, had to add the return type as *Provider*


packages/node_modules/@cerebral/forms/index.d.ts
```diff
@@ -1,6 +1,6 @@
-import { ResolveValue } from 'function-tree'
+import { ResolveValue, Provider } from 'function-tree';
```
henri-hulski: The semicolon at the end of the line should be removed.

PR #1269, opened 2018-02-18
Fix outdated doc for providers

### LINKED ISSUE #1266: Is providers an object or array?
first doc:

image

another doc:

image

? which ? is ? it ?

quinn:
in version 4 providers is now an object. Fixes #1266


quinn:
@christianalfoni it looks like you like doing the `{ http }` style for providers, let me know and I will update the doc to create a `http` var before the config.

Issue #1167, opened 2017-10-17
UniversalController is not included in the typescript definitions


garth:

christianalfoni:
This can be added here: https://github.com/cerebral/cerebral/blob/next/packages/node_modules/cerebral/index.d.ts

PR #1271, opened 2018-02-18
Add missing UniversalController typing

br1anchen:
Related issue #1167 

For `runSequence()`, I was looking through how `function-tree` is typed, and find out it is using:
```
createStaticTree(tree: Primitive | Array<Function | Primitive>): Array<Primitive>
```
I wonder for `controller` usage, we might want more specific `SequenceFunction<TStateAndProps>() => any` than `Function` since I see that's used for in example:
```
controller.runSequence([
  function myAction ({ state, props }) {
    state.set('app.isAwesome', props.isAwesome)
  }
], {
  isAwesome: true
})
  .then(() => {
    // I am done running
  })
```


christianalfoni:
Awesome! Thanks a bunch @br1anchen ! Regarding your comment there will be quite a few changes with `fluent`, so lets have a review after it is merged in :)

br1anchen:
Agree. I can see what `fluent` solve from yesterday talk :) Will try to find time to check `fluent`.
Good work 👍 for your guys.

christianalfoni:
Awesome! :) Here are some docs:

https://github.com/cerebral/cerebral/blob/canary/docs/guides/typescript.md

Issue #774, opened 2017-03-22
allow setting timeout in HttpProvider


kgdev:
Right now I have to do this like:

```
            HttpProvider(providerOptions => ({
                ...providerOptions,
                baseUrl: config.REQUEST_BASE_URL,
                headers: {
                    'X-Requested-With': 'XMLHttpRequest'
                },
                onRequest (xhr, options) {
                    xhr.timeout = config.REQUEST_TIMEOUT;
                    providerOptions.onRequest(xhr, options);
                }
            })),

```

christianalfoni:
This has to be changed here: https://github.com/cerebral/cerebral/blob/next/packages/node_modules/%40cerebral/http/src/request.js. Basically check if the options has a `timeout` and use it :)

PR #1276, opened 2018-02-20
feat(http): add support for timeout

codrin-iftimie:
Added `timeout` option
Added `ontimeout` callback to throw as an error

Upgraded xhr-mock to be able to add tests for timeout

TODO: 
- [x] update docs

ISSUES CLOSED: #774

### DIFF
--- a/packages/node_modules/@cerebral/http/README.md
+++ b/packages/node_modules/@cerebral/http/README.md
@@ -206,6 +206,9 @@ export default [
     abort: [
       /* PROPS: {error: {...}} */
     ],
+    timeout: [
+      /* PROPS: {error: {...}} */
+    ],
 
     // Optionally any status code, ex. 404: []
     '${STATUS_CODE}': [
@@ -264,6 +267,9 @@ export default [
     abort: [
       /* PROPS: {error: {...}} */
     ],
+    timeout: [
+      /* PROPS: {error: {...}} */
+    ],
 
     // Optionally any status code, ex. 404: []
     '${STATUS_CODE}': [
@@ -320,6 +326,9 @@ export default [
       /* PROPS: {error: {...}} */
     ],
     abort: [
+      /* PROPS: {error: {...}} */
+    ],
+    timeout: [
       /* PROPS: {error: {...}} */
     ],
 
@@ -386,6 +395,9 @@ export default [
     abort: [
       /* PROPS: {error: {...}} */
     ],
+    timeout: [
+      /* PROPS: {error: {...}} */
+    ],
 
     // Optionally any status code, ex. 404: []
     '${STATUS_CODE}': [
@@ -444,6 +456,9 @@ export default [
       /* PROPS: {error: {...}} */
     ],
     abort: [
+      /* PROPS: {error: {...}} */
+    ],
+    timeout: [
       /* PROPS: {error: {...}} */
     ],
 
--- a/packages/node_modules/@cerebral/http/README.md
+++ b/packages/node_modules/@cerebral/http/README.md
@@ -62,7 +62,7 @@ function updateDefaultHttpOptions({http}) {
 ```
 
 ## abort
-You can abort any running request, causing the request to resolve as status code **0** and set an **isAborted** property on the response object.
+You can abort any running request, causing the request to resolve as status code **0** and sets the type to **abort** on the error object.
 
 ```js
 function searchItems({input, state, path, http}) {
@@ -70,7 +70,7 @@ function searchItems({input, state, path, http}) {
   return http.get(`/items?query=${input.query}`)
     .then(path.success)
     .catch((error) => {
-      if (error.isAborted) {
+      if (error.type === 'abort') {
         return path.abort()
       }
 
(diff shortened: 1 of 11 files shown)


packages/node_modules/@cerebral/http/src/request.js
```diff
@@ -1,11 +1,16 @@
...
-export default function request(options, cb) {
+export default function request(options = {}, cb) {
   const xhr = new XMLHttpRequest()
+
   xhr.addEventListener('progress', cb)
   xhr.addEventListener('load', cb)
   xhr.addEventListener('error', cb)
   xhr.addEventListener('abort', cb)
+
+  xhr.ontimeout = cb.bind(null, { type: 'timeout', currentTarget: xhr })
```
codrin-iftimie: There is a [eslint rule](https://eslint.org/docs/rules/callback-return) that forbids me from calling the callback with only one parameter. I can either keep this, or change the name of the variable.

On this same topic: I'm not 100% sure that this was in the scope of the initial issue, but it would be a nice to have.

packages/node_modules/@cerebral/http/src/utils.js
```diff
@@ -33,6 +33,17 @@ export function createResponse(options, resolve, reject) {
           )
         )
         break
+      case 'timeout':
```
codrin-iftimie: Would we need to add an extra param to the `HttpProviderError` class?
I think that most of the cases you'd like to treat them as generic, but there can be cases where you want to distinct between `abort`, `error` and `timeout` errors in paths.

I think adding a `type` param at the beginning  would be more efficient than the last arg (`isAborted`). I haven't thought too much on what are the implications, but it might bring a breaking change.

henri-hulski: Alternatively we could create some error subclasses like `HttpAbort`, `HttpTimeout` or `HttpError`, which extend `HttpProviderError`.
This way we can either catch the generic `HttpProviderError` or one of the specific errors.

packages/node_modules/@cerebral/http/src/request.js
henri-hulski: I don't think we use this rule in Cerebral as it's not part of standard.

codrin-iftimie: Oh... my bad, wrong link... I do get this 
> [eslint] Unexpected literal in error position of callback. (standard/no-callback-literal)
when trying to use it like:
```js
xhr.ontimeout = function() {
  cb({ type: 'timeout', currentTarget: xhr })
}
```
I've done further reading and fixed it

packages/node_modules/@cerebral/http/src/utils.js
christianalfoni: I would suggest adding a type first, as @codrin-iftimie suggests. The only reason is that we do not know the need for separate error flexibility yet. If we do need it that can be implemented later. A typical way to use the types would be:

```js
[
  equals(props`error.type`),
  {
    'timeout': [],
    'whatever': [],
    otherwise: []
  }
]
```

henri-hulski: Makes sense. :+1:

packages/node_modules/@cerebral/http/src/request.js
henri-hulski: Great!

packages/node_modules/@cerebral/http/src/utils.js
christianalfoni: @codrin-iftimie Oki, add a `type` property to **HttpProviderError**. Would be great if you also document it. Let us know if you need any help :)

codrin-iftimie: @christianalfoni I can take care of the documentation, no problem. I just want to make sure I took the right path here, before that :)

Thanks.

christianalfoni:
Coming together now @codrin-iftimie ! :) There seems to be a breaking change there though... the `isAborted` property of the error is removed? We still need to have that there, as it might break existing apps. We could deprecate it by:

```js
// After this.response is defined
Object.defineProperty(this.response, 'isAborted', {
  get () {
    console.warn('DEPRECATED - Please use error.type === "abort" instead)
    return isAborted
  }
})
```

This would be the best I think :)

codrin-iftimie:
This one is now ready to be merged. Let me know if there is anything I'm missing.

henri-hulski:
@christianalfoni  Maybe we should remove WIP from `.cz-config.js`.
It makes more sense to mark the whole PR as WIP, and it's difficult to change after the commit messages.

packages/node_modules/@cerebral/http/README.md
```diff
@@ -41,7 +41,13 @@ const http = HttpProvider({
 
   // When talking to cross origin (cors), pass cookies
   // if set to true
-  withCredentials: false
+  withCredentials: false,
+
+  // Provide a global request timeout for all calls
+  // which can be overwriten for request by providing
```
henri-hulski: Typo, should be "overwritten".

henri-hulski:
@codrin-iftimie Could you fix the typo?
Then I will merge it.

codrin-bytex:
@henri-hulski Done!

henri-hulski:
@codrin-iftimie  Great work!

PR #1293, opened 2018-03-21
feat(fluent): add module parameter to with connector

schotime:


packages/demos/todomvc-fluent/tslint.json
```diff
@@ -0,0 +1 @@
+{}
```
henri-hulski: I think this file should be removed from repo.

schotime:
@henri-hulski try running the fluent demo without it. You cannot. It needs to be there.

henri-hulski:
@schotime  That's interesting. Does it use tslint somewhere?
Will take a look when I'm back home.

henri-hulski:
Yep seems like tslint is somehow part of CRA.
Didn't know that so I removed that file.
@schotime  I think we should revert 70f2f4d2ce6f7528057 in this case.

henri-hulski:
Maybe in a separate commit.

schotime:
Reverted and updated PR

henri-hulski:
Great!

PR #1296, opened 2018-03-24
docs(release): allow only fast-forward merge into master

henri-hulski:


henri-hulski:
Yep. Actually when the merge is not fast-forward that means, that something is wrong with the master branch which has to be fixed like we had it in the recent release.
I saw you you made some merge commits in the releases before. I think the problem was already there. When using `--only-ff` this would not happen.

PR #1307, opened 2018-04-01
docs(fluent): change to existing view bindings approach

christianalfoni:
Okay. So I have been refactoring Codesandbox.io to TypeScript like crazy lately. We are using the new `fluent` package and I have realized that the connector we are exposing does not embrace Mobx, makes us reinvent the wheel and it requires more mental effort to understand it.

Just take this example:

```ts
export default connect()
  .with(({ state }) => ({
    user: { ...state.user, address: { ...state.user.address } }
  })
  .to(
	function UserComponent ({ user }) {
      // Whatever logic
    }
  )
```

This is what you have to do to ensure that any change to the properties of a user and the nested address properties, will render the component. Unlike the new approach in the updated docs:

```ts
import { connect, WithCerebral } from 'app/fluent'

type Props = WithCerebral

const UserComponent: React.SFC<Props> = ({ store, signals ) => (
  // whatever
)

export default connect<Props>()(UserComponent)
```

The last example uses `mobx-react` which automatically renders the component based on what state you access. This is really the beauty of Mobx, though less explicit. But considering the mindbendyness of the first example I think this is the best approach :)

### DIFF
--- a/docs/guides/typescript.md
+++ b/docs/guides/typescript.md
@@ -119,6 +119,12 @@ export type Context<Props = {}> = IContext<Props> & Providers
 export type BranchContext<Paths, Props = {}> = IBranchContext<Paths, Props> &
   Providers
 
+// This type is used to expose the state and signal typings to the components
+export type WithCerebral = {
+  store?: State
+  signals?: Signals
+}
+
 // This function is used to connect components to Cerebral and observing accessed state
 export const connect = <Props>() => {
   return (Component: IReactComponent<Props>): IReactComponent<Props> =>
@@ -241,6 +247,11 @@ export type Context<Props> = IContext<Props> & Providers
 
 export type BranchContext<Paths, Props> = IBranchContext<Paths, Props> &
   Providers
+
+export type WithCerebral = {
+  store?: State
+  signals?: Signals
+}
 
 export const connect = <Props>() => {
   return (Component: IReactComponent<Props>): IReactComponent<Props> =>
--- a/docs/guides/typescript.md
+++ b/docs/guides/typescript.md
@@ -57,24 +57,25 @@ controller.signals
 
 ## Exposing the controller to React
 
-You expose the controller using the **Container** from **fluent**:
+You will need to inject the state and signals of the Cerebral controller into your view layer of choice. Since the state store is based on **Mobx** you will need to use
+a relevant bindings. For example [mobx-react](https://github.com/mobxjs/mobx-react) as shown in this example.
 
 ```ts
 import * as React from 'react'
 import { render } from 'react-dom'
 import { controller } from './controller'
-import { Container } from '@cerebral/fluent'
+import { Provider } from 'mobx-react'
 import { App } from './components/App'
 
 render(
-  <Container controller={controller}>
+  <Provider store={controller.state} signals={controller.signals}>
     <App />
   </Container>,
   document.querySelector('#app')
 )
 ```
 
-You will later see examples of how you actually connect components.
+You can inject this however you want, this is just an example. You will see later how you actually provide the state and signals to the components.
 
 ## Creating the fluent file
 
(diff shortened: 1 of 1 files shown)


Guria:
Hmm. I like `with` / `to` approach because I can leave there simple computations and bindings like:
```ts
import * as React from 'react'
import { ListItem, ListItemText } from 'material-ui'
import { connect } from 'fluent'

export default connect()
  .with(({ state, signals, props }) => ({
    label: state.dimensionsLabel || 'N/A',
    onClick: () => signals.showDialog({ dialog: 'dimensions' })
  }))
  .to(function DimensionsListItem({ label, onClick }) {
    return (
      <ListItem button onClick={onClick}>
        <ListItemText primary="Основные размеры" secondary={label} />
      </ListItem>
    )
  })
```

For me it looks better also because we have here good separation of concerns. I really don't like accessing deep values inside render even with typesafety provided by TS. If we omit `with` part component becomes to tight to implementation. Separated way allows to extract component and easily test just by feeding with props or even connecting with other state controllers in another project.

christianalfoni:
Yeah, I do not see those as negatives, but the complexity of deciding what to observe is so big. Take an example of a simple array:

```ts
.with(({ state }) => ({
  items: state.items
})
```

This will not work when you access properties on individual items. You have to map it like this:

```ts
.with(({ state }) => ({
  items: state.items.map(item => ({ ...item }))
})
```

It is a huge deal I think

Guria:
And what about my points above?
I'd like to make clean components that doesn't aware of state and signals.
I want to be able to test them, put them in storybooks, etc without extra wiring up and mocking around.
Would you propose such way with `react-mobx` connector?
Can we take good parts of both worlds?

Guria:
Also what about Components tab in debugger. I am missing it with fluent addon. It is also great and essential part of debuuger.

christianalfoni:
It would be great to get best of both worlds here, cause I totally agree with all your points. Which is why I wanted to do it like this in the first place. But lets disrupt the whole API here and think differently. What about:

```ts
const MyComponent: React.SFC<Props> = ({ observe, signals }) => observe(state =>
	<div>{state.user.name}</div>
)
connect<Props>()(MyComponent)
```

And then:

```ts
class MyComponent extends React.Component<Props> {
  render () {
    const { observe, signals } = this.props;

    return observe(state => (
      <div>{state.user.name}</div>
    )
  }
}
connect<Props>()(MyComponent)
```

Or you can pre-observe:

```ts
const MyComponent: React.SFC<Props> = ({ user }) => (
	<div>{user.name}</div>
)
connect<Props>().observe(({ state }) => ({
  user: state.user
})(MyComponent)
```

I know it does not cover everything, but just to disrupt a bit here, think fresh :) I think passing in an `observe` function makes a lot of sense. It gives the flexibility to observe "what is used".

christianalfoni:
What is nice about it as well is that we can also expose like `reaction` as a prop, which is useful to convert changes to state to imperative logic.

Ackos95:
@christianalfoni not sure that I see advantage of approach you've described. With `observe` as a prop to component, don't we just complicate component's internal stuff (having responsibility to wrap each function using state with `observe`, then confusion over whether component would update as well if 
`observe` is in some custom function of component, and not render etc). 

And if using `pre-observed` pattern, don't we get same structure as with `with-to` or am I missing something? Would this pattern simplify array observation?

How would you use signals inside component, does `signals` prop (injected by cerebral) gives us access to all signals in `controller`, do we need to use whole paths for each signal? If so that is a bit too much binding between component and controller (simple refactoring of signals structure could take even hours because of all component updates).

I agree tho that some kind of update on connecting state (`mobx` part) should be done, to avoid those examples you've mentioned (`state.items.map(item => ({ ...item }))`), but not sure if complete change of interface is a best solution. Think that connection part shouldn't be simplified too much, in `redux` for example, also requires function to extract state and signals, which again gives developer more freedom to customize state and signals on per-component level.

And also there is always an option to separate extracting connection to state and component in two separate files, which can ease up testing, and reusability of both components and connectors. So in conclusion, I agree with you that connecting state should be a bit more simplified, but proposed approach introduces some restraints if used in planned way (`observe`  and `signal` props), and just a syntax change if used in `pre-observe` pattern.

christianalfoni:
@Ackos95 So sorry for not responding to this! Just been difficult to keep up. I hope C5 has wrapped up nicely!

PR #1309, opened 2018-04-02
chore(fluent): clean up typescript build process

henri-hulski:
This PR does the following:
* In @cerebral/fluent it moves the declaration files to lib directory and tsc builds also for es2015 target.
* In function-tree/fluent tsc builds directly for commenjs and es2015 targets and tests are running directly on ts files.
* Removes all built files from version control.
* Updates ts-node and eslint-plugin-typescript dependencies.


packages/node_modules/@cerebral/fluent/tsconfig.json
```diff
@@ -2,13 +2,17 @@
...
     "outDir": "./lib",
     "target": "es5",
-    "module": "es2015",
     "strict": true,
     "strictPropertyInitialization": false,
     "moduleResolution": "node",
     "declaration": true,
-    "declarationDir": "./",
-    "lib": ["es2015", "dom"]
+    "declarationDir": "./lib",
```
Guria: i guess the default is outDir, so we can omit it

henri-hulski: Yep that's right.
Fixed.

PR #1312, opened 2018-04-09
fix(fluent): fix debugger double message

christianalfoni:


christianalfoni:
Will have to investigate this further

PR #1314, opened 2018-04-11
fix(fluent): make semicolon optional so pathFor works in minified builds

gabrielsadaka:
In the production builds the js minifier is stripping out semicolons causing `pathFor` in fluent to not find the signals stopping the router from triggering signals on route load.


packages/node_modules/@cerebral/fluent/src/paths.ts
```diff
@@ -1,4 +1,4 @@
-const pathLambdaRegex = /function\s*\((\w)\)\s*{\s*return\s*\1[.]?([\w.[\]]+);\s*}/g
+const pathLambdaRegex = /function\s*\((\w)\)\s*{\s*return\s*\1[.]?([\w.[\]]+);*\s*}/g
```
henri-hulski: I think to make it optional a question mark ("?") instead an asterisk ("*") would be correct.

gabrielsadaka: That is a good point, asterisk would have allowed 0 or many which is not incorrect but it would be highly unlikely so I have updated it to a question mark.

PR #1318, opened 2018-04-18
Cerebral proxy

christianalfoni:


Guria:
.cache must be ignored

christianalfoni:
Okay, so I did some changes to how we organize:

1. By default there are no "devDependencies" in packages anymore, all is installed on root. There was tons of duplicates
2. The only reason to use "devDependencies" in a package is for specific dev dependencies for that package. For example parcel plugins (required to be in package), conflicting types like jest and mocha
3. There is a new script type called "setup". If you insert this script in a package it will run when the root `npm run setup` is run. This allows you to prepare stuff inside the package. For example installing the devDependencies required specifically for that package

Cleaned typing, errors etc. Now demos should just work out of the box etc.

.eslintrc.json
```diff
@@ -23,7 +23,7 @@
   },
   "overrides": [
     {
-      "files": ["*.ts", "*.tsx"],
+      "files": ["*.ts", "*.tsx", "*.d.ts"],
```
henri-hulski: `"*.d.ts"` is not needed as it's already included in `"*.ts"`

henri-hulski:
Forgot that this is a older branch. So everything is okay.
Just one minor adjustment above.

christianalfoni:
Cool, done! :)

PR #1319, opened 2018-04-23
Update signal.md

pprimm:
line 82: someAction imported from '../actions/someAction' is referenced incorrectly as actions.someAction; should be referenced as just someAction with out the actions. prefix

### DIFF
--- a/docs/api/signal.md
+++ b/docs/api/signal.md
@@ -79,7 +79,7 @@ import * as actions from '../actions'
 
 export const myOtherSequence = [actions.doThis]
 
-export const mySequence = [actions.someAction, myOtherSequence]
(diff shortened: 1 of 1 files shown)


docs/api/signal.md
```diff
@@ -79,7 +79,7 @@ import * as actions from '../actions'
 
 export const myOtherSequence = [actions.doThis]
 
-export const mySequence = [actions.someAction, myOtherSequence]
+export const mySequence = [someAction, myOtherSequence]
 ```
```
henri-hulski: The `actions.` prefix is correct.
Just
```js
 import someAction from '../actions/someAction'
```
is obsolente and should be removed.

christianalfoni:
C5 imminent

PR #1320, opened 2018-04-25
chore(fluent): add type properties on ModuleClass type

etienne-dldc:
Adding type properties in ModuleClass type allow TypeScript to infer
them and merge state and signals from modules.


etienne-dldc:
I still think this can improve typings but it needs more investigation so I'm closing this.

PR #1323, opened 2018-04-26
chore: remove @type dependencies from monorepo

henri-hulski:


henri-hulski:
@christianalfoni found a better solution. :)

PR #1324, opened 2018-04-29
fix(shortcuts): override default browser behaviour

menberg:
Shortcuts like CMD-S trigger a save dialog by the browser but not the requested Cerebral signal. In
such a case, event.preventDefault is necessary to run the requested Cerebral signal.


christianalfoni:
@menberg Thanks for the contribution, but will have to close this as C5 does not have addons. There are many good shortcut tools out there that can be used as just a plain provider. Thanks so much for helping out, feels bad not bringing this in!

PR #1325, opened 2018-04-30
[Fluent] Connecting computed values / dictionaries

samdenty99:


christianalfoni:
Cerebral 5 will have typescript support, though thanks for contributing! And sorry for late action on this :-/

PR #1380, opened 2018-10-29
fix(cerebral): throw error when trying to replace computed

christianalfoni:


NervosaX:
This is not quite catching them in my particular case. If the values are nested inside an object, it accepts them.



As you can see here, when verifying it's seeing it as an object, rather than an object with computeds inside it, so it passes through just fine.

christianalfoni:
Okay, fixed that as well :)

Issue #1382, opened 2018-11-08
Stale values when computed function relies on another computed function


NervosaX:
If a computed value is reliant on another computed value, and the value changes in an asynchronous signal, not all computed functions update correctly.

Full reproduction of issue:
https://codesandbox.io/s/r4vp7wvzn

NervosaX:
A second issue with stale computed functions in a repro:
Press the plus button and it doesn't update the computed value, and after that it's fine.
https://codesandbox.io/s/rwpkkz6nln

NervosaX:
This is fixed with commit https://github.com/cerebral/cerebral/commit/989be397a6daccaa00fbab5fdf629e145315d910

PR #1385, opened 2018-11-09
fix(cerebral): add state and computed tags as dependencies to computeds

NervosaX:
Looked into my problem - looks like computed values aren't being added to the dependency map inside a computed function. Not entirely sure if this code is the way things should be done, but my app still works - and it fixes my problem. :)

Resolves https://github.com/cerebral/cerebral/issues/1382


NervosaX:
Fixed with commit:
https://github.com/cerebral/cerebral/commit/989be397a6daccaa00fbab5fdf629e145315d910

PR #1387, opened 2018-11-14
refactor(view): don’t recreate getters

fmal:
@christianalfoni while browsing the code for `mergeProps` i noticed that getters are probably unnecessarily recreated (if i'm wrong please discard this PR).

Another thing that made me wonder: if i define `mergeProps` fn, then `getProps` return early without attaching `get` and `reaction` to props. Wouldn't it be more consistent if these were always available?

### DIFF
--- a/packages/node_modules/cerebral/src/View.js
+++ b/packages/node_modules/cerebral/src/View.js
@@ -270,7 +270,6 @@ class View extends Watch {
     }
 
     if (this.mergeProps) {
-      const getters = this.controller.createContext(props)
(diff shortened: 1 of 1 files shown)


christianalfoni:
Hi there!

Thanks for this, yeah, I think you are right :) 

About the last part. The `mergeProps` function is really considered a "pure function". First you provide what you need and then you change the structure of it. If we were to provide `get` and `reaction` it would not longer be pure and we would have to track and manage those dependencies as well.

Do you see any limitations with that restriction or you consider it "nice to have"? :)

PR #1405, opened 2019-07-22
Fix typo

kirillDanshin:


christianalfoni:
Ops, could you fix the test as well?

1) Devtools
2052       should warn when remember message sent if storeMutations option is false:

kirillDanshin:
@christianalfoni done

christianalfoni:
Great! :)

PR #1416, opened 2019-09-29
docs(cerebral): modify Get Started download instruction

ElliotPsyIT:
Codesandbox no longer displays a download button, making the current cerebral get started documentation incorrect.  I modified the 3rd bullet to indicate the need to use the Codesandbox File Menu to download a zip file.


christianalfoni:
Hi @ElliotPsyIT , I changed the documentation here to not go to Codesandbox. The reason is that unzipping and copying over files can be confusing as you do not copy over hidden files etc. I rather just refer to other build tools to get up and running :)

ElliotPsyIT:
Understood.  Actually, that was my first pull request :) Thanks for explaining the reason for closing without merging.

PR #1427, opened 2019-11-06
fix(cerebral): make undefined serializable

bernhardreisenberger:
If dev tools are enabled, the `verifyValue` function in Model.js will check if the given value is serializable. Trying to set a value to `undefined` in a sequence, results in an error ("You are passing a non serializable value into the state tree on path...")

Setting a value to `null` seems to be OK for cerebral, but `undefined` is not. A value from props (e.g. accessing an undefined property in a map) is likely to be `undefined` so i guess that behavior is not intended.

check the browser console after clicking on toggle, in this example: https://codesandbox.io/s/null-vs-undefined-tbxjd

### DIFF
--- a/packages/node_modules/cerebral/src/utils.js
+++ b/packages/node_modules/cerebral/src/utils.js
@@ -35,6 +35,9 @@ export function isComplexObject(obj) {
 }
 
 export function isSerializable(value, additionalTypes = []) {
+	if (value == null) {
+		return true
+	}
   const validType = additionalTypes.reduce((currentValid, type) => {
     if (currentValid || value instanceof type) {
       return true
@@ -43,9 +46,7 @@ export function isSerializable(value, additionalTypes = []) {
     return currentValid
   }, false)
 
-  if (
-    value !== undefined &&
-    (validType ||
+  if (validType ||
       (isObject(value) &&
         Object.prototype.toString.call(value) === '[object Object]' &&
         (value.constructor === Object ||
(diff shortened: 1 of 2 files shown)


christianalfoni:
@bernhardreisenberger Hi there!

This is an interesting topic. `undefined` is not really a serializable value. The reason being that if you set a key with a value to `undefined` and then serialize the it, the key is completely gone. Unlike if you set `null` the key holding that value will still be there when serializing.

This can also have other unintended effects in the underlying code.

Does that make sense?

bernhardreisenberger:
I can understand that convention (like the JSON standard).
`isSerializable` is only used, when devtools are enabled. If i read the code correctly, values are passed as JSON to the cerebral debugger, and therefore this convention is needed? 

Nevertheless, how should i deal with the problem when accessing a value in an object, that does not exist?
e.g. `set(state.listElement, state.list[props.index])`

In this example, `state.list[props.index]` is `undefined`, because the entry in the list does not exist. Assigning this value to `state.listElement` will result in an error (only if dev tools are enabled)
i came up with `set(state.listElement, state.list[props.index], (elem) => (elem || null))`

but that does not look as nice and easily readable. in both cases, i have to deal with the `undefined or null` in `state.listElement` anyway.

Do you have an idea or solution for that case?

codyseibert:
maybe `store.set(state.myKey, undefined)` should just alias `store.unset(state.myKey)` internally?  Any thoughts on that?

christianalfoni:
Ah, I see! Yeah, with that example I see it is not ideal. I think your suggestion @codyseibert makes a lot of sense. Cause then your state tree is kept in the state it has when serialized. Made a note of this to fix for next release :)

christianalfoni:
This is out now 😄

PR #1432, opened 2020-04-30
Update README.md

hipertracker:
Add info about a runtime required for Vue imported directly from 'vue' package.


christianalfoni (approved):
Great stuff! ❤️

PR #1436, opened 2020-10-29
fix(cerebral): prevent hard crash when missing dep

NervosaX:
I've had an issue in a particular version of nwjs (v0.44.1) which comes bundled with chromium 80.0.3987.87.
Some stars seem to align in this version only, and it's seemingly possible for dependencies to have the children it expects.

See below for the output: It seems to only have "**".

image

At any rate, this MR adds a guard to prevent a hard crash when this occurs. In normal circumstances this will never be needed, but error guards are never a bad thing.


christianalfoni:
Closing this as it keeps popping up on my Arch Pull Requests folder 😅

PR #1438, opened 2021-09-17
fix(cerebral): prevent checkForComputed from throwing when the path does not exist in state

### LINKED ISSUE #1412: Setting non-existent path in state works without devtools, throws error with devtools
When setting a non-existent path in the state (using a tag and store.set), if devtools is turned off the path in the state is created appropriately.  When devtools is turned on, the browser throws an error about reading a property of undefined on line 93 of Store.js in cerebral.  The Cerebral debugger shows the value in the state as if it had been set appropriately in the first place.

To reproduce, simply do this in any action with devtools enabled in your app
```store.set(state`path.does.not.exist`, 'hello');```
This will throw an error in the console that "does" does not exist.  Now look in the debugger and you will see a top-level state key named `path` whose value is an object that contains a key named `does`, .... and so on ... just as you would have expected.  However, the browser code threw the error and stopped.  If you catch the error and follow up with a get(state`path.does.not.exist`), you get undefined even though the debugger shows it in the state.

Then disable devtools and repeat, and you will see that it works as expected.

As a workaround, you can manually loop through and create all the elements of the path one at a time in your app and it will work, but this is messy.

JohannesAnd:
When the devtools are open, checkForComputed will try to access the current data that will be overwritten. When this data does not exist, it throws an error trying to access something on undefined

ISSUES CLOSED: #1412

### DIFF
--- a/packages/node_modules/cerebral/src/Model.js
+++ b/packages/node_modules/cerebral/src/Model.js
@@ -65,7 +65,7 @@ class Model extends BaseModel {
 
   checkForComputed(path) {
     const valueToReplace = path.reduce(
-      (currentState, key) => currentState[key],
+      (currentState, key) => key in currentState ? currentState[key]: {},
       this.state
     )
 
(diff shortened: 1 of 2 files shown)


henri-hulski (approved):
Makes sense. Thank you for the PR.

